PR #32, opened 2016-09-08
Fixed nvme write command in nvmetest

narasimhan-v:
nvme write command expects an input from stdin or a file.
Without that input, write waits forever expecting an input.
Fixed that in nvmetest.py


PraveenPenguin:
@narasimhan-v thanks

PR #44, opened 2016-09-16
Added Xfs test to avocado[v2]

PraveenPenguin:
v1 #42 
define variable as meaningful name
remove unused method

Signed-off-by: Praveen K Pandey <praveen@linux.vnet.ibm.com>

PR #45, opened 2016-09-21
Added Xfs test to avocado[V3]

PraveenPenguin:
V2 #44 
Added: helper function for validation
     : helper function for test failure
remove test_number as already cover in
       test_range
make logic more robust

### DIFF
--- a/fs/xfstests.py
+++ b/fs/xfstests.py
@@ -1,0 +1,185 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+#
+# Copyright: 2016 IBM
+# Author: Praveen K Pandey <praveen@linux.vnet.ibm.com>
+#
+# Based on code by Cleber Rosa <crosa@redhat.com>
+#   copyright: 2011 Redhat
+#   https://github.com/autotest/autotest-client-tests/tree/master/xfstests
+
+
+import os
+import glob
+import re
+import shutil
+
+from avocado import Test
+from avocado import main
+from avocado.utils import process, build, git
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Xfstests(Test):
+
+    def setUp(self):
+        """
+        Build xfstest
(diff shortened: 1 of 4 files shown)


PraveenPenguin:
job log  https://paste.fedoraproject.org/431877/44526261/

fs/xfstests.py
```diff
@@ -0,0 +1,185 @@
...
+    def setUp(self):
+        """
+        Build xfstest
+        Source: git://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git
+        """
+        sm = SoftwareManager()
+        packages = ['xfslibs-dev', 'uuid-dev', 'libtool-bin', 'e2fsprogs',
+                    'automake', 'gcc', 'libuuid1', 'quota', 'attr',
+                    'libattr1-dev', 'make', 'libacl1-dev', 'xfsprogs',
+                    'libgdbm-dev', 'gawk', 'fio', 'dbench', 'uuid-runtime']
```
apahim: No Fedora support? Avocado claims to support Fedora, so it would be nice if you could make the tests to work on Fedora as well. I see a packages list fro Fedora in `xfstests` README file, so you could use that as reference.

fs/xfstests.py.data/README
```diff
@@ -0,0 +1,41 @@
+This is a simple wrapper for running xfstests inside avocado. The steps to get
+started are really simple:
+
+1) Edit the configuration variables on the yaml file as well environment variable
+1.1) The variables 'TEST_DEV' and 'TEST_DIR' are mandatory and should be set to
+     a block device path and mount point path, respectively, that will be used
+     *exclusively* for xfstests. It must have the filesystem of your choice
+     previously created.
+     ex : export TEST_DEV =/dev/vda1 , export TEST_DIR=/mnt
```
apahim: Why not to set those using the yaml file? The test can then create the environment variables. No strong opinion here, but it's kinda annoying when the test is not self-sufficient. What do you think?

apahim:
Also, this should probably be improved:
```
2016-09-21 05:39:28,950 xfstests L0066 INFO | Tests available in srcdir: , 001, 001, 001, 001, 001, 001, 001, 002, 002, 002, 002, 002, 002, ....
```

fs/xfstests.py.data/README
PraveenPenguin: @apahim i think it can be using export variable as of now user won't do mistake.In yaml case user may  use older entry  , i thinking when it pass through yaml want to have check on test case that device is valid or not so i kept this in TODO , what you think ?

fs/xfstests.py
PraveenPenguin: @apahim sure for Fedora having some issue , working on this

PraveenPenguin:
@apahim raised New PR #46 with suggested change Please have a look .

Regards
Praveen

PR #50, opened 2016-09-22
Type changed from str to int for size variable

vrbagalkote:
Type changed and removed one repatative line


PraveenPenguin:
@vrbagalkote could you explain your commit in the commit message and please send only diff of your changes .

regarding travis failure Please fix pep8 issue

PR #58, opened 2016-09-23
Added Rmaptest in Avocado

PraveenPenguin:
Rmaptest create random VMAs mapped tasks

### DIFF
--- a/kernel/rmaptest.py
+++ b/kernel/rmaptest.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


PraveenPenguin:
job log : https://paste.fedoraproject.org/433546/74644010/

apahim:
When will the test be considered as failed?

PraveenPenguin:
@apahim this test is a stress test for system so i don't think there is any fail senario after test run system should alive

PraveenPenguin:
@apahim  can we check return code of the command please suggest ?

Regards
Praveen

apahim:
Well, you know better. If there isn't a fail condition, no need to check the test result.

apahim:
lgtm. thanks.

PR #59, opened 2016-09-24
Added Pmqa Testsuite test in avocado

PraveenPenguin:
Pmqa Testsuite run cpu test  as cpufreq
cpuhotplug, cputopology, cpuidle, thermal

### DIFF
--- a/cpu/pmqa.py
+++ b/cpu/pmqa.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author:Praveen K Pandey <praveen@linux.vnet.ibm.com>
+#
+
+
+import os
+
+from avocado import Test
+from avocado import main
+from avocado.utils import process, git
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Pmqa(Test):
+
+    """
+    pmqa Testsuite these cpu test
+    cpufreq, cpuhotplug, cputopology
+    cpuidlee, thermal
+    """
+
+    def setUp(self):
+        '''
+        Build Pmqa Test
+        Source:
+        git://git.linaro.org/power/pm-qa.git
+        '''
+        self.failtest = 0
+
+        # Check for basic utilities
+        smm = SoftwareManager()
+        if not smm.check_installed("gcc") and not smm.install("gcc"):
+            self.error('Gcc is needed for the test to be run')
+
+        git.get_repo('git://git.linaro.org/power/pm-qa.git',
+                     destination_dir=self.srcdir)
+
+    def _run_one(self, testcase):
+
+        os.chdir(self.srcdir)
+
+        log = self.logdir + '/stdout'
+
+        ext_opt = ''
+        if testcase == 'cpuhotplug':
+            ext_opt = 'hotplug_allow_cpu0=0'
+
+        cmd = testcase + ' ' + ext_opt + ' run_tests'
+
+        process.system('make -C %s' % cmd, ignore_status=True, shell=True)
+
+        result = process.run("grep -wF 'fail' %s" % log, ignore_status=True)
+
+        if not result == '':
+            self.failtest += 1
+            self.log.info(result.stdout)
+
+    def test(self):
+
+        arg = self.params.get('run_arg', default='all')
+        tests = []
+
+        if arg == 'all':
+            tests = ['cpufreq', 'cpuhotplug',
+                     'cputopology', 'cpuidle', 'thermal']
+        else:
+            tests = [arg]
+
+        for test in tests:
+            self._run_one(test)
+
+        if not self.failtest == 0:
+            self.fail("few tests failed")
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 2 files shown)


PraveenPenguin:
job log :https://paste.fedoraproject.org/433800/17478147/

cpu/pmqa.py
```diff
@@ -0,0 +1,89 @@
...
+            self.error('Gcc is needed for the test to be run')
+
+        git.get_repo('git://git.linaro.org/power/pm-qa.git',
+                     destination_dir=self.srcdir)
+
+    def _run_one(self, testcase):
+
+        os.chdir(self.srcdir)
+
+        log = self.logdir + '/stdout'
```
apahim: `log = os.path.join(self.logdir, 'stdout')`

cpu/pmqa.py
```diff
@@ -0,0 +1,89 @@
...
+
+        ext_opt = ''
+        if testcase == 'cpuhotplug':
+            ext_opt = 'hotplug_allow_cpu0=0'
+
+        cmd = testcase + ' ' + ext_opt + ' run_tests'
```
apahim: This math'ish string concatenation hurts my eyes. Even more when it's not the only method used in the code. Do you mind in use the same method as in line 63 bellow? `cmd = '%s %s run_tests' % (testcase, ext_opt)`

cpu/pmqa.py
```diff
@@ -0,0 +1,89 @@
...
+
+        process.system('make -C %s' % cmd, ignore_status=True, shell=True)
+
+        result = process.run("grep -wF 'fail' %s" % log, ignore_status=True)
+
+        if not result == '':
```
apahim: `result` will never be `''` since it's an instance of the `CmdResult` class. You can both check the `result.exit_status` (preferred) or the `result.stdout`.

cpu/pmqa.py
```diff
@@ -0,0 +1,89 @@
...
+            self.failtest += 1
```
apahim: This function could return `True` or `False` so you don't have to bother controlling this `self.failtest`.

cpu/pmqa.py.data/pmqa.yaml
```diff
@@ -0,0 +1,7 @@
+#Usage: run_arg can  any of one from below list
```
apahim: Can't it be more than one? Like `cpufreq` and `cputopology`? I guess it could, since you're accepting `all`.

cpu/pmqa.py
```diff
@@ -0,0 +1,89 @@
...
+            self.log.info(result.stdout)
+
+    def test(self):
+
+        arg = self.params.get('run_arg', default='all')
+        tests = []
+
+        if arg == 'all':
+            tests = ['cpufreq', 'cpuhotplug',
+                     'cputopology', 'cpuidle', 'thermal']
```
apahim: This is not validating the `arg` neither accepting more than one test at once (except when it's `all`). Seems like this logics deserve it's own function, which returns a test list to be executed.

cpu/pmqa.py
PraveenPenguin: @apahim thanks , will do

cpu/pmqa.py.data/pmqa.yaml
PraveenPenguin: yes that make sense will do changes

cpu/pmqa.py
PraveenPenguin: sure

cpu/pmqa.py.data/pmqa.yaml
apahim: Or, you could take advantage of the multiplex, running always one test only in the code, and multiplexing the tests in the yaml file.

cpu/pmqa.py
PraveenPenguin: `self.failtest` using this variable In test method for make test fail because I trying  to run  all test   should run. Not exit with a  single sub test  fail what you think ?

cpu/pmqa.py
PraveenPenguin: sure that make sense will do, thanks

cpu/pmqa.py.data/pmqa.yaml
PraveenPenguin: sure

apahim: Like this:
```
~$ cat pmqa.yaml
pmqatest: !mux
    cpufreq:
        run_arg: cpufreq
    cpuhotplug:
        run_arg: cpuhotplug
    cputopology:
        run_arg: cputopology
    cpuidle:
        run_arg: cpuidle
    thermal:
        run_arg: thermal
```
Then the test would be only:
```python
def setUp(self):
    self.test_type = self.params.get('run_arg', default='cpufreq')

def test(self):
    cmd = 'make -C %s' % self.test_type
    if self.test_type == 'cpuhotplug':
        cmd += ' hotplug_allow_cpu0=0'
    cmd += ' run_tests'
    result = process.run(cmd, ignore_status=True)
    if result.exit_status != 0:
        self.fail('Test failed: %s' % result.stderr)
```

cpu/pmqa.py
apahim: You can control this in the `test()` itself. This function will return True/False and in the `test()` you can have a variable that is set to tru when this function returns False.

cpu/pmqa.py.data/pmqa.yaml
apahim: Anyway, just an option, case you want to make simpler test and enjoy the multiplex.

PraveenPenguin: @apahim thanks , but in result get any command status but i want to grep Fail as well. that's  why checking fail log in file

apahim: @PraveenPenguin no, you can grep the log if the exit_status does not provide what you want.

PraveenPenguin:
@apahim raised New PR #63  with suggested change Please have a look .

Regards

PR #64, opened 2016-09-27
Adding MOFED Install test

narasimhan-v:
This test verifies the installation of MOFED iso with different
combinations of input parameters.

### DIFF
--- a/io/net/infiniband/mofed_install_test.py
+++ b/io/net/infiniband/mofed_install_test.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Narasimhan V <sim@linux.vnet.ibm.com>
+
+"""
+Mellanox OpenFabrics Enterprise Distribution (MOFED) System Update Package,
+provides necessary softwares to operates across all Mellanox network adapter
+solutions supporting 10, 20, 40 and 56 Gb/s InfiniBand (IB); 10, 40 and 56 Gb/s
+Ethernet; and 2.5 or 5.0 GT/s PCI Express 2.0 and 8 GT/s PCI Express 3.0
+uplinks to servers.
+"""
+
+import os
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+
+
+class MOFEDInstallTest(Test):
+
+    """
+    This test verifies the installation of MOFED iso with different
+    combinations of input parameters, as specified in multiplexer file.
+
+    """
+
+    def setUp(self):
+        """
+        Mount MOFED iso.
+        """
+        self.iso_location = self.params.get('iso_location', default='')
+        if self.iso_location is '':
+            self.skip("No ISO location given")
+        self.option = self.params.get('option', default='')
+        self.tarball = self.fetch_asset(self.iso_location, expire='10d')
+        cmd = "mount -o loop %s %s" % (self.tarball, self.srcdir)
+        process.run(cmd, shell=True)
+        self.pwd = os.getcwd()
+        os.chdir(self.srcdir)
+
+    def install(self):
+        """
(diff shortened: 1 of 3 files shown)


narasimhan-v:
[job_log.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/494683/job_log.txt)

io/net/infiniband/mofed_install_test.py
```diff
@@ -0,0 +1,92 @@
...
+        Uninstalls MOFED, if installed fine.
+        """
+        cmd = "etc/init.d/openibd restart"
+        if not process.system(cmd, ignore_status=True, shell=True):
+            return
+        cmd = "ibstat"
+        if not process.system(cmd, ignore_status=True, shell=True):
+            return
+        cmd = './uninstall.sh --force'
+        if process.system(cmd, ignore_status=True, shell=True):
```
PraveenPenguin: here also can you please `if process.system('./uninstall.sh --force', ignore_status=True, shell=True):` ?

io/net/infiniband/mofed_install_test.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def uninstall(self):
+        """
+        Uninstalls MOFED, if installed fine.
+        """
+        cmd = "etc/init.d/openibd restart"
+        if not process.system(cmd, ignore_status=True, shell=True):
+            return
+        cmd = "ibstat"
+        if not process.system(cmd, ignore_status=True, shell=True):
```
PraveenPenguin: here as well if not `process.system('ibstat', ignore_status=True, shell=True): ` please

io/net/infiniband/mofed_install_test.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        self.iso_location = self.params.get('iso_location', default='')
+        if self.iso_location is '':
+            self.skip("No ISO location given")
+        self.option = self.params.get('option', default='')
+        self.tarball = self.fetch_asset(self.iso_location, expire='10d')
+        cmd = "mount -o loop %s %s" % (self.tarball, self.srcdir)
+        process.run(cmd, shell=True)
+        self.pwd = os.getcwd()
+        os.chdir(self.srcdir)
```
narasimhan-v: This is needed, as we need to execute the command from this location.

io/net/infiniband/mofed_install_test.py
```diff
@@ -0,0 +1,92 @@
...
+        cmd = './mlnxofedinstall %s --force' % self.option
+        if process.system(cmd, ignore_status=True, shell=True):
+            self.fail("Install Failed with %s" % self.option)
+
+    def uninstall(self):
+        """
+        Uninstalls MOFED, if installed fine.
+        """
+        cmd = "etc/init.d/openibd restart"
+        if not process.system(cmd, ignore_status=True, shell=True):
```
narasimhan-v: I am using the variable `cmd` only everywhere. So, I do not think it is a problem

io/net/infiniband/mofed_install_test.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def setUp(self):
+        """
+        Mount MOFED iso.
+        """
+        self.iso_location = self.params.get('iso_location', default='')
+        if self.iso_location is '':
+            self.skip("No ISO location given")
+        self.option = self.params.get('option', default='')
+        self.tarball = self.fetch_asset(self.iso_location, expire='10d')
```
narasimhan-v: Sure, will change it.

io/net/infiniband/mofed_install_test.py
PraveenPenguin: sure can you use this change dir in other helper  function  as there is no command executed after change dir

```
       self.option = self.params.get('option', default='')
        self.tarball = self.fetch_asset(self.iso_location, expire='10d')
        cmd = "mount -o loop %s %s" % (self.tarball, self.srcdir)
        process.run(cmd, shell=True)
        self.pwd = os.getcwd()
        os.chdir(self.srcdir)

def install(self):
```

io/net/infiniband/mofed_install_test.py
PraveenPenguin: sure that make sense but still why use at least once to make garbage collector to do the job 
:)

PraveenPenguin: but again it your's call it is just a suggestion

narasimhan-v: I agree with your point, but I prefer this way where each command is assigned to a variable :-)

io/net/infiniband/mofed_install_test.py
narasimhan-v: Agreed, will push this to install()

PR #74, opened 2016-10-01
Added stutter testsuite Test

PraveenPenguin:
stutter testsuite which run memory Test(memhog)

### DIFF
--- a/memory/stutter.py
+++ b/memory/stutter.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author:Praveen K Pandey <praveen@linux.vnet.ibm.com>
+#
+
+
+import os
+
+from avocado import Test
+from avocado import main
+from avocado.utils import process, archive, memory, build
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Stutter(Test):
+
+    """
+    stutter benchmark
+    """
+
+    def setUp(self):
+        '''
+        Build Stutter Test
+        Source:
+        https://github.com/gaowanlong/stutter/archive/master.zip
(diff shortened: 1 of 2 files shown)


PraveenPenguin:
job log : https://paste.fedoraproject.org/439940/34249514/

memory/stutter.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def test(self):
+
+        os.chdir(self.srcdir)
+
+        ret = process.system('./run.sh', shell=True)
+
+        if ret != 0:
+            self.fail("Test failed")
+
```
narasimhan-v: if `process.system()` returns a non-zero value, the script will error out.
So, use `ignore_status=True` parameter in `process.sytem()`.

Also, you can remove the usage of `ret` and use it as `if process.system('./run.sh', shell=True, ignore_status=True) !=0` :-)

memory/stutter.py
```diff
@@ -0,0 +1,83 @@
...
+        tarball = self.fetch_asset("stutter.zip", locations=locations,
+                                   expire='7d')
+
+        archive.extract(tarball, self.srcdir)
+
+        self.srcdir = os.path.join(self.srcdir, 'stutter-master')
+
+        mem_byte = str(memory.memtotal())
+
```
narasimhan-v: There are many empty lines. Any intentional usage ?

memory/stutter.py
```diff
@@ -0,0 +1,83 @@
...
+        self._memory = self.params.get('runarg', default=mem_byte)
+        self._iteration = self.params.get('runarg', default='10')
+        self._logdir = self.params.get('runarg', default='/root/logdir')
+        self._rundir = self.params.get('runarg', default='/tmp')
+
+        process.run('mkdir -p  %s' % self._logdir)
+
```
narasimhan-v: `os.makedirs()` will do the same

memory/stutter.py
PraveenPenguin: i used it because i think through this code look clean , will remove white spaces where i feel logic should in a block

memory/stutter.py
PraveenPenguin: sure

memory/stutter.py
PraveenPenguin: thanks make sense

PraveenPenguin: sorry i can't use as i want that it won't throw error when dir as present  and not want to use `os.path.isdir`

` no error if existing, make parent directories as needed`
and don't want to delete dir in clean up section  as if test fail user can go and review logs in logdir

PraveenPenguin:
closing this PR as raised new PR #75 with suggested change

PR #76, opened 2016-10-03
Software Raid number of disk fixes

narasimhan-v:
Software raid test used to skip if number of disks were less than 4.
Changed this behaviour, so that if number of disks are applicable for a
certain raid level, it works. Else, test fails for that level.
Updated Readme file accordingly.

### DIFF
--- a/io/disk/softwareraid.py
+++ b/io/disk/softwareraid.py
@@ -51,14 +51,8 @@ class SoftwareRaid(Test):
                 self.skip("Unable to install mdadm")
         cmd = "mdadm -V"
         self.check_pass(cmd, "Unable to get mdadm version")
-        self.disk = self.params.get('disk', default='').strip(" ").split(" ")
-        self.sparedisk = self.disk.pop()
-        self.remadd = ''.join(self.disk[-1:])
+        self.disk = self.params.get('disk', default='').strip(" ")
         self.raidlevel = str(self.params.get('raid', default='0'))
-        self.disk_count = len(self.disk)
-        self.disk = ' '.join(self.disk)
-        if self.disk_count < 4:
-            self.skip("Please give minimum of 5 disk to execute this test case")
 
     def test_run(self):
         """
@@ -73,6 +67,7 @@ class SoftwareRaid(Test):
         """
         Only basic operations are run viz create and delete
         """
+        self.disk_count = len(self.disk.split(" "))
         cmd = "echo 'yes' | mdadm --create --verbose --assume-clean \
             /dev/md/mdsraid --level=%s --raid-devices=%d %s \
             --force" \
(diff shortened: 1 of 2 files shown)


narasimhan-v:
[job_log.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/505903/job_log.txt)

io/disk/softwareraid.py
```diff
@@ -86,6 +81,11 @@ def extensivetest(self):
         Extensive software raid options are run viz create, delete, assemble,
         create spares, remove and add drives
         """
+        self.disk = self.disk.split(" ")
+        self.sparedisk = self.disk.pop()
+        self.remadd = ''.join(self.disk[-1:])
+        self.disk_count = len(self.disk)
+        self.disk = ' '.join(self.disk)
```
PraveenPenguin: why this is required i think it having  same effect when it is in  setup and also it make sense that we can get all variable in test setup ?

io/disk/softwareraid.py
```diff
@@ -51,14 +51,8 @@ def setUp(self):
...
         self.check_pass(cmd, "Unable to get mdadm version")
-        self.disk = self.params.get('disk', default='').strip(" ").split(" ")
-        self.sparedisk = self.disk.pop()
-        self.remadd = ''.join(self.disk[-1:])
+        self.disk = self.params.get('disk', default='').strip(" ")
         self.raidlevel = str(self.params.get('raid', default='0'))
-        self.disk_count = len(self.disk)
-        self.disk = ' '.join(self.disk)
-        if self.disk_count < 4:
-            self.skip("Please give minimum of 5 disk to execute this test case")
```
PraveenPenguin: can we have check for raid level mapping with respect to disk provided  by user ?

io/disk/softwareraid.py
narasimhan-v: For basic test, we do not need spare disk. So, direct disk count is ok.
For advanced levels, we need spare disk. So, after removing spare disk from the main list of disks, we are calculating disk count.

io/disk/softwareraid.py
PraveenPenguin: sure just a thought seems me OK that we already have this in README file 

`Test fails, if number of disks are not applicable for a certain raid level.`

io/disk/softwareraid.py
PraveenPenguin: sure ,it is good that we can setup variable in Setup section it self 
```

 if self.raidlevel == 'linear' or self.raidlevel == '0':
  self.disk_count = len(self.disk.split(" "))
else :
        self.disk = self.disk.split(" ")
        self.sparedisk = self.disk.pop()
        self.remadd = ''.join(self.disk[-1:])
        self.disk_count = len(self.disk)
        self.disk = ' '.join(self.disk)
```

narasimhan-v: Agreed, let me do that and send a new PR

narasimhan-v:
@PraveenPenguin opened a new PR #78

PR #77, opened 2016-10-03
Adding test to nvme-cli test

narasimhan-v:
Added the following tests to nvme-cli test.
*) compare
*) flush
*) writezeroes
*) writeuncorrectable
*) dsm
*) reset
*) subsystemreset

### DIFF
--- a/io/disk/ssd/nvmetest.py
+++ b/io/disk/ssd/nvmetest.py
@@ -106,6 +106,63 @@ class NVMeTest(Test):
         if process.system(cmd, timeout=300, ignore_status=True, shell=True):
             self.fail("Write failed")
 
+    def testcompare(self):
+        """
+        Compares data written on the device with given data.
+        """
+        self.testwrite()
+        cmd = 'echo 1|nvme compare %s -z %d' % (self.id_ns, self.format_size)
+        if process.system(cmd, timeout=300, ignore_status=True, shell=True):
+            self.fail("Compare failed")
+
+    def testflush(self):
+        """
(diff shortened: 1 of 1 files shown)


narasimhan-v:
[job_log.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/505918/job_log.txt)

narasimhan-v:
@PraveenPenguin Agreed, and I have done the same for create namespace, format namespace, etc.
But, the person who is running the test needs to know what all tests are run.

Can you suggest a better way to do that, please ?

PraveenPenguin:
@narasimhan-v yes i saw that , can we pass a string in self.log.info("test run <test>") used in  pass as well in fail case ? 
couple of   request 
1-can  can you move `testread` and `testwrite` also  in helper function  and use it from test function  
2- use expire  flag 
for that we need to use fetch_asset and link can be 
`https://github.com/linux-nvme/nvme-cli/archive/master.zip `

code snippet can be :

```
        locations = ["https://github.com/linux-nvme/nvme-cli/archive/"
                     "master.zip"]
        tarball = self.fetch_asset("nvme-cli.zip", locations=locations,
                                   expire='10d') 
        self.srcdir = os.path.join(self.srcdir, 'nvme-cli-master')
```

PraveenPenguin:
suggesting  using fetch asset  as It can be cached and downloads only the snapshot, not the whole history...

narasimhan-v:
@PraveenPenguin Created new PR #115

PR #101, opened 2016-10-07
SMARTCTL tool tests - v1

manvanthar:
This scripts performs S.M.A.R.T relavent tests using smartctl tool on one or more disks.

[Test_result.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/515770/Test_result.txt)

### DIFF
--- a/io/disk/smartctl.py
+++ b/io/disk/smartctl.py
@@ -52,7 +52,7 @@ class SmartctlTest(Test):
         disk_list = self.params.get('disk').strip(" ")
         self.disk_list = disk_list.split()
         self.to = self.params.get("timeout", default="2400")
-        if self.disk_list is '' or self.option is '':
+        if self.disk_list is [] or self.option is '':
             self.skip(" Test skipped!!, please ensure Block device and options are specified in yaml file")
 
     def test(self):
--- a/io/disk/smartctl.py
+++ b/io/disk/smartctl.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+#
+# Copyright: 2016 IBM
+# Author: Manvanthara B Puttashankar <manvanth@linux.vnet.ibm.com>
+
+
+"""
+smartctl - controls  the Self-Monitoring, Analysis and Reporting 
+Technology (SMART) system built into most ATA/SATA and SCSI/SAS 
+hard drives and solid-state drives
+"""
+
+import time
+from avocado import Test
+from avocado.utils.software_manager import SoftwareManager
+from avocado.utils import process
+from avocado import main
+
+
+class SmartctlTest(Test):
+
+    """
+    This scripts performs S.M.A.R.T relavent tests using smartctl tool 
+    on one or more disks
+    """
+
+    def setUp(self):
+
+        """
+        Checking if the required packages are installed,
+        if not found packages will be installed
+        """
+
+        smm = SoftwareManager()
+        if not smm.check_installed("smartctl"):
+            self.log.info("smartctl should be installed prior to the test")
+            if SoftwareManager().install("smartmontools") is False:
+                self.skip("Unable to install smartctl")
+        self.option = self.params.get('option')
+        disk_list = self.params.get('disk').strip(" ")
+        self.disk_list = disk_list.split()
+        self.to = self.params.get("timeout", default="2400")
+        if self.disk_list is '' or self.option is '':
+            self.skip(" Test skipped!!, please ensure Block device and options are specified in yaml file")
+
+    def test(self):
+        """
+        executes S.M.A.R.T options using smartctl tool 
+        """
+        for disks in self.disk_list:
+            self.log.info(" Trying with Disk %s" % (disks))
+            if self.option == "--test=long":
+                # Self Test will be run for 2 min
+                cmd = "timeout %s smartctl %s %s; sleep 120; smartctl -X %s; smartctl -A %s" % ( self.to, self.option, disks, disks, disks )
+            else:
+                cmd = "timeout %s smartctl %s %s" % ( self.to, self.option, disks )
+            if process.system(cmd, ignore_status=True, shell=True):
+                # TODO: tool can be improved to validate the disk data for different vendors
+                self.fail("Smartctl option %s FAILS" % self.option)
+
+if __name__ == "__main__":
+    main()
--- a/io/disk/smartctl.py.data/Readme
+++ b/io/disk/smartctl.py.data/Readme
@@ -1,0 +1,11 @@
+smartctl - controls  the Self-Monitoring, Analysis and Reporting Technology (SMART) system built into most ATA/SATA and SCSI/SAS hard drives and solid-state drives 
+
+Values to be passed in yaml file:
+
+* Name of the devices on which this scripts should be run.
+
+* smartctl tool options.
+
+Note: 
+1. Please do not input OS drives. 
+2. Logical block device drives do not support SMART.
--- a/io/disk/smartctl.py.data/smartctl.yaml
+++ b/io/disk/smartctl.py.data/smartctl.yaml
@@ -1,0 +1,68 @@
+scenario:
+    # please ensure to provide one or more non-root block device 
+    disk: /dev/sdb /dev/sdc
+Options: !mux
+    self_test:
+        option: --test=long
+    info:
+        option: -i
+    device_setting:
+        option: -g all
+    scan_drives:
+        option: --scan
+    scan_drives_open:
+        option: --scan-open
+    quietmode:
+        option: -q errorsonly
+    quietmode:
+        option: -q silent
+    bad_checksum:
+        option: -b warn
+    report_transaction:
+        option: -r ioctl
+    check_mode:
+        option: -n sleep 
+    automatic_offline_off:
+        option: -o off
+    automatic_offline_on:
+        option: -o on
+    autosave_attribute_off:
+        option: -S off
+    autosave_attribute_on:
+        option: -S on
+    device_setting:
+        option: -s on
+    tolerance_opt1:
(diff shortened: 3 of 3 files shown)


io/disk/smartctl.py
```diff
@@ -0,0 +1,74 @@
...
+        if not smm.check_installed("smartctl"):
+            self.log.info("smartctl should be installed prior to the test")
+            if SoftwareManager().install("smartmontools") is False:
+                self.skip("Unable to install smartctl")
+        self.option = self.params.get('option')
+        disk_list = self.params.get('disk').strip(" ")
+        self.disk_list = disk_list.split()
+        self.to = self.params.get("timeout", default="2400")
+        if self.disk_list is [] or self.option is '':
+            self.skip(" Test skipped!!, please ensure Block device and options are specified in yaml file")
```
narasimhan-v: please have a check on each line having less than 80 chars. pep8 or pylint will catch it.

PraveenPenguin:
Could you please squash the commits into one commit only?

PraveenPenguin:
please make script as pylint or  pep8 compliant it will fix  travis

io/disk/smartctl.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def test(self):
+        """
+        executes S.M.A.R.T options using smartctl tool 
+        """
+        for disks in self.disk_list:
+            self.log.info(" Trying with Disk %s" % (disks))
+            if self.option == "--test=long":
+                # Self Test will be run for 2 min
+                cmd = "timeout %s smartctl %s %s; sleep 120; smartctl -X %s; smartctl -A %s" % ( self.to, self.option, disks, disks, disks )
```
narasimhan-v: In case of running multiple commands in a single `proces.system()`, only the last command's return code is caught.
So, running a single command at a time is better

io/disk/smartctl.py
```diff
@@ -0,0 +1,74 @@
...
+        self.disk_list = disk_list.split()
+        self.to = self.params.get("timeout", default="2400")
+        if self.disk_list is [] or self.option is '':
+            self.skip(" Test skipped!!, please ensure Block device and options are specified in yaml file")
+
+    def test(self):
+        """
+        executes S.M.A.R.T options using smartctl tool 
+        """
+        for disks in self.disk_list:
```
narasimhan-v: Just a suggestion, if you can multiplex disks also, this for loop can be eliminated, and test will run for n times for n disks. See which suits your requirement better.

io/disk/smartctl.py.data/Readme
```diff
@@ -0,0 +1,11 @@
+smartctl - controls  the Self-Monitoring, Analysis and Reporting Technology (SMART) system built into most ATA/SATA and SCSI/SAS hard drives and solid-state drives 
+
+Values to be passed in yaml file:
+
+* Name of the devices on which this scripts should be run.
+
+* smartctl tool options.
+
+Note: 
+1. Please do not input OS drives.
```
narasimhan-v: Can you add a check in the test for this, and skip in such cases ?
`df -h /boot` helps here

io/disk/smartctl.py
```diff
@@ -0,0 +1,74 @@
...
+            self.log.info(" Trying with Disk %s" % (disks))
+            if self.option == "--test=long":
```
narasimhan-v: Would it not be better to have it as if `"--test=long" in self.option`,  so that all options which has --test=long can be handled by this case

io/disk/smartctl.py.data/Readme
manvanthar: hmm ok will do it.

io/disk/smartctl.py
manvanthar: there is only one --test-long option,  please refer the multiplexer,  in future no enhancement will be made by anyone to test multiple times the same option.

io/disk/smartctl.py
manvanthar: sure narasimhan, will do it

io/disk/smartctl.py
```diff
@@ -0,0 +1,74 @@
...
+
+        """
+        Checking if the required packages are installed,
+        if not found packages will be installed
+        """
+
+        smm = SoftwareManager()
+        if not smm.check_installed("smartctl"):
+            self.log.info("smartctl should be installed prior to the test")
+            if SoftwareManager().install("smartmontools") is False:
```
PraveenPenguin: can `smm` used instead of `SoftwareManager() `as you already created a reference of `SoftwareManager()`

io/disk/smartctl.py
```diff
@@ -0,0 +1,74 @@
...
+                self.skip("Unable to install smartctl")
+        self.option = self.params.get('option')
+        disk_list = self.params.get('disk').strip(" ")
```
PraveenPenguin: can this be as 

` self.disk_list = self.params.get('disk').strip(" ").split()`

 as for extra variable garbage collector need to clear reserved memory

io/disk/smartctl.py
manvanthar: suggestion looks great narasimhan, do you have any example in mind ?  I would really like to explore

wouldn't mutiplexing disks consider it as an seperate iterations rather than n times * n disks.

io/disk/smartctl.py
```diff
@@ -0,0 +1,74 @@
...
+        self.disk_list = disk_list.split()
+        self.to = self.params.get("timeout", default="2400")
```
PraveenPenguin: just curious about is time out will change in any case please help me ?

io/disk/smartctl.py
PraveenPenguin: @manvanthar problem here is if this `smartctl -A 'disk'` pass and previous command fail so proces.system() will return true that i think not expected 

can it be like 

```
  if process.system('smartctl %s %s'% (self.option, disks),ignore_status=True, shell=True,timeout=self.to):
      time.sleep(30)
      if process.system('smartctl -X %s'% disks,ignore_status=True, shell=True,timeout=self.to): 
         if  process.system('smartctl -A %s'% disks,ignore_status=True, shell=True,timeout=self.to) :
                 self.fail("Smartctl option %s FAILS" % self.option)

```

io/disk/smartctl.py.data/Readme
```diff
@@ -0,0 +1,11 @@
+smartctl - controls  the Self-Monitoring, Analysis and Reporting Technology (SMART) system built into most ATA/SATA and SCSI/SAS hard drives and solid-state drives 
+
+Values to be passed in yaml file:
+
```
PraveenPenguin: please remove this white space

io/disk/smartctl.py.data/Readme
```diff
@@ -0,0 +1,11 @@
...
+* Name of the devices on which this scripts should be run.
+
```
PraveenPenguin: here also please

io/disk/smartctl.py.data/Readme
```diff
@@ -0,0 +1,11 @@
...
+* smartctl tool options.
+
```
PraveenPenguin: here also please

io/disk/smartctl.py.data/Readme
```diff
@@ -0,0 +1,11 @@
+smartctl - controls  the Self-Monitoring, Analysis and Reporting Technology (SMART) system built into most ATA/SATA and SCSI/SAS hard drives and solid-state drives
```
PraveenPenguin: please try to wrap this in 80 Char

io/disk/smartctl.py
manvanthar: sorry folks, i should have been more clear,  going one step inside to technicality

--test-long runs the SMART tests but puts the test into background.  so irrespective of it passed or failed , the return is allways 0,    the test can take upto 2 days depending on the size

-X aborts the test in 2 min..

the actual output of interest is -A.

sure i can change the code to comply with the coding standards. i will get the changes in -v2

io/disk/smartctl.py
manvanthar: sure will do.

io/disk/smartctl.py
manvanthar: changed pushed to -v2

io/disk/smartctl.py
PraveenPenguin: sure

io/disk/smartctl.py
PraveenPenguin: @manvanthar  sure it's your call as this yours test :)

apahim:
Let me give some feedback here as well. To provide a list using yaml, one can write:

```yaml
scenario:
    disk:
        - /dev/sdb
        - /dev/sdc
```

This will result in the list `['/dev/sdb', '/dev/sdc']` inside the test. Anyway, I agree you should remove the loop and use the multiplex for it. For the lengthy logs, you have a number of ways to handle that. You can, for example, create your own log stream:

```python
import logging
progress_log = logging.getLogger('progress')
```
Use it where you want to:
```python
...
            progress_log.error('test failed')
            # The self.fail() is still needed to set the test status to FAIL,
            # since you're using ignore_status=True
            self.fail("Smartctl option %s FAILS" % self.option)
        #else:
        #    progress_log.info('test ok')
...
```

And then show only the messages generated by your custom log stream:
```
avocado --show progress run smartctl.py -m smartctl.py.data/smartctl.yaml
```
Reference: http://avocado-framework.readthedocs.io/en/latest/LoggingSystem.html

I see you have a `self.fail()` inside the loop, meaning you're stooping the test on first failed `cmd` execution. When using the multiplex to create the loop over the disks, you can use the `--failfast` option to have the similar behaviour:

```
avocado --show progress run smartctl.py -m smartctl.py.data/smartctl.yaml --failfast
```

Another point where you could improve is in the `cmd` construction. You can start with the basic string and add options when it applies:
```python
        cmd = "timeout %s smartctl %s %s;" % (self.to, self.option, self.disk)
        if self.option == "--test=long":
            cmd += " sleep 120; smartctl -X %s; smartctl -A %s;" % (self.disk, self.disk)
```

Last but not least, assuming you're using `vim`, you can add in your `~/.vimrc`:
```vim
"Show a column at line 80, so you know when you're crossing
"the pep8 line width limit.
set colorcolumn=80
```

io/disk/smartctl.py
apahim: You also consider just to replace the `;` by `&&`. This will stop the chain of commands on first error and return the correspondent failed command exit code.

io/disk/smartctl.py
narasimhan-v: @manvanthar Agreed

PraveenPenguin:
@manvanthar ping

PR #119, opened 2016-10-14
Enhance libhugetest

PraveenPenguin:
*added patch file  as well logic for RHEL and fedora 24 which
enable test suite able to run in RHEL/Fedora
*logic for catch 31 bit 64 bit test failure

### DIFF
--- a/memory/libhugetlbfs.py
+++ b/memory/libhugetlbfs.py
@@ -31,6 +31,7 @@ from avocado.utils import distro
 
 
 class libhugetlbfs(Test):
+
     def setUp(self):
         # Check for root permission
         if os.geteuid() != 0:
@@ -98,6 +99,13 @@ class libhugetlbfs(Test):
         os.chdir(self.srcdir)
         patch = self.params.get('patch', default='elflink.patch')
         process.run('patch -p1 < %s' % data_dir + '/' + patch, shell=True)
+
+        if (detected_distro.name == "redhat" or
+                detected_distro.name == "fedora"):
+            falloc_patch = 'patch -p1 < %s ' % (
+                os.path.join(data_dir, 'falloc.patch'))
+            process.run(falloc_patch, shell=True)
+
         build.make(self.srcdir, extra_args='BUILDTYPE=NATIVEONLY')
 
     def test(self):
--- a/memory/libhugetlbfs.py
+++ b/memory/libhugetlbfs.py
@@ -19,6 +19,7 @@
 
 import os
 import glob
+
 from avocado import Test
 from avocado import main
 from avocado.utils import process
@@ -108,9 +109,28 @@ class libhugetlbfs(Test):
 
         build.make(self.srcdir, extra_args='BUILDTYPE=NATIVEONLY')
 
+    def fail_check(self, log, check):
+
+        if check == 1:
+            env = "32"
+        else:
+            env = "64"
+
+        for line in log.splitlines():
+            if ('FAIL:' in line) and ('Expected' not in line):
+                if (line.split('FAIL:')[1].split('   ')[check].strip()) > 0:
+                    self.fail(" hugpage page " + env + " bit test failed")
+
     def test(self):
         os.chdir(self.srcdir)
-        build.make(self.srcdir, extra_args='BUILDTYPE=NATIVEONLY check')
+        result = build.run_make(
+            self.srcdir, extra_args='BUILDTYPE=NATIVEONLY check')
+
+        # 64 bit hugepage test check
+        self.fail_check(result.stdout, 3)
+
+        # 32 bit hugepage test check
+        self.fail_check(result.stdout, 1)
 
     def tearDown(self):
         if self.hugetlbfs_dir:
--- a/memory/libhugetlbfs.py.data/falloc.patch
+++ b/memory/libhugetlbfs.py.data/falloc.patch
@@ -1,0 +1,33 @@
+diff -ru a/tests/fallocate_align.c b/tests/fallocate_align.c
+--- a/tests/fallocate_align.c	2016-10-13 09:23:26.083992061 -0500
++++ b/tests/fallocate_align.c	2016-10-13 09:25:16.145417545 -0500
+@@ -24,6 +24,7 @@
+ #include <signal.h>
+ #include <sys/mman.h>
+ #include <fcntl.h>
++#include <linux/falloc.h>
+ 
+ #include <hugetlbfs.h>
+ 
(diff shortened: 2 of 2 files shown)


PraveenPenguin:
job log : https://paste.fedoraproject.org/449855/27297147/

memory/libhugetlbfs.py
```diff
@@ -98,11 +100,37 @@ def setUp(self):
         os.chdir(self.srcdir)
         patch = self.params.get('patch', default='elflink.patch')
         process.run('patch -p1 < %s' % data_dir + '/' + patch, shell=True)
+
+        if (detected_distro.name == "redhat" or
+                detected_distro.name == "fedora"):
+            falloc_patch = 'patch -p1 < %s ' % (
+                os.path.join(data_dir, 'falloc.patch'))
+            process.run(falloc_patch, shell=True)
```
narasimhan-v: Is it fine if code errors out in case of failure to apply patch ?

memory/libhugetlbfs.py
```diff
@@ -98,11 +100,37 @@ def setUp(self):
...
+
         build.make(self.srcdir, extra_args='BUILDTYPE=NATIVEONLY')
 
+    def fail_check(self, log, check):
+
+        if check == 1:
+            env = "32"
+        else:
```
narasimhan-v: Should it not be `elif check == 3:` ?
Correct me if I am wrong here

memory/libhugetlbfs.py
```diff
@@ -98,11 +100,37 @@ def setUp(self):
...
         os.chdir(self.srcdir)
-        build.make(self.srcdir, extra_args='BUILDTYPE=NATIVEONLY check')
+        result = build.run_make(
+            self.srcdir, extra_args='BUILDTYPE=NATIVEONLY check')
+
+        # 64 bit hugepage test check
+        self.fail_check(result.stdout, 3)
+
+        # 32 bit hugepage test check
+        self.fail_check(result.stdout, 1)
```
narasimhan-v: Just a suggestion:
Why can't the `fail_check()` be called only once and checks for both 32 bit and 64 bit errors ?

PraveenPenguin: thanks sure that make sense

memory/libhugetlbfs.py
```diff
@@ -98,11 +100,37 @@ def setUp(self):
...
+
+        if (detected_distro.name == "redhat" or
+                detected_distro.name == "fedora"):
+            falloc_patch = 'patch -p1 < %s ' % (
+                os.path.join(data_dir, 'falloc.patch'))
+            process.run(falloc_patch, shell=True)
+
         build.make(self.srcdir, extra_args='BUILDTYPE=NATIVEONLY')
 
+    def fail_check(self, log, check):
```
apahim: Since this is an enhancement pull request, let me suggest something here. It took some time for me to understand this function. Instead of a `fail_check()` that digs into the log to find the errors, I'd create a log parser so you can easily verify the results inside your `test()`. For this log results, something like this:

```python

def _log_parser(self, log):
    """
    Parses the log, returning a dictionary with the test results.

    Test summary section example:
        ********** TEST SUMMARY
        *                      16M          
        *                      32-bit 64-bit
        *     Total testcases:     0     93  
        *             Skipped:     0      0  
        *                PASS:     0     90  
        *                FAIL:     0      3  
        *    Killed by signal:     0      0  
        *   Bad configuration:     0      0  
        *       Expected FAIL:     0      0  
        *     Unexpected PASS:     0      0  
        * Strange test result:     0      0  
        **********

    Return example:
        {32: {'Bad': 0,
              'Expected': 0,
              'FAIL': 0,
              'Killed': 0,
              'PASS': 0,
              'Skipped': 0,
              'Strange': 0,
              'Total': 0,
              'Unexpected': 0},
         64: {'Bad': 0,
              'Expected': 0,
              'FAIL': 3,
              'Killed': 0,
              'PASS': 90,
              'Skipped': 0,
              'Strange': 0,
              'Total': 93,
              'Unexpected': 0}}
    """
    section = False
    parsed_results = {32: {},64: {}}

    for line in log.splitlines():
        if line == '********** TEST SUMMARY\n':
            section = True
        if line == '**********\n':
            section = False
        if section and  ':' in line:
            key, values = line.split(':')
            parsed_results[32][key.split()[1]] = int(values.split()[0])
            parsed_results[64][key.split()[1]] = int(values.split()[1])

    return parsed_results

def test(self):
    ...
    ...
    results = build.run_make(self.srcdir, extra_args='BUILDTYPE=NATIVEONLY check')
    parsed_results = self._log_parser(results.stdout)
    error = None
    if parsed_results[32]['FAIL']:
        error = "%s tests failed for 32-bit\n" % parsed_results[32]['FAIL'])
    if parsed_results[64]['FAIL']:
        error += "%s tests failed for 64-bit" % parsed_results[64]['FAIL'])
   if error:
        self.fail(error)
```

narasimhan-v: Just minor improvements.

`if line == '********** TEST SUMMARY\n':` should be `if line == '********** TEST SUMMARY':`

and

`parsed_results[32][key.split()[1]] = int(values.split()[0])`
can be 
`parsed_results[32][key.split("*")[1].strip()] = int(values.split()[0])`

PraveenPenguin: @narasimhan-v yes i know working on this

narasimhan-v: @PraveenPenguin cool.

apahim: Yep. Not tested code, provided only as reference. Thanks.
You can also include a `continue` just bellow the `section = True` and the `section = False`. This will improve a little bit the performance.

apahim: Btw, insted of:

`parsed_results[32][key.split("*")[1].strip()] = int(values.split()[0])`

it can be like:

`parsed_results[32][key.lstrip('* ')] = int(values.split()[0])`

Easier to understand :)

PraveenPenguin:
@apahim  @narasimhan-v  thanks 

raised new PR #120 so closing this

memory/libhugetlbfs.py
PraveenPenguin: without this also it works as 

parsed_results[64][key.split()[1]] = int(values.split()[1]) 
thanks

apahim: @PraveenPenguin works, but drops part of the key name. `Killed by signal` becomes `Killed`. That's what @narasimhan-v pointed out.

PraveenPenguin: @apahim oh miss that @narasimhan-v thanks

PR #120, opened 2016-10-14
Enhance libhugetest v2

PraveenPenguin:
```
*added patch file as well logic for RHEL and fedora 24 which
enable test suite able to run in RHEL/Fedora
*logic for catch 32 bit 64 bit test failure
```


PraveenPenguin:
job log : https://paste.fedoraproject.org/450084/14764432/

apahim:
Great. Thank you!

PR #129, opened 2016-10-23
Added perftool-testsuite test

PraveenPenguin:
Added test which run perftool-testsuite and notify
user when test fail

### DIFF
--- a/perf/perftool.py
+++ b/perf/perftool.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author:Praveen K Pandey <praveen@linux.vnet.ibm.com>
+#
+
+
+import os
+
+from avocado import Test
+from avocado import main
+from avocado.utils import archive, build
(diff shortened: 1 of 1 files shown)


PraveenPenguin:
job log : https://paste.fedoraproject.org/458457/20522314/

perf/perftool.py
```diff
@@ -0,0 +1,61 @@
...
+        for package in ['gcc', 'make', 'perf']:
+            if not smm.check_installed(package) and not smm.install(package):
+                self.error('%s is needed for the test to be run' % package)
+
+        locations = ["https://github.com/rfmvh/perftool-testsuite/archive/"
+                     "master.zip"]
+        tarball = self.fetch_asset("perftool.zip", locations=locations,
+                                   expire='7d')
+        archive.extract(tarball, self.srcdir)
+        self.srcdir = os.path.join(self.srcdir, 'perftool-testsuite-master')
```
narasimhan-v: Are you sure you can change `self.srcdir` and not have any effects in any test ? Please clarify

perf/perftool.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def test(self):
+
+        for line in build.run_make(self.srcdir, extra_args='check',
+                                   ignore_status=True).stdout.splitlines():
+            if 'FAIL' in line:
+                self.fail("Test failed")
```
PraveenPenguin: not very sure about this i think log will have all the failure as user can any why need to go through the log , what you think ?

perf/perftool.py
narasimhan-v: @PraveenPenguin I agree with this, but as a general practice, is it right ? I see many tests doing this recently..

PraveenPenguin: @narasimhan-v i think so should be ok

perf/perftool.py
PraveenPenguin: sure will do

PraveenPenguin:
raised PR #130 with suggested change

PR #135, opened 2016-10-25
IP Over IB Test

prudmirya:
IP Over IB Test

IPoIB can run over two infiniband transports, Unreliable Datagram (UD) mode or Connected mode (CM)

Signed-off-by: prudhvi <mprudhvi@linux.vnet.ibm.com>

### DIFF
--- a/io/net/infiniband/ip_over_ib.py
+++ b/io/net/infiniband/ip_over_ib.py
@@ -1,0 +1,205 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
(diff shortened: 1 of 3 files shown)


prudmirya:
[job.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/550831/job.txt)
log added

PraveenPenguin:
still confused about root privileges @prudmirya can you please help me on this

narasimhan-v:
@PraveenPenguin since the test needs to `scp` the perf tools to remote machine, and `configure` on both machines, we need 'root' access on both machines.

PraveenPenguin:
@narasimhan-v my point that by default ubuntu have non root user having root privileges there is extra step that user need to take care 
sure if test is like that so i am ok with it

PR #148, opened 2016-11-08
Test case for lshw tool (Hardware lister)

ramyabs:
Description:
lshw is a small tool to provide detailed informaton on the hardware
configuration of the machine. It can report exact memory
configuration, firmware version, mainboard configuration, CPU version
and speed, cache configuration, bus speed, etc. on DMI-capable x86
systems and on some PowerPC machines (PowerMac G4 is known to work).

Signed-off-by: Ramya BS ramya@linux.vnet.ibm.com
Tested-by : Ramya BS <ramya@linux.vnet.ibm.co

### DIFF
--- a/generic/lshw.py
+++ b/generic/lshw.py
@@ -1,0 +1,171 @@
+#!/usr/bin/env python
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM.
+# Author: Ramya BS <ramya@linux.vnet.ibm.com>
+
+import re
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Lshwrun(Test):
+
+    """
+    lshw is a small tool to extract detailed information on the
+    hardware configuration of the machine.
+    It can report exact memory configuration,firmware version,
+    mainboard configuration, CPU version and speed,cache configuration, bus
+    speed, etc. on DMI-capable x86 or IA-64 systems and on some PowerPC
+    machines (PowerMac G4 is known to work).
+    """
+    interface = process.system_output("ip route show")
+    active_interface = re.search(
+        r"default via\s+\S+\s+dev\s+(\w+)\s+proto", interface).group(1)
+
+    def run_cmd(self, cmd):
+        if process.system(cmd, ignore_status=True, sudo=True, shell=True):
+            self.is_fail += 1
+            self.fail_cmd = cmd
+        return
+
+    def setUp(self):
+        sm = SoftwareManager()
+        for package in ("lshw", "net-tools", "iproute", "pciutils"):
+            if not sm.check_installed(package) and not sm.install(package):
+                self.error("Fail to install %s required for this"
+                           " test." % package)
+
+    def test_lshw(self):
+        """
+        lshw without any options would generate full information
+        report about all detected hardware.
+        """
+        self.log.info("===============Executing lshw test ===="
+                      "===========")
+        self.is_fail = 0
+        self.run_cmd("lshw")
+        self.run_cmd("lshw -version")
+        if self.is_fail >= 1:
+            self.fail("%s command(s) failed to execute  "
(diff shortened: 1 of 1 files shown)


generic/lshw.py
```diff
@@ -0,0 +1,177 @@
...
+    machines (PowerMac G4 is known to work).
+    """
+    interface = process.system_output("ip route show")
+    active_interface = re.search(
+        r"default via\s+\S+\s+dev\s+(\w+)\s+proto", interface).group(1)
+    is_fail = 0
+
+    def run_cmd(self, cmd):
+        cmd_result = process.run(cmd, ignore_status=True, sudo=True,
+                                 shell=True)
```
PraveenPenguin: can please use process.system(cmd, ignore_status=True, sudo=True,  shell=True) it return exit code ?

code can be 

```
if process.system(cmd, ignore_status=True, sudo=True,shell=True):
   self.is_fail += 1
```

generic/lshw.py
```diff
@@ -0,0 +1,177 @@
...
+        if cmd_result.exit_status != 0:
+            self.is_fail += 1
+        return
+
+    def setUp(self):
+        if os.geteuid() != 0:
+            self.error("This program requires super user priv to run.")
```
PraveenPenguin: i think this is not required as you are using `sudo=True` as param while running command

generic/lshw.py
```diff
@@ -0,0 +1,177 @@
...
+        soft = SoftwareManager()
```
PraveenPenguin: i think `soft` as variable name not looks good can please change it to some valid name ?

generic/lshw.py
```diff
@@ -0,0 +1,177 @@
...
+        self.run_cmd("lshw -short  -class network")
+        self.run_cmd("lshw -short  -class storage")
+        self.run_cmd("lshw -short  -class memory")
+        self.run_cmd("lshw -short  -class power")
+        self.run_cmd("lshw -short  -class bus")
+        self.run_cmd("lshw -short  -class processor")
+        self.run_cmd("lshw -short  -class system")
+        if self.is_fail >= 1:
+            self.fail("%s command(s) failed to execute "
+                      "verification" % self.is_fail)
```
PraveenPenguin: it will good if we show user which command is fail for that `self.fail() `can be use in` run_cmd` helper function what you think ?

generic/lshw.py
```diff
@@ -0,0 +1,177 @@
...
+        if get_bus_info_act_inter not in get_bus_info_lspci:
+            self.fail("Verification of network bus info failed ")
+
+    def test_sanitize(self):
+        """
+        sanitize output(remove sensitive information like serial numbers,etc.)
+        """
+        out = []
+        out_lshw = subprocess.Popen("lshw", stdout=subprocess.PIPE,
+                                    stderr=subprocess.PIPE, shell=True)
```
PraveenPenguin: can process.run will be used 

like 
`out_lshw = process.run("lshw",shell=True)`

generic/lshw.py
```diff
@@ -0,0 +1,177 @@
...
+    lshw is a small tool to extract detailed information on the
+    hardware configuration of the machine.
+    It can report exact memory configuration,firmware version,
+    mainboard configuration, CPU version and speed,cache configuration, bus
+    speed, etc. on DMI-capable x86 or IA-64 systems and on some PowerPC
+    machines (PowerMac G4 is known to work).
+    """
+    interface = process.system_output("ip route show")
+    active_interface = re.search(
+        r"default via\s+\S+\s+dev\s+(\w+)\s+proto", interface).group(1)
```
narasimhan-v: This can be added to avocado utils, since others could use it as well. Please consider

generic/lshw.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def test_lshw_verification(self):
+        """
+        compare the output of lshw with other tools
+        which produces similar info of hardware.
+        """
+        # verifying mac address
+        get_mac = process.system_output(
+            " ip link show %s " % self.active_interface)
+        mac = re.search(r'link\/\ether (.*) brd', get_mac).group(1)
```
narasimhan-v: This can be added to avocado utils, since others could use it as well. Please consider

generic/lshw.py
```diff
@@ -0,0 +1,177 @@
...
+        lshw_out_mac = process.system_output("lshw")
+        if mac not in lshw_out_mac:
+            self.fail("lshw failed to show correct mac address")
+
+        # verify network
+        lshw_out_net = process.system_output("lshw -class network")
+        if self.active_interface not in lshw_out_net:
```
narasimhan-v: `if self.active_interface not in process.system_output("lshw -class network"):`

PraveenPenguin:
@ramyabs i think new version of this test is raised as #209 if yes please close this PR

PR #154, opened 2016-11-09
Added tbench test - v2

poojabs7:
Added tbench test that produces only the TCP and process load.

Signed-off-by: poojabs <pooja@oc2013156016.ibm.com>

### DIFF
--- a/perf/tbench.py
+++ b/perf/tbench.py
@@ -1,0 +1,87 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author:  Pooja B Surya <pooja@linux.vnet.ibm.com>
+#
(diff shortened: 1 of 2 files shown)


perf/tbench.py
```diff
@@ -0,0 +1,87 @@
...
+        # should support separate I suppose, but nobody uses it
+        nprocs = self.params.get('nprocs', default=None)
+        args = self.params.get('args',  default=None)
+        if not nprocs:
+            nprocs = commands.getoutput("nproc")
+        args = args + ' %s' % nprocs
+
+        pid = os.fork()
+        if pid:                         # parent
+            client = self.srcdir + '/client.txt'
```
PraveenPenguin: can you please use `os.path.join` ?

perf/tbench.py
```diff
@@ -0,0 +1,87 @@
...
+            args = '-c ' + client + ' ' + '%s' % args
```
PraveenPenguin: string add not looks good using `+`

it can be like 
`args = '-c %s %s'  % (client,args)`

perf/tbench.py
```diff
@@ -0,0 +1,87 @@
...
+            cmd = os.path.join(self.srcdir, "tbench") + " " + args
+            # Standard output is verbose and merely makes our debug logs huge
+            # so we don't retain it.  It gets parsed for the results.
+            self.results = process.system_output(cmd, shell=True)
+            os.kill(pid, signal.SIGTERM)    # clean up the server
+        else:                           # child
+            server = self.srcdir + '/tbench_srv'
```
PraveenPenguin: here also please use `os.path.join`

poojabs7:
Raised new PR 
https://github.com/avocado-framework-tests/avocado-misc-tests/pull/156

PR #178, opened 2016-11-18
arcconf configuration utility - v2

manvanthar:
command line interface tool to configure PMC-Sierra controller

- Includes the fixes requested in v1.
- PCI IDs correct updated to match PMC cards.
- Physical drive gets formated before each RAID creation.

### DIFF
--- a/io/disk/arcconf/arcconf_raid_oper.py
+++ b/io/disk/arcconf/arcconf_raid_oper.py
@@ -1,0 +1,232 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+#
+# Copyright: 2016 IBM
+# Author: Manvanthara B Puttashankar <manvanth@linux.vnet.ibm.com>
+
+
+"""
+arcconf - Array configuration utility for PMC-Sierra
+(Microsemi) Controllers.
+"""
+
+import time
+from avocado import Test
+from avocado.utils.software_manager import SoftwareManager
+from avocado.utils import process
+from avocado import main
+
+
+class Arcconftest(Test):
+
+    """
+    Covers functionality relavent to RAID operations like creation,
+    deletion, Physical & logical disk format and status information.
+    """
+
+    def setUp(self):
+
+        """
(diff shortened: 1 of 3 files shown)


manvanthar:
[job.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/600897/job.txt)
[results.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/600898/results.txt)

PraveenPenguin:
@manvanthar looks good only small issue cmd1,cmd2,cmd3,cmd4 can be cmd as extra variable make python garbage collector to invoke please have a look

io/disk/arcconf/arcconf_raid_oper.py
```diff
@@ -0,0 +1,232 @@
...
+        self.check_pass(cmd1, "Failed to create filesystem")
+
+        cmd2 = "mount %s /mnt" % drive
+        self.check_pass(cmd2, "Failed to mount filesystem")
+
+        cmd3 = "cd /mnt && dd if=/dev/random of=Gfile.txt bs=3M count=1"
+        self.check_pass(cmd3, "Failed to create file")
+
+        cmd4 = "cd / && umount %s" % drive
+        self.check_pass(cmd4, "Failed to unmount file")
```
PraveenPenguin: @manvanthar one help i think all four check can be inter related 

like 1- if file system not created it not able to mount 
2- if it not mounted in not able to create any file in mount point 
3- if device not mounted device not able to mount 

can it will handle in if else

```
if (file system create == PASS):
else if (mounted == PASS):
  else if (file write == PASS):
```
this is just a sample please consider

manvanthar:
How about the below one Praveen, are you okay ?   i will get the filesystem type and mount point from yaml

def make_fs(self, fs_type, mount_drv):
    """
    creates filesystem
    """
      cmd = "lsscsi  | grep LogicalDrv | awk \'{print $7}\'"
    drive = self.cmdop_list(cmd)

    if drive != "":
        cmd = "echo y | mkfs.%s %s && mount %s %s && cd %s && dd if=/dev/random of=Gfile.txt bs=3M count=1 && cd / && umount %s" % (fs_type, drive, drive, mount_drv, mount_drv, drive) 
        self.check_pass(cmd, "Failed to create filesystem")
    else:
        self.log.info(" Filesystem creation skipped!!")

narasimhan-v:
@manvanthar @PraveenPenguin As an addition, please make sure tearDown() is used in this case, like *) when mounting is failed, clear the created filesystem before exiting from script.
*) when doing IO fails, umount and clear the created filesystembefore exiting from script.

io/disk/arcconf/arcconf_raid_oper.py
manvanthar: @narasimhan-v sure.

PraveenPenguin:
@manvanthar @narasimhan-v looks good

PR #218, opened 2016-12-15
Program to stress filesystem

hathyaga:
Program to stress filesystem

Running filesystem related tests using LTP test suite

### DIFF
--- a/io/disk/ltp_fs.py
+++ b/io/disk/ltp_fs.py
@@ -1,0 +1,118 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Harsha Thyagaraja <harshkid@linux.vnet.ibm.com>
+#
+# Based on code by Martin Bligh <mbligh@google.com>
+# copyright 2006 Google, Inc.
+# https://github.com/autotest/autotest-client-tests/tree/master/ltp
+
+
+import os
+from avocado import Test
+from avocado import main
+from avocado.utils import build
+from avocado.utils import process, archive
+from avocado.utils.software_manager import SoftwareManager
+
+
+class ltp(Test):
+    '''
+    LTP (Linux Test Project) testsuite
+    '''
+    def setUp(self):
+        '''
+        To check and install dependencies for the test
+        '''
+        sm = SoftwareManager()
+        deps = ['gcc', 'make', 'automake', 'autoconf']
+        for package in deps:
+            if not sm.check_installed(package) and not sm.install(package):
+                self.error(package + ' is needed for the test to be run')
+        self.disk = self.params.get('disk')
+        self.mount_point = self.params.get('mount_point', default=self.srcdir)
+        self.script = self.params.get('script')
+        self.fs = self.params.get('fs', default='ext4')
+        self.args = self.params.get('args', default='').split(',')
+        url = "https://github.com/linux-test-project/ltp/\
+               archive/master.zip"
+        tarball = self.fetch_asset("ltp-master.zip",
+                                   locations=[url], expire='7d')
+        archive.extract(tarball, self.srcdir)
+        ltp_dir = os.path.join(self.srcdir, "ltp-master")
+        os.chdir(ltp_dir)
+        build.make(ltp_dir, extra_args='autotools')
+        ltpbin_dir = os.path.join(ltp_dir, 'bin')
+        os.mkdir(ltpbin_dir)
+        process.system('./configure --prefix=%s' % ltpbin_dir)
+        build.make(ltp_dir)
+        build.make(ltp_dir, extra_args='install')
+
+    def test(self):
+        for val in self.args:
+            self.log.info("creating ext4 filesystem on disk %s" % self.disk)
+            create_fs = "mkfs.%s %s" % (self.fs, self.disk)
+            self.log.info("%s" % create_fs)
+            ret = process.system(create_fs, shell=True)
+            if ret != 0:
+                self.fail("Created filesystem on %s failed" % self.disk)
+            self.log.info("Mounting disk %s to mount location %s"
+                          % (self.disk, self.mount_point))
+            mounting = "mount %s %s" % (self.disk, self.mount_point)
+            self.log.info("%s" % mounting)
+            ret = process.system(mounting, shell=True)
+            if ret != 0:
+                self.fail("Unable to mount to location %s" % self.mount_point)
+            if self.script == 'runltp':
+                logfile = os.path.join(self.logdir, 'ltp.log')
+                failcmdfile = os.path.join(self.logdir, 'failcmdfile')
+                val += (" -q -p -l %s -C %s -d %s"
+                        % (logfile, failcmdfile, self.mount_point))
+            self.log.info("Args = %s" % self.args)
+            ltpbin_dir = os.path.join(self.srcdir, "ltp-master", 'bin')
+            cmd = os.path.join(ltpbin_dir, self.script) + ' ' + val
+            result = process.run(cmd, ignore_status=True)
+            # Walk the stdout and try detect failed tests from lines
+            # like these:
+            # aio01       5  TPASS  :  Test 5: 10 reads and
+            # writes in  0.000022 sec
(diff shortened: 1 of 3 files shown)


hathyaga:
[job.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/654290/job.txt)

PraveenPenguin:
@hathyaga can this also not a case in https://github.com/avocado-framework-tests/avocado-misc-tests/blob/master/generic/ltp.py (as yaml param )? please consider

hathyaga:
@PraveenPenguin
The existing ltp.py runs a set of default filesystem tests on the current disk. 
I wanted to consider only specific filesystem tests to run on a directory and disk provided by the user.

PraveenPenguin:
@hathyaga yes i know that can we enhance same test case ?

hathyaga:
@PraveenPenguin  let me run the existing ltp.py and check what fs related tests it runs.

PraveenPenguin:
sure i think you need to add some more param in yaml file as releated to fs 

like this ,
    args: '-s fs_di,-s fs_inod'
    mount_point: '/home'
    fs: 'ext4'
in this case you need to have a check if this not possible we can think about custom yaml file

io/disk/ltp_fs.py.data/ltp_fs.yaml
```diff
@@ -0,0 +1,6 @@
+general:
+    disk: '/dev/sda'
+    script: 'runltp'
+    args: '-s fs_di,-s fs_inod'
+    mount_point: '/home'
```
narasimhan-v: would be better if mount point can be dynamically created by the script, and not as an input from the user.

io/disk/ltp_fs.py.data/ltp_fs.yaml
```diff
@@ -0,0 +1,6 @@
...
+    fs: 'ext4'
```
narasimhan-v: Consider adding more fs types

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,118 @@
...
+        build.make(ltp_dir, extra_args='autotools')
+        ltpbin_dir = os.path.join(ltp_dir, 'bin')
+        os.mkdir(ltpbin_dir)
+        process.system('./configure --prefix=%s' % ltpbin_dir)
+        build.make(ltp_dir)
+        build.make(ltp_dir, extra_args='install')
+
+    def test(self):
+        for val in self.args:
+            self.log.info("creating ext4 filesystem on disk %s" % self.disk)
```
narasimhan-v: `self.log.info("creating %s filesystem on disk %s" % (self.fs, self.disk)) `

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,118 @@
...
+            create_fs = "mkfs.%s %s" % (self.fs, self.disk)
+            self.log.info("%s" % create_fs)
+            ret = process.system(create_fs, shell=True)
+            if ret != 0:
+                self.fail("Created filesystem on %s failed" % self.disk)
+            self.log.info("Mounting disk %s to mount location %s"
+                          % (self.disk, self.mount_point))
+            mounting = "mount %s %s" % (self.disk, self.mount_point)
+            self.log.info("%s" % mounting)
```
narasimhan-v: not necessary to log the command

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,118 @@
...
+            ret = process.system(mounting, shell=True)
+            if ret != 0:
```
narasimhan-v: you can eliminate `ret` as mentioned before

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,118 @@
...
+            split_lines = (line.split(None, 3)
+                           for line in result.stdout.splitlines())
+            failed_tests = [items[0] for items in split_lines
+                            if len(items) == 4 and items[2] in fail_statuses]
+            if failed_tests:
+                self.fail("LTP tests failed: %s" % ", ".join(failed_tests))
+            elif result.exit_status != 0:
+                self.fail("No test failures detected, but LTP finished with %s"
+                          % (result.exit_status))
+            self.log.info("Removing the filesystem created on %s" % self.disk)
```
narasimhan-v: Removing filesystem should be handled in `tearDown()`

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,118 @@
...
+        process.system('./configure --prefix=%s' % ltpbin_dir)
+        build.make(ltp_dir)
+        build.make(ltp_dir, extra_args='install')
+
+    def test(self):
+        for val in self.args:
+            self.log.info("creating ext4 filesystem on disk %s" % self.disk)
+            create_fs = "mkfs.%s %s" % (self.fs, self.disk)
+            self.log.info("%s" % create_fs)
+            ret = process.system(create_fs, shell=True)
```
balamuruhans: its good practise to use process.system with try, and with @narasimhan-v's comment mentioned below

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,118 @@
...
+            if ret != 0:
+                self.fail("Created filesystem on %s failed" % self.disk)
+            self.log.info("Mounting disk %s to mount location %s"
+                          % (self.disk, self.mount_point))
+            mounting = "mount %s %s" % (self.disk, self.mount_point)
+            self.log.info("%s" % mounting)
+            ret = process.system(mounting, shell=True)
```
balamuruhans: same here

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,118 @@
...
+            if failed_tests:
+                self.fail("LTP tests failed: %s" % ", ".join(failed_tests))
+            elif result.exit_status != 0:
+                self.fail("No test failures detected, but LTP finished with %s"
+                          % (result.exit_status))
+            self.log.info("Removing the filesystem created on %s" % self.disk)
+            delete_fs = "dd if=/dev/zero bs=512 count=512 of=%s" % self.disk
+            self.log.info("%s" % delete_fs)
+            ret = process.system(delete_fs, shell=True)
+            if ret != 0:
```
balamuruhans: here

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,118 @@
...
+                self.fail("Unable to delete filesystem on disk %s" % self.disk)
+            self.log.info("Unmounting directory %s" % self.mount_point)
+            unmounting = "umount %s" % self.mount_point
+            self.log.info("%s" % unmounting)
+            ret = process.system(unmounting, shell=True)
+            if ret != 0:
```
balamuruhans: here

io/disk/ltp_fs.py
hathyaga: If I have a tearDown(), then the script will compile and install ltp for every argument that I pass. So consciously avoided that.

narasimhan-v: Not an issue. That is how it is supposed to work.
For example, if you decide to multiplex filesystem type, you need to delete each time, which tearDown() will handle.
Also, if there is any error/failure in script, tearDown() will make sure fs is removed(or tried to remove).

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,118 @@
...
+        build.make(ltp_dir)
+        build.make(ltp_dir, extra_args='install')
+
+    def test(self):
+        for val in self.args:
+            self.log.info("creating ext4 filesystem on disk %s" % self.disk)
+            create_fs = "mkfs.%s %s" % (self.fs, self.disk)
+            self.log.info("%s" % create_fs)
+            ret = process.system(create_fs, shell=True)
+            if ret != 0:
```
narasimhan-v: Also, use `ignore_status=True` in `process.run()` and `process.system()`

PraveenPenguin:
@hathyaga me and @narasimhan-v having discussion about this PR and outcome comes that it not good to touch generic ltp stuff as we required some major set up (in this case create file system mounting and un mounting).

so we can have restriction like if there is major change in in setup it self we need to create new Test case for that  and if any other test which is using partially  same setup can extend this test case .
I think that can be ok with you when other test is going to added in this test case :)

One really good point comes as ltp generic as well this test is going to use same mechanism for error or failure catching why not we can have this as a library in avocado   it self .

@santhosh93 any input from you .

waiting v2 for suggested review comment .

Regards
Praveen

PraveenPenguin:
@narasimhan-v are you want to add something ?

narasimhan-v:
@PraveenPenguin I think you have summarized our discussion well.
Next step is checking if an LTP util can be added to avocado.utils. Lets create an issue in avocado for it to check with people for its feasibility and usability.

hathyaga:
@PraveenPenguin @narasimhan-v Sure. I'll send in a V2 for review ASAP

hathyaga:
@PraveenPenguin @narasimhan-v  
Closing this PR. 
New PR: https://github.com/avocado-framework-tests/avocado-misc-tests/pull/223

PraveenPenguin:
i think you forget to press close button :)

PR #219, opened 2016-12-15
renamed yaml file so its easier for Jenkins integration

manvanthar:
renamed yaml file so its easier for Jenkins integration


manvanthar:
will push one single comment will optimised code

PR #223, opened 2016-12-19
Program to stress Filesystem on a disk

hathyaga:
This program uses LTP to run fs tests on user specifed disk

### DIFF
--- a/io/disk/ltp_fs.py
+++ b/io/disk/ltp_fs.py
@@ -1,0 +1,116 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Harsha Thyagaraja <harshkid@linux.vnet.ibm.com>
+#
+# Based on code by Martin Bligh <mbligh@google.com>
+# copyright 2006 Google, Inc.
+# https://github.com/autotest/autotest-client-tests/tree/master/ltp
+
+
+import os
+from avocado import Test
+from avocado import main
+from avocado.utils import build
+from avocado.utils import process, archive
+from avocado.utils.software_manager import SoftwareManager
+
+
+class ltp(Test):
+    '''
+    LTP (Linux Test Project) testsuite
+    '''
+    def setUp(self):
+        '''
+        To check and install dependencies for the test
+        '''
+        sm = SoftwareManager()
+        deps = ['gcc', 'make', 'automake', 'autoconf']
+        for package in deps:
+            if not sm.check_installed(package) and not sm.install(package):
+                self.error(package + ' is needed for the test to be run')
+        self.disk = self.params.get('disk')
+        self.mount_point = self.params.get('mount_point', default=self.srcdir)
+        self.script = self.params.get('script')
+        self.fs = self.params.get('fs', default='ext4')
+        self.args = self.params.get('args', default='').split(',')
+        self.log.info("creating ext4 filesystem on disk %s" % self.disk)
+        create_fs = "mkfs.%s %s" % (self.fs, self.disk)
+        if process.system(create_fs, shell=True, ignore_status=True) != 0:
+            self.fail("Creating filesystem %s on\
+                       %s failed" % (self.fs, self.disk))
+        self.log.info("Mounting disk %s to mount location %s"
+                      % (self.disk, self.mount_point))
+        mounting = "mount %s %s" % (self.disk, self.mount_point)
+        if process.system(mounting, shell=True, ignore_status=True) != 0:
+            self.fail("Unable to mount to location  %s" % self.mount_point)
+        url = "https://github.com/linux-test-project/ltp/\
+               archive/master.zip"
+        tarball = self.fetch_asset("ltp-master.zip",
+                                   locations=[url], expire='7d')
+        archive.extract(tarball, self.srcdir)
+        ltp_dir = os.path.join(self.srcdir, "ltp-master")
+        os.chdir(ltp_dir)
+        build.make(ltp_dir, extra_args='autotools')
+        ltpbin_dir = os.path.join(ltp_dir, 'bin')
+        os.mkdir(ltpbin_dir)
+        process.system('./configure --prefix=%s' % ltpbin_dir)
+        build.make(ltp_dir)
+        build.make(ltp_dir, extra_args='install')
+
+    def test(self):
+        for val in self.args:
+            if self.script == 'runltp':
+                logfile = os.path.join(self.logdir, 'ltp.log')
+                failcmdfile = os.path.join(self.logdir, 'failcmdfile')
+                val += (" -q -p -l %s -C %s -d %s"
+                        % (logfile, failcmdfile, self.mount_point))
+            self.log.info("Args = %s" % self.args)
+            ltpbin_dir = os.path.join(self.srcdir, "ltp-master", 'bin')
+            cmd = os.path.join(ltpbin_dir, self.script) + ' ' + val
+            result = process.run(cmd, ignore_status=True)
+            # Walk the stdout and try detect failed tests from lines
+            # like these:
+            # aio01       5  TPASS  :  Test 5: 10 reads and
+            # writes in  0.000022 sec
+            # vhangup02    1  TFAIL  :  vhangup02.c:88:
+            # vhangup() failed, errno:1
+            # and check for fail_statuses The first part contain test name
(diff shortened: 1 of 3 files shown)


hathyaga:
[job_log.tar.gz](https://github.com/avocado-framework-tests/avocado-misc-tests/files/660426/job_log.tar.gz)

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,116 @@
...
+from avocado import Test
+from avocado import main
+from avocado.utils import build
+from avocado.utils import process, archive
+from avocado.utils.software_manager import SoftwareManager
+
+
+class ltp(Test):
+    '''
+    LTP (Linux Test Project) testsuite
```
PraveenPenguin: i think this for file system related test from LTP can you please be more specific about this ?

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,116 @@
...
+    '''
+    def setUp(self):
+        '''
+        To check and install dependencies for the test
+        '''
+        sm = SoftwareManager()
+        deps = ['gcc', 'make', 'automake', 'autoconf']
+        for package in deps:
```
PraveenPenguin: can this be 

`for package in ['gcc', 'make', 'automake', 'autoconf']:` just want to avoid extra variables

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,116 @@
...
+            if not sm.check_installed(package) and not sm.install(package):
+                self.error(package + ' is needed for the test to be run')
+        self.disk = self.params.get('disk')
```
PraveenPenguin: if yaml file not passed by the user i think we need to fail the test what you say ? if you agree you can check disk value is not `None`

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,116 @@
...
+        self.mount_point = self.params.get('mount_point', default=self.srcdir)
+        self.script = self.params.get('script')
+        self.fs = self.params.get('fs', default='ext4')
+        self.args = self.params.get('args', default='').split(',')
+        self.log.info("creating ext4 filesystem on disk %s" % self.disk)
```
PraveenPenguin: please change this log as for generic  info as user pass other file system from yaml file this information is misleading

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,116 @@
...
+        create_fs = "mkfs.%s %s" % (self.fs, self.disk)
+        if process.system(create_fs, shell=True, ignore_status=True) != 0:
```
PraveenPenguin: please avoid more variable if you don't want to embed command as argument of  `process.system` than please use a single variable like cmd 

ex: can be 
```

cmd = "mkfs.%s %s" % (self.fs, self.disk)
```

same cmd variable you can assign again

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,116 @@
...
+            self.fail("Creating filesystem %s on\
+                       %s failed" % (self.fs, self.disk))
+        self.log.info("Mounting disk %s to mount location %s"
```
PraveenPenguin: better string can be  `Mounting disk <diskname > to mount point <mount point>` what you say ,any way not any strong opinion  here

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,116 @@
...
+                      % (self.disk, self.mount_point))
+        mounting = "mount %s %s" % (self.disk, self.mount_point)
+        if process.system(mounting, shell=True, ignore_status=True) != 0:
```
PraveenPenguin: here also please use `cmd` variable or embed  as a argument of `process.system`

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,116 @@
...
+    def test(self):
+        for val in self.args:
+            if self.script == 'runltp':
+                logfile = os.path.join(self.logdir, 'ltp.log')
+                failcmdfile = os.path.join(self.logdir, 'failcmdfile')
+                val += (" -q -p -l %s -C %s -d %s"
+                        % (logfile, failcmdfile, self.mount_point))
+            self.log.info("Args = %s" % self.args)
+            ltpbin_dir = os.path.join(self.srcdir, "ltp-master", 'bin')
+            cmd = os.path.join(ltpbin_dir, self.script) + ' ' + val
```
PraveenPenguin: using + for concatenation of string not looks good can you please use like this 
`
 cmd ='%s   %s' % (os.path.join(ltpbin_dir, self.script), val)`

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,116 @@
...
+        os.chdir(ltp_dir)
+        build.make(ltp_dir, extra_args='autotools')
+        ltpbin_dir = os.path.join(ltp_dir, 'bin')
+        os.mkdir(ltpbin_dir)
+        process.system('./configure --prefix=%s' % ltpbin_dir)
+        build.make(ltp_dir)
+        build.make(ltp_dir, extra_args='install')
+
+    def test(self):
+        for val in self.args:
```
PraveenPenguin: can you please rename this a valid variable name as `val` not looks good ?

io/disk/ltp_fs.py.data/README.txt
```diff
@@ -0,0 +1,11 @@
+This program downloads the LTP test suite, compiles it and runs 2 filesystem related tests 'fs_di' and 'fs_inod' by involking the runltp test script. 
+This program creates a filesystem on a disk and then mounts it to the mount point specified and then the runltp script is triggered that runs the 2 tests on the mount point.
+
+Inputs needed to run the program:
+disk: ''   eg: /dev/sda
+dir: ''    eg: /mnt
+fs: ''     eg: ext4
+args: '-s fs_di,-s fs_inod'
+
```
PraveenPenguin: please remove all these white spaces

io/disk/ltp_fs.py
```diff
@@ -0,0 +1,116 @@
...
+                            if len(items) == 4 and items[2] in fail_statuses]
+            if failed_tests:
+                self.fail("LTP tests failed: %s" % ", ".join(failed_tests))
+            elif result.exit_status != 0:
+                self.fail("No test failures detected, but LTP finished with %s"
+                          % (result.exit_status))
+
+    def tearDown(self):
+        '''
+        Revert to initial state
```
PraveenPenguin: not very sure about this comment as it wipe off file system as well think can you please make this as cleanup ?

io/disk/ltp_fs.py.data/README.txt
```diff
@@ -0,0 +1,11 @@
+This program downloads the LTP test suite, compiles it and runs 2 filesystem related tests 'fs_di' and 'fs_inod' by involking the runltp test script. 
+This program creates a filesystem on a disk and then mounts it to the mount point specified and then the runltp script is triggered that runs the 2 tests on the mount point.
+
```
PraveenPenguin: can you add warning as well don't pass root disk

narasimhan-v:
`PASS 1-ltp_fs.py:ltp.test`

Can you change the class name to `ltp_fs` please

io/disk/ltp_fs.py.data/ltp_fs.yaml
```diff
@@ -0,0 +1,6 @@
+general:
+    disk: ''
+    script: 'runltp'
+    args: '-s fs_di,-s fs_inod'
+    mount_point: ''
+    fs: ''
```
narasimhan-v: Have `fs` parameter as multiplexing, as an example

io/disk/ltp_fs.py.data/ltp_fs.yaml
```diff
@@ -0,0 +1,6 @@
+general:
+    disk: ''
+    script: 'runltp'
+    args: '-s fs_di,-s fs_inod'
```
narasimhan-v: Consider this for multiplexing.

hathyaga: Sure. I have accounted for ext2, ext3 and ext4 filesystems. If the user wants more, he/she can add

io/disk/ltp_fs.py
hathyaga: @narasimhan-v  
Let's consider the following Instance example if we multiplex the args:
args: !mux
    test:
        args: '-s fs_di'
    test: 
        args: '-s fs_inod'
fs: !mux
    ext4:
        fs: 'ext4'
    ext3:
        fs: 'ext3'

So now my script will run for 4 variants: 
variant1: fs_di ext4
variant2: fs_di ext3
variant3: fs_inod ext4
variant4: fs_inod ext3

This causes the execution time to be more.


If we consider my current config in yaml:
args: '-s fs_di,-s fs_inod'
fs: !mux
    ext4:
        fs: ext4
    ext3:
        fs: ext3


We will have 2 variants here instead of 2:
variant1: fs_di, fs_inod ext4
variant2: fs_di, fs_inod ext3

I want the execution time to be cut down with my consideration of 'args' usage

narasimhan-v: @hathyaga There is no denial of execution time being better.
But my *suggestion* is(I am reiterating), there is a better usability of the test if we give this control to the user.

hathyaga: Okay. I will take care of this.

PR #227, opened 2016-12-21
Enhancing avago9361_vd.py script

vrbagalkote:
Added following test case:
Rebuild
Copyback
create and delete GHS/DHS

### DIFF
--- a/io/disk/Avago_storage_adapter/avago9361_vd.py
+++ b/io/disk/Avago_storage_adapter/avago9361_vd.py
@@ -41,6 +41,11 @@ class Avago9361(Test):
         self.raid_level = str(self.params.get('raid_level', default='0'))
         self.size = str(self.params.get('size', default='all'))
         self.on_off = str(self.params.get('on_off'))
+        self.hotspare = str(self.params.get('hotspare'))
+        self.copyback = self.on_off.replace("e", "").replace("s", "").replace(
+            "/", ":")
+        if not self.hotspare:
+            self.skip("Hotspare test needs a drive to create/test hotspare")
         if not self.on_off:
             self.skip("Online/Offline test needs a drive to operate on")
         if not self.disk:
@@ -64,12 +69,13 @@ class Avago9361(Test):
         elif self.raid_level == 'r50' or self.raid_level == 'r60':
             self.pdperarray = 3
         if self.raid_level == 'r0':
-            if 'cc' in str(self.name) or 'offline' in str(self.name):
+            if str(self.name) in ['cc', 'offline', 'rebuild']:
                 self.skip("Test not applicable for Raid0")
         self.write_policy = ['WT', 'WB', 'AWB']
         self.read_policy = ['nora', 'ra']
         self.io_policy = ['direct', 'cached']
         self.stripe = [64, 128, 256, 512, 1024]
+        self.state = ['start', 'stop', 'start', 'pause', 'resume']
 
     def test_createall(self):
 
@@ -150,6 +156,96 @@ class Avago9361(Test):
                 time.sleep(10)
         self.vd_delete()
 
+    def test_ghs_dhs(self):
+
+        """
+        Test to create Global hot spare and Dedicated hot spare
+        """
+        self.vd_create('WT', 'nora', 'direct', 512)
+        for spare in ['', 'DGs=0']:
+            cmd = "./storcli64 /c%d/%s add hotsparedrive %s" % (
+                self.controller, self.hotspare, spare)
+            self.check_pass(cmd, "Failed to create")
+            self.set_online_offline('offline')
+            self.rebuild('progress', self.hotspare)
+            self.copyback_operation()
+            cmd = "./storcli64 /c%d/%s delete hotsparedrive" % (
+                self.controller, self.hotspare)
+            self.check_pass(cmd, "Failed to delete hotsparedrive")
+        self.vd_delete()
+
+    def copyback_state(self, state):
+
+        """
+        Helper function to run copyback test
+        """
+        if state == 'start':
+            cmd = "./storcli64 /c%d/%s start copyback target=%s" \
+                   % (self.controller, self.hotspare, self.copyback)
+        else:
+            cmd = "./storcli64 /c%d/%s %s copyback" % (self.controller,
+                                                       self.on_off, state)
+        self.check_pass(cmd, "Failed to %s copyback" % state)
+
+    def copyback_operation(self):
+
+        """
+        Function to run copyback operations
+        """
+        for state in self.state:
+            self.copyback_state(state)
+            cmd = "./storcli64 /c%d/%s show copyback" % (self.controller,
+                                                         self.on_off)
+            self.showprogress(cmd)
+        self.sleep_function(cmd)
+
+    def test_rebuild(self):
+
+        """
+        Function to handle rebuild scenario
+        """
+        self.vd_create('WB', 'nora', 'direct', 256)
+        self.set_online_offline('offline')
+        self.rebuild('operations')
+        self.vd_delete()
+
+    def rebuild(self, perform, disk=None):
+
+        """
+        Helper function of all types of rebuild operations
+        """
+        if perform.lower() == "operations":
+            for state in self.state:
+                self.rebuild_state(state)
+                cmd = "./storcli64 /c%d/%s show rebuild" % (self.controller,
+                                                            self.on_off)
+                self.showprogress(cmd)
+            self.sleep_function(cmd)
+        elif perform.lower() == "progress":
+            cmd = "./storcli64 /c%d/%s show rebuild" % (self.controller,
+                                                        disk)
+            self.showprogress(cmd)
+            self.sleep_function(cmd)
+
+    def sleep_function(self, cmd):
+
+        """
+        Helper function for scrit to wait, till the background operation is
+        complete
+        """
+        while self.showprogress(cmd):
+            time.sleep(30)
+
+    def rebuild_state(self, state):
+
+        """
+        Helper function for all CC operations
+        """
+        cmd = "./storcli64 /c%d/%s %s rebuild" % (self.controller,
+                                                  self.on_off, state)
+        self.check_pass(cmd, "Failed to %s Rebuild" % state)
+        time.sleep(10)
+
     def set_online_offline(self, state):
 
         """
@@ -217,7 +313,7 @@ class Avago9361(Test):
         """
         output = process.run(cmd, ignore_status=True, shell=True)
         if output.exit_status != 0:
-            self.fail("Failed to display the CC progress")
+            self.fail("Failed to display the progress")
         for lines in output.stdout.splitlines():
             for times in ['Hour', 'Minute', 'Second']:
                 if times in lines:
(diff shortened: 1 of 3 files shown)


vrbagalkote:
[job.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/665884/job.txt)

Attached is the job.log file

io/disk/Avago_storage_adapter/avago9361_vd.py
```diff
@@ -150,6 +156,99 @@ def test_online_offline(self):
...
+            cmd = "./storcli64 /c%d/%s %s copyback" % (self.controller,
+                                                       self.on_off, state)
+        self.check_pass(cmd, "Failed to start copyback")
+
+    def copyback_operation(self):
+
+        """
+        Function to run copyback operations
+        """
+        for state in ['start', 'stop', 'start', 'pause', 'resume']:
```
narasimhan-v: declare `self.state = ['start', 'stop', 'start', 'pause', 'resume']` in `setUp()` and use it everywhere

io/disk/Avago_storage_adapter/avago9361_vd.py
```diff
@@ -150,6 +156,99 @@ def test_online_offline(self):
...
+        """
+        Helper function to run copyback test
+        """
+        if state == 'start':
+            cmd = "./storcli64 /c%d/%s start copyback target=%s" \
+                   % (self.controller, self.hotspare, self.copyback)
+        else:
+            cmd = "./storcli64 /c%d/%s %s copyback" % (self.controller,
+                                                       self.on_off, state)
+        self.check_pass(cmd, "Failed to start copyback")
```
narasimhan-v: Should be `self.check_pass(cmd, "Failed to %s copyback" % state) `

io/disk/Avago_storage_adapter/avago9361_vd.py
```diff
@@ -150,6 +156,99 @@ def test_online_offline(self):
...
+        self.vd_create('WT', 'nora', 'direct', 512)
+        cmd = "./storcli64 /c%d/%s add hotsparedrive" % (self.controller,
+                                                         self.hotspare)
+        self.check_pass(cmd, "Failed to create GHS")
+        self.set_online_offline('offline')
+        self.rebuild_progress(self.hotspare)
+        self.copyback_operation()
+        cmd = "./storcli64 /c%d/%s delete hotsparedrive" % (self.controller,
+                                                            self.hotspare)
+        self.check_pass(cmd, "Failed to delete hotsparedrive")
```
vrbagalkote: For dedicated, I am anyway running VD delete, so automatically DHS will be deleted.

io/disk/Avago_storage_adapter/avago9361_vd.py
```diff
@@ -150,6 +156,99 @@ def test_online_offline(self):
...
+        """
+        Helper function of all types of rebuild operations
+        """
+        for state in ['start', 'stop', 'start', 'pause', 'resume']:
+            self.rebuild_state(state)
+            cmd = "./storcli64 /c%d/%s show rebuild" % (self.controller,
+                                                        self.on_off)
+            self.showprogress(cmd)
+        while self.showprogress(cmd) is True:
+            time.sleep(30)
```
narasimhan-v: Can this chunk of monitoring progress also become a helper function ?

io/disk/Avago_storage_adapter/avago9361_vd.py
vrbagalkote: Ah, but ultimately the code path traversed is same. So I feel this should be fine.

io/disk/Avago_storage_adapter/avago9361_vd.py
vrbagalkote: I thought of it, but it really is not required.

io/disk/Avago_storage_adapter/avago9361_vd.py
```diff
@@ -150,6 +156,99 @@ def test_online_offline(self):
...
+    def copyback_operation(self):
+
+        """
+        Function to run copyback operations
+        """
+        for state in ['start', 'stop', 'start', 'pause', 'resume']:
+            self.copyback_state(state)
+            cmd = "./storcli64 /c%d/%s show copyback" % (self.controller,
+                                                         self.on_off)
+            self.showprogress(cmd)
```
vrbagalkote: I have used each command only once, and there is a anyway helper function for show progress. And moreover, we need to pass VD number for few operations, and for others we need to pass disk slot and also few command expects in different formats. So, its better to be in this way.

io/disk/Avago_storage_adapter/avago9361_vd.py
```diff
@@ -64,7 +69,8 @@ def setUp(self):
         elif self.raid_level == 'r50' or self.raid_level == 'r60':
             self.pdperarray = 3
         if self.raid_level == 'r0':
-            if 'cc' in str(self.name) or 'offline' in str(self.name):
+            if 'cc' in str(self.name) or 'offline' in str(self.name) or \
+               'rebuild' in str(self.name):
```
balamuruhans: where does this self.name got declared?

balamuruhans: Also `if str(self.name) in ['cc', 'offline', 'rebuild']:` can be used if you check against them.

io/disk/Avago_storage_adapter/avago9361_vd.py.data/avago9361_vd.yaml
```diff
@@ -1,8 +1,9 @@
 cenario:
     contorller: 0
```
balamuruhans: `self.controller = int(self.params.get('controller', default='0'))` but in yaml it is %s/"contorller"/controller

io/disk/Avago_storage_adapter/avago9361_vd.py
balamuruhans: `def rebuild_progress(self, cmd):` can be reused by giving `cmd` instead of `disk` and you can form the required cmd before calling `rebuild_progress(cmd)`

vrbagalkote: I the disk I pass are not the same to know the rebuild progress. Both are two different disks. So, I am preparing cmd and calling the function with relevant cmd. I am I missing anything, I didnt get exactly what you are trying to tell.

balamuruhans: @vrbagalkote but it will avoid redundancy of code.

io/disk/Avago_storage_adapter/avago9361_vd.py
PraveenPenguin: @balamuruhans self.name return test name so I think no need of declaration

io/disk/Avago_storage_adapter/avago9361_vd.py
```diff
@@ -150,6 +156,99 @@ def test_online_offline(self):
...
+
+    def rebuild_state(self, state):
+
+        """
+        Helper function for all CC operations
+        """
+        cmd = "./storcli64 /c%d/%s %s rebuild" % (self.controller,
+                                                  self.on_off, state)
+        self.check_pass(cmd, "Failed to %s Rebuild" % state)
+        time.sleep(10)
```
balamuruhans: sleep() inside test is not a good way

io/disk/Avago_storage_adapter/avago9361_vd.py
```diff
@@ -150,6 +156,99 @@ def test_online_offline(self):
...
+
+    def rebuild_progress(self, disk):
+
+        """
+        Helper function to see the progress of rebuild operation
+        """
+        cmd = "./storcli64 /c%d/%s show rebuild" % (self.controller,
+                                                    disk)
+        self.showprogress(cmd)
+        while self.showprogress(cmd) is True:
```
balamuruhans: `while self.showprogress(cmd):` suffice the condition

io/disk/Avago_storage_adapter/avago9361_vd.py
balamuruhans: @PraveenPenguin, Thanks for clarifying

io/disk/Avago_storage_adapter/avago9361_vd.py
```diff
@@ -150,6 +156,99 @@ def test_online_offline(self):
...
+
+        """
+        Function to handle rebuild scenario
+        """
+        self.vd_create('WB', 'nora', 'direct', 256)
+        self.set_online_offline('offline')
+        self.rebuild_operations()
+        self.vd_delete()
+
+    def rebuild_operations(self):
```
balamuruhans: just curious to know separate methods are required for rebuild_progress() and rebuild_operations() as both looks similar, instead it can be
```
def rebuild(self, perform):
    if (perform.lower() == "operations"):
        for state in ['start', 'stop', 'start', 'pause', 'resume']:
            self.rebuild_state(state)
             cmd = "./storcli64 /c%d/%s show rebuild" % (self.controller, self.on_off)
             self.showprogress(cmd)
    elif (perform.lower() == "progress"):
        cmd = "./storcli64 /c%d/%s show rebuild" % (self.controller, disk)
        self.showprogress(cmd)
    while self.showprogress(cmd):
        time.sleep(30)
```

io/disk/Avago_storage_adapter/avago9361_vd.py
narasimhan-v: I don't think code path traversed in both cases is same.
Deleting hotspare and then deleting vd is different from deleting vd directly.
If procedure is former, try to stick with the right procedure.

io/disk/Avago_storage_adapter/avago9361_vd.py
narasimhan-v: Ok.

io/disk/Avago_storage_adapter/avago9361_vd.py
narasimhan-v: Exactly, I am telling from a code redundancy point of view:
```
while self.showprogress(cmd) is True:
    time.sleep(30)
```
can be a helper function

io/disk/Avago_storage_adapter/avago9361_vd.py
narasimhan-v: @balamuruhans Generally, I agree. But since these are IO operations, they need some time before starting any other operation.
@vrbagalkote Pls check if all such sleeps are valid.

PraveenPenguin:
@vrbagalkote please squash  commits in a single commit and provide a valid commit message as well sign attribute to make Travis happy :)

vrbagalkote:
Everything was, but I happened to make a small change from web and commit, but then it failed. I will send this patch from cli only.

PraveenPenguin:
web always a issue , please rebase

vrbagalkote:
Done..

PraveenPenguin:
can you please attach job log as well

vrbagalkote:
[job.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/668568/job.txt)
Attached is the job.txt

PraveenPenguin:
@balamuruhans @narasimhan-v thanks

PR #231, opened 2017-01-11
Added deps check and job file

abdhaleegit:
FIO requires libaio package needs to be installed,
so check before test gets executed, the old version
of job file was not working so added latest job file with more scenarios

### DIFF
--- a/cpu/cpupower.py
+++ b/cpu/cpupower.py
@@ -56,7 +56,7 @@ class cpupower(Test):
         initial_governor = self.cpu_freq_path('scaling_governor', self.cpu)
         return (min, max, cur, initial_governor)
 
-    def cpu_freq_path(self, file, cpu_num, path=None):
+    def cpu_freq_path(self, file, cpu_num):
         """
         get cpu_freq values
         :param: file: is filename which data needs to be fetched


perf/fiotest.py
```diff
@@ -44,16 +46,34 @@ def setUp(self):
...
+        os.chdir(self.srcdir)
+        process.run('./configure', ignore_status=True, sudo=True)
         build.make(self.srcdir)
 
     def test(self):
         """
         Execute 'fio' with appropriate parameters.
         """
+        self.testdir = '/data1'
+        self.scrtchdir = '/scratch'
```
PraveenPenguin: can these two param pass as a yaml input or it is standard for fio ?

PraveenPenguin:
@abdhaleegit job log please

perf/fiotest.py
abdhaleegit: these are the default directories required for test, and is used by perf/fiotest.py.data/fio-mixed.job 
As the dir name will not change, I do not think we should pass it from yaml.. 

we can change the dir from yaml. but than we have to also change fio-mixed.job with new dir names passed from yaml

abdhaleegit: perhaps, the test directories are created in root '/' and it may use the disk space of root device. as long as root disk is more than 5G than no space issue

PraveenPenguin: @abdhaleegit sure

PraveenPenguin:
@abdhaleegit jog log please

perf/fiotest.py
```diff
@@ -44,16 +46,34 @@ def setUp(self):
...
+        if not os.path.exists(self.testdir):
+            os.mkdir(self.testdir)
```
PraveenPenguin: in that case can this be like 
`os.mkdir('/data1')`

perf/fiotest.py
```diff
@@ -44,16 +46,34 @@ def setUp(self):
...
+        if not os.path.exists(self.scrtchdir):
+            os.mkdir(self.scrtchdir)
```
PraveenPenguin: same here as well 
`os.mkdir('/scratch')`
just want to eliminate extra variable

perf/fiotest.py
abdhaleegit: this will fail if /data1 is already exists

PraveenPenguin: oh my bad luck , i think it not logic change just want to eliminate extra variable 

```
       if not os.path.exists('/data1'):
           os.mkdir('/data1')
```

perf/fiotest.py
PraveenPenguin: here also same

perf/fiotest.py
abdhaleegit: But, it is good to always have a user given dir to test from yaml.. will add it

PraveenPenguin: @abdhaleegit sure make sense as we have off line discussion in the same line 
listing TODO 
1- add yaml support 
2- if you want we can add a logic in main script as when yaml entry pass we can patch  perf/fiotest.py.data/fio-mixed.job also with run directory 
3- add README file about the info 
what you say ?

perf/fiotest.py
```diff
@@ -44,16 +46,34 @@ def setUp(self):
         """
         Build 'fio'.
         """
-        tarball = self.fetch_asset('http://brick.kernel.dk/snaps/fio-2.1.10.tar.gz')
+        sm = SoftwareManager()
+        dist = distro.detect()
+        if 'Ubuntu' in dist.name:
+            deps = ['libaio1', 'libaio-dev', 'bzip2']
+        else:
+            deps = ['libaio', 'libaio-devel', 'bzip2']
```
apahim: At least in Fedora, `libaio` is a dependency for `libaio-devel`, so it will be installed anyway.
I'm in favour of now repeating code, so I think you could initialize `deps` with `['bzip2']` before this if and, inside the if, extend it with what is needed for each distro.

abdhaleegit: sure will do it

perf/fiotest.py
abdhaleegit: I will do it

PraveenPenguin:
@abdhaleegit any update here ?

PR #233, opened 2017-01-15
Fix kernel selftest

PraveenPenguin:
fix kernel:kselftest.py
1- added support for fedora
2- added expire parameter in fetch_asset

improve failure reporting in kselftest.py
added logic for check self test failure log from debug log

### DIFF
--- a/kernel/kselftest.py
+++ b/kernel/kselftest.py
@@ -60,7 +60,7 @@ class kselftest(Test):
             deps = deps + rdeps
         elif 'SuSE' in detected_distro.name:
             deps = deps + sdeps
-        elif 'centos' in detected_distro.name:
+        elif detected_distro.name in ['centos', 'fedora']:
             deps = deps + cdeps
         for package in deps:
             if not smg.check_installed(package) and not smg.install(package):
@@ -74,7 +74,7 @@ class kselftest(Test):
             tarball_base = 'linux-%s.tar.gz' % (version)
             tarball_url = '%s/v%s.x/%s' % (url, version[:1], tarball_base)
             self.log.info('Downloading linux kernel tarball')
-            self.tarball = self.fetch_asset(tarball_url)
+            self.tarball = self.fetch_asset(tarball_url, expire='7d')
             archive.extract(self.tarball, self.srcdir)
             linux_src = 'linux-%s' % (version)
             self.buldir = os.path.join(self.srcdir, linux_src)
--- a/kernel/kselftest.py
+++ b/kernel/kselftest.py
@@ -14,6 +14,7 @@
 # Author: Abdul Haleem <abdhalee@linux.vnet.ibm.com>
 
 import os
+import re
 
 from avocado import Test
 from avocado import main
@@ -92,10 +93,19 @@ class kselftest(Test):
         """
         Execute the kernel selftest
         """
+        self.count = 0
         result = build.make(self.srcdir, extra_args='run_tests')
         for line in str(result).splitlines():
             if '[FAIL]' in line:
-                self.fail("Selftest testcase failed, please check the test logs")
+                self.count += 1
+                self.log.info("testcase failed log from build log %s" % line)
+        for line in open(os.path.join(self.logdir, 'debug.log')).readlines():
+            match = re.search(r'selftests:\s+\w+\s+\[FAIL]', line)
+            if match:
+                self.count += 1
+                self.log.info("fail from debug log%s" % match.group(0))
+        if self.count > 0:
+            self.fail("Selftest testcase failed")
 
 
 if __name__ == "__main__":


PraveenPenguin:
job log : 
[joblog.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/706569/joblog.txt)

PraveenPenguin:
@apahim can you please have a look in this PR

kernel/kselftest.py
```diff
@@ -60,7 +60,7 @@ def setUp(self):
             deps = deps + rdeps
         elif 'SuSE' in detected_distro.name:
             deps = deps + sdeps
-        elif 'centos' in detected_distro.name:
+        elif detected_distro.name in ['centos', 'fedora']:
             deps = deps + cdeps
```
apahim: Ok. Those variable names are not very clear (cdeps, sdeps, ...), but that's a different matter.

kernel/kselftest.py
```diff
@@ -74,7 +74,7 @@ def setUp(self):
             tarball_base = 'linux-%s.tar.gz' % (version)
             tarball_url = '%s/v%s.x/%s' % (url, version[:1], tarball_base)
             self.log.info('Downloading linux kernel tarball')
-            self.tarball = self.fetch_asset(tarball_url)
+            self.tarball = self.fetch_asset(tarball_url, expire='7d')
```
apahim: ok

kernel/kselftest.py
```diff
@@ -92,10 +93,19 @@ def test(self):
         """
         Execute the kernel selftest
         """
+        self.count = 0
```
apahim: Two points here:
- You're not using `self.count` outside `test()`, so no need for it to be an object attribute. A local variable is enough. So, you could use `count` instead of `self.count`
- You're using `self.count` to make a boolean-like test in line 107. I mean, you're not really using its value other than to check if it's 0 or non-0. In that case, you could call it `errors` instead of `count` and initialize it as `False`, changing it to `True` when some error happens.

kernel/kselftest.py
```diff
@@ -92,10 +93,19 @@ def test(self):
...
         result = build.make(self.srcdir, extra_args='run_tests')
         for line in str(result).splitlines():
             if '[FAIL]' in line:
-                self.fail("Selftest testcase failed, please check the test logs")
+                self.count += 1
+                self.log.info("testcase failed log from build log %s" % line)
```
apahim: `self.log.info("Testcase failed. Log from build: %s" % line)`

kernel/kselftest.py
```diff
@@ -92,10 +93,19 @@ def test(self):
...
+        for line in open(os.path.join(self.logdir, 'debug.log')).readlines():
+            match = re.search(r'selftests:\s+\w+\s+\[FAIL]', line)
+            if match:
+                self.count += 1
+                self.log.info("fail from debug log%s" % match.group(0))
+        if self.count > 0:
+            self.fail("Selftest testcase failed")
```
apahim: `Testcase failed during selftests`

kernel/kselftest.py
```diff
@@ -92,10 +93,19 @@ def test(self):
...
         for line in str(result).splitlines():
             if '[FAIL]' in line:
-                self.fail("Selftest testcase failed, please check the test logs")
+                self.count += 1
+                self.log.info("testcase failed log from build log %s" % line)
+        for line in open(os.path.join(self.logdir, 'debug.log')).readlines():
+            match = re.search(r'selftests:\s+\w+\s+\[FAIL]', line)
+            if match:
+                self.count += 1
+                self.log.info("fail from debug log%s" % match.group(0))
```
apahim: `Testcase failed. Log from debug: %s`

kernel/kselftest.py
PraveenPenguin: @apahim sure make sense , thanks will do

kernel/kselftest.py
PraveenPenguin: it is already in older code  i think it denote for release like cdeps for centos dependency , any will rectify this

PraveenPenguin: it is already in older code  i think it denote for release like cdeps for centos dependency , any will rectify this

kernel/kselftest.py
PraveenPenguin: sure

kernel/kselftest.py
PraveenPenguin: ok

kernel/kselftest.py
PraveenPenguin: ok

PraveenPenguin:
@apahim thanks for looking in to this PR , will adopt changes and raise PR

PraveenPenguin:
@apahim raised new PR #235 with changes , please have a look

PR #252, opened 2017-01-31
Added proper check for the script arcconf_cntl_oper

vrbagalkote:
Firmware flash test case was skipping due to im-proper check,
fixed the same

### DIFF
--- a/io/disk/arcconf/arcconf_cntl_oper.py
+++ b/io/disk/arcconf/arcconf_cntl_oper.py
@@ -15,7 +15,6 @@
 # Copyright: 2016 IBM
 # Author: Manvanthara B Puttashankar <manvanth@linux.vnet.ibm.com>
 
-
 """
 arcconf - Array configuration utility for PMC-Sierra
 (Microsemi) Controllers.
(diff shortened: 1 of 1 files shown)


vrbagalkote:
@PraveenPenguin Any idea why `inspekt indent` is failing because of issues in other scripts ?

PraveenPenguin:
@vrbagalkote thanks for pointing this , sorry for delay as i was on vacation , fixed travis issue now , can you please rebase ?  now it build should pass .

Regards
Praveen

io/disk/arcconf/arcconf_cntl_oper.py
```diff
@@ -83,7 +83,7 @@ def setUp(self):
                 http_repo = "%s%s.rpm" % (self.http_path, self.tool_name)
                 self.repo = self.fetch_asset(http_repo, expire='10d')
                 cmd = "rpm -ivh %s" % self.repo
-            if process.system(cmd, ignore_status=True, shell=True) == 0:
+            if process.system(cmd, ignore_status=True, shell=True) != 0:
```
PraveenPenguin: Well I wouldn't call this a big change, would you? (the purpose of authorship is to mark significant changes like new API, complete reworks etc. This change is only a minor modification which does not require the change of copyright) .

See eg. the first response to http://programmers.stackexchange.com/questions/203014/author-has-inserted-copyright-into-code-with-gnu-public-license-notice-implica

vrbagalkote: Yes, agreed!! I was doing few experiments while I got the travis error, and forgot to delete this. Thanks.

PraveenPenguin: @vrbagalkote thanks

PR #261, opened 2017-02-07
Add check for ppc64_cpu:subcore sub-test

sacsant:
subcore test fails on Linux running as a PowerVM guest.
In this environment subcores_per_core sysfs attribute is missing.
Add a check to skip the test if subcores_per_core is not available.

### DIFF
--- a/cpu/ppc64_cpu_test.py
+++ b/cpu/ppc64_cpu_test.py
@@ -18,6 +18,7 @@
 Test to verify ppc64_cpu command.
 """
 
+import os
 from avocado import Test
 from avocado import main
 from avocado.utils import process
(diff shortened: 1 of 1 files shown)


cpu/ppc64_cpu_test.py
```diff
@@ -76,7 +80,8 @@ def test(self):
             process.system_output("ppc64_cpu --info")
             self.smt()
             self.core()
-            self.subcore()
+            if self.smt_subcores == 1:
```
nasastry: @sacsant Instead of having a flag, this can be directly checked.
ex:
if os.path.exists("/sys/devices/system/cpu/subcores_per_core"):
    self.subcore()

sacsant: Since this code is inside a loop, a flag will help save multiple calls to os.path.exists

nasastry: @sacsant makes sense, I missed the loop.

PraveenPenguin: i think should be ok as this check need to run in loop so flag seems me ok

PraveenPenguin:
@narasimhan-v @nasastry thanks

PR #262, opened 2017-02-07
Fix posixtest runtime failure

sacsant:
posixtest fails to run due to following error

    Command 'rpm -q gcc' finished with 0 after 0.164994955063s
    Fetching http://ufpr.dl.sourceforge.net/sourceforge/posixtest/posixtestsuite-1.5.2.tar.gz -> /home/avocado-fvt-wrapper/data/cache/posixtestsuite-1.5.2.tar.gz.0JvAha
    PARAMS (key=patch, path=*, default=posix-linux.patch) => 'posix-linux.patch'

    Reproduced traceback from: /usr/lib/python2.7/site-packages/avocado_framework-44.0-py2.7.egg/avocado/core/test.py:469
    Traceback (most recent call last):
      File "/home/avocado-misc-tests/kernel/posixtest.py", line 55, in setUp
        os.chdir(self.srcdir)
    OSError: [Errno 2] No such file or directory: '/var/tmp/avocado_yGDhUM/1-kernel_posixtest.py_Posixtest.test/src/avocado'

Fix the failure by using a correct delimiter to extract version string.

### DIFF
--- a/kernel/posixtest.py
+++ b/kernel/posixtest.py
@@ -47,7 +47,7 @@ class Posixtest(Test):
                                    '/sourceforge/posixtest/posixtestsuite-1.5.2.tar.gz')
         data_dir = os.path.abspath(self.datadir)
         archive.extract(tarball, self.srcdir)
-        version = os.path.basename(tarball.split('-')[0])
+        version = os.path.basename(tarball.split('-1.')[0])
         self.srcdir = os.path.join(self.srcdir, version)
 
         patch = self.params.get(


PraveenPenguin:
@sacsant  I tired to run this  and test  working fine for me  as i am running this in fedora 25 and my avocado version is 44 

[root@localhost avocado-misc-tests]# avocado -v
Avocado 44.0
[root@localhost avocado-misc-tests]#

and when i try to untar after wget http://ufpr.dl.sourceforge.net/sourceforge/posixtest/posixtestsuite-1.5.2.tar.gz  it have directory as posixtestsuite

[root@localhost avocado-misc-tests]# ll | grep posix
drwxr-xr-x.  7 root root    4096 Oct 19  2006 posixtestsuite
-rw-r--r--.  1 root root 1092856 Oct 19  2006 posixtestsuite-1.5.2.tar.gz
[root@localhost avocado-misc-tests]# 


can you please help me what environment you have

sacsant:
I have tried with 44 version running on RHEL 7.3.

In this environment I have seen that the above check returns 'avocado' as string. This is then appended to the directory which results in non-existent path and the test fails.

PraveenPenguin:
@sacsant sure i also tried in RHEL7.3 and it problem with delimiter as `-`as it produce 

['/root/avocado', 'fvt', 'wrapper/data/cache/posixtestsuite', '1.5.2.tar.gz']

in other distribution it like this  
['/usr/share/avocado/data/cache/posixtestsuite', '1.5.2.tar.gz']

any way if we change delimiter to `-1.` should not effect to other distribution to  run  so we can have this .

PR #282, opened 2017-02-22
Code fix for lvsetup.py

hathyaga:
Fix for xfs and btrfs filesystem

### DIFF
--- a/io/disk/lvsetup.py
+++ b/io/disk/lvsetup.py
@@ -31,6 +31,7 @@ import os
 import avocado
 from avocado import Test
 from avocado import main
+from avocado.utils.software_manager import SoftwareManager
 from avocado.utils import lv_utils
 
 
@@ -45,11 +46,20 @@ class Lvsetup(Test):
         """
         Check existence of input PV,VG, LV and snapshots prior to Test.
         """
+        pkgs = []
+        smm = SoftwareManager()
         self.disk = self.params.get('disk', default=None)
         vg_name = self.params.get('vg_name', default='avocado_vg')
         lv_name = self.params.get('lv_name', default='avocado_lv')
         self.lv_size = self.params.get('lv_size', default='1G')
         self.fs_name = self.params.get('fs', default='ext4')
+        if self.fs_name == 'xfs':
+            pkgs.append('xfsprogs')
+        if self.fs_name == 'btrfs':
+            pkgs.append('btrfs-progs')
+        for pkg in pkgs:
+            if not smm.check_installed(pkg) and not smm.install(pkg):
+                self.skip("%s package is need to test" % pkg)
         lv_snapshot_name = self.params.get(
             'lv_snapshot_name', default='avocado_sn')
         self.lv_snapshot_size = self.params.get(
--- a/io/disk/lvsetup.py.data/lvsetup.yaml
+++ b/io/disk/lvsetup.py.data/lvsetup.yaml
@@ -1,7 +1,7 @@
 scenario: !mux
 # btrfs needs minimum LV size of 641M
     real:
-        disk: /dev/vde
+        disk: /dev/sdc
         lv_size: "641M"
 filesystem: !mux
     ext2:


hathyaga:
Fix to handle xfs and btrfs filesystem

hathyaga:
[lvsetup_joblog.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/794222/lvsetup_joblog.txt)

narasimhan-v:
Looks good to me

io/disk/lvsetup.py
```diff
@@ -45,11 +46,20 @@ def setUp(self):
...
         self.disk = self.params.get('disk', default=None)
         vg_name = self.params.get('vg_name', default='avocado_vg')
         lv_name = self.params.get('lv_name', default='avocado_lv')
         self.lv_size = self.params.get('lv_size', default='1G')
         self.fs_name = self.params.get('fs', default='ext4')
+        if self.fs_name == 'xfs':
+            pkgs.append('xfsprogs')
+        if self.fs_name == 'btrfs':
+            pkgs.append('btrfs-progs')
+        for pkg in pkgs:
```
PraveenPenguin: for loop is not required as you are checking about only one package

io/disk/lvsetup.py
```diff
@@ -45,11 +46,20 @@ def setUp(self):
...
         Check existence of input PV,VG, LV and snapshots prior to Test.
         """
+        pkgs = []
+        smm = SoftwareManager()
         self.disk = self.params.get('disk', default=None)
         vg_name = self.params.get('vg_name', default='avocado_vg')
         lv_name = self.params.get('lv_name', default='avocado_lv')
         self.lv_size = self.params.get('lv_size', default='1G')
         self.fs_name = self.params.get('fs', default='ext4')
+        if self.fs_name == 'xfs':
```
balamuruhans: its good to compare strings as,
`if self.fs_name.lower() == 'xfs':`

io/disk/lvsetup.py
```diff
@@ -45,11 +46,20 @@ def setUp(self):
...
+            pkgs.append('xfsprogs')
+        if self.fs_name == 'btrfs':
```
balamuruhans: same here

io/disk/lvsetup.py
```diff
@@ -45,11 +46,20 @@ def setUp(self):
         """
         Check existence of input PV,VG, LV and snapshots prior to Test.
         """
+        pkgs = []
```
PraveenPenguin: it can be string not a list like 

`pkgs = " "`

io/disk/lvsetup.py
```diff
@@ -45,11 +46,20 @@ def setUp(self):
...
+        smm = SoftwareManager()
         self.disk = self.params.get('disk', default=None)
         vg_name = self.params.get('vg_name', default='avocado_vg')
         lv_name = self.params.get('lv_name', default='avocado_lv')
         self.lv_size = self.params.get('lv_size', default='1G')
         self.fs_name = self.params.get('fs', default='ext4')
+        if self.fs_name == 'xfs':
+            pkgs.append('xfsprogs')
```
PraveenPenguin: than` pkgs = 'xfsprogs'`

io/disk/lvsetup.py
```diff
@@ -45,11 +46,20 @@ def setUp(self):
...
+        if self.fs_name == 'btrfs':
+            pkgs.append('btrfs-progs')
```
PraveenPenguin: here also string assignment

io/disk/lvsetup.py
narasimhan-v: On that lines, would be better to do
`self.fs_name = self.params.get('fs', default='ext4').lower()`

io/disk/lvsetup.py
hathyaga: I want to keep the for loop for the reason that if we had more packages to be updated for the test in the future, the list could be easily appended for that. Otherwise, I have to do a string assignment everywhere for new packages

PraveenPenguin: i don't think it can be requirement if some body want to add more package (if you want to add more packages please add in this PR ) that time it will implement in this way as this not looks good

PraveenPenguin:
closing  this PR as i think #286 is a new version of this PR , @hathyaga  please close older PR if you have new version of same test it help to reviewer to review

PR #284, opened 2017-02-27
Skip test case if sensors are not found

harish-24:
This patch skips the test in case if there are no sensors found

### DIFF
--- a/cpu/sensors.py
+++ b/cpu/sensors.py
@@ -42,7 +42,7 @@ class Sensors(Test):
         Checks any errors in command result
         """
         errs = []
-        s_output = process.system_output(cmd)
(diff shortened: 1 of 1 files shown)


cpu/sensors.py
```diff
@@ -98,7 +98,10 @@ def test(self):
             except process.CmdError:
                 self.error(
                     'Starting Service Failed. Make sure module is loaded')
-        process.run('yes | sudo sensors-detect', shell=True)
+        det_op = process.run('yes | sudo sensors-detect', shell=True).stdout
```
PraveenPenguin: just one minor change can you please pass `ignore_status=True` as one parameter in `process.run` ?

harish-24: sure. pushed with change.

PR #290, opened 2017-02-28
Added valgrind test case

harish-24:
Valgrind is an instrumentation framework for building dynamic analysis tools. Valgrind tools that can automatically detect many memory management and threading bugs

### DIFF
--- a/toolchain/valgrind.py
+++ b/toolchain/valgrind.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+# Copyright: 2017 IBM
+# Author: Harish S <harish@linux.vnet.ibm.com>
+
+import os
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+from avocado.utils import build
+from avocado.utils import archive
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Valgrind(Test):
+    '''
+    Valgrind is an instrumentation framework for building dynamic analysis
+    tools. Valgrind tools can automatically detect many memory management
+    and threading bugs, and profile your programs in detail
+
+    The Test fetches the recent valgrind tar and runs the tests in it
+    '''
+
+    def setUp(self):
+        smm = SoftwareManager()
+        self.failures = []
+        deps = ['gcc', 'make']
+
+        for package in deps:
+            if not smm.check_installed(package) and not smm.install(package):
+                self.skip(package + ' is needed for the test to be run')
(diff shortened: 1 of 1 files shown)


toolchain/valgrind.py
```diff
@@ -0,0 +1,83 @@
...
+
+    The Test fetches the recent valgrind tar and runs the tests in it
+    '''
+
+    def setUp(self):
+        smm = SoftwareManager()
+        self.failures = []
+        deps = ['gcc', 'make']
+
+        for package in deps:
```
PraveenPenguin: please use like this way 
`
for package in ['gcc', 'make']:`  as want to eliminate extra variable

toolchain/valgrind.py
```diff
@@ -0,0 +1,83 @@
...
+        """
+        run 'make' of given command and write the summary to respective files
+        """
+        summary = ''
+        flag = False
+        results = build.run_make(
+            self.srcdir, extra_args=cmd, ignore_status=True).stdout
+        for line in results.splitlines():
+            if line.startswith('==') and line.endswith('=='):
+                flag = True
```
PraveenPenguin: flag variable is not required it can be like this 
```
if line.startswith('==') and line.endswith('=='):
      summary +=  '%s \n' % line
```

toolchain/valgrind.py
```diff
@@ -0,0 +1,83 @@
...
+            if flag:
+                summary += line + '\n'
+        with open(os.path.join(self.outputdir, cmd + '_result'), 'w') as f_obj:
```
PraveenPenguin: can this not like this 

`with open(os.path.join(self.outputdir,  '%s_result' %cmd), 'w') as f_obj: `

as concatenating strings using `+` not looks good

toolchain/valgrind.py
```diff
@@ -0,0 +1,83 @@
...
+            if not smm.check_installed(package) and not smm.install(package):
+                self.skip(package + ' is needed for the test to be run')
+        tarball = self.fetch_asset(
+            "http://valgrind.org/downloads/valgrind-3.12.0.tar.bz2",
+            expire='7d')
+        archive.extract(tarball, self.srcdir)
+        version = os.path.basename(tarball.split('.tar.')[0])
+        self.srcdir = os.path.join(self.srcdir, version)
+
+        os.chdir(self.srcdir)
```
narasimhan-v: os.chdir(os.path.join(self.srcdir, version))

PraveenPenguin: i think should be fine as @harish-24 is using `self.srcdir` in multiple places but again it his call

narasimhan-v: Ok, didn't notice the other usages

toolchain/valgrind.py
harish-24: @PraveenPenguin  `flag` is required because it does not append `summary` for that line alone, but till the end after that line

toolchain/valgrind.py
harish-24: @narasimhan-v Yeah. It is used at multiple places after this, so I'd prefer it to be the existing way.

harish-24:
@PraveenPenguin Closing this. Opened #292

toolchain/valgrind.py
PraveenPenguin: @harish-24 may be i got confused , logically it should can you please elaborate more ?

PR #293, opened 2017-03-01
Added username in args of bonnie test

santhosh93:
Bonnie test expects the username in args.
So modified the args in the test.

### DIFF
--- a/perf/bonnie.py
+++ b/perf/bonnie.py
@@ -23,6 +23,7 @@ Bonnie test
 """
 
 import os
+import getpass
 from avocado import Test
 from avocado import main
 from avocado.utils import archive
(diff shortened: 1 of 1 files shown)


perf/bonnie.py
```diff
@@ -46,7 +47,8 @@ def setUp(self):
         """
 
         self.scratch_dir = self.params.get('dir', default=self.srcdir)
-        self.uid_to_use = self.params.get('uid-to-use', default=None)
+        self.username = getpass.getuser()
+        self.uid_to_use = self.params.get('uid-to-use', default=self.username)
```
apahim: `self.username` is not being used anywhere outside `setUp()`. No need to create it as an instance variable. Actually, no need to create it at all:
```python
from getpass import getuser
...

        self.uid_to_use = self.params.get('uid-to-use', default=getuser())
```

perf/bonnie.py
```diff
@@ -68,8 +70,7 @@ def test(self):
         args.append('-d %s' % self.scratch_dir)
         args.append('-n %s' % self.number_to_stat)
         args.append('-s %s' % self.data_size)
-        if self.uid_to_use:
-            args.append('-u %s' % self.uid_to_use)
+        args.append('-u %s' % self.uid_to_use)
```
apahim: Ok

santhosh93:
@apahim : Removed the self.username variable...

apahim:
Ok, thank you. Looks good.

PR #295, opened 2017-03-01
Implemented disk mount on dir with partition utility in tiobench.py

Naresh-ibm:
[tiobench-logs-with-partition-utils.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/810234/tiobench-logs-with-partition-utils.txt)
Added disk parameter and different file system types in
tiobench.yaml and then implemented the unmount if it is already
mounted and then create the file system on it and mount on given
directory. used partition utility to do all these operations.

Signed-off-bynced: Naresh Bannoth <nbannoth@in.ibm.com>

### DIFF
--- a/io/disk/tiobench.py
+++ b/io/disk/tiobench.py
@@ -31,6 +31,7 @@ from avocado.utils import archive
 from avocado.utils import build
 from avocado.utils import process
 from avocado.utils.software_manager import SoftwareManager
+from avocado.utils.partition import Partition
 
 
 class Tiobench(Test):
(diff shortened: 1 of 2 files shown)


io/disk/tiobench.py
```diff
@@ -64,10 +65,21 @@ def test(self):
...
         size = self.params.get('size', default=1024)
         num_runs = self.params.get('numruns', default=2)
+
+        part_obj = Partition(disk, mountpoint=target)
+        self.log.info("Test will run on %s" % target)
+        self.log.info("Unmounting the disk/dir before creating file system")
+        part_obj.unmount()
+        self.log.info("creating file system")
+        part_obj.mkfs(fstype)
+        self.log.info("Mounting disk %s on directory %s" % (disk, target))
```
apahim: Please notice that the recommended syntax for message string formatting in loggers is:
```python
        self.log.info("Mounting disk %s on directory %s", disk, target)
```
(https://docs.python.org/2/library/logging.html#logging.Logger.debug)

Naresh-ibm: done the changes. Thanks you.

PraveenPenguin:
@apahim thanks

PR #297, opened 2017-03-02
Fix a typo in link used for filebench tarball download.

sacsant:
A typo in link used for download of filebench tarball causes test failure.

### DIFF
--- a/fs/filebench.py
+++ b/fs/filebench.py
@@ -49,8 +49,8 @@ class Filebench(Test):
         self._testfile = self.params.get('testfile', default='fileserver.f')
 
         tarball = self.fetch_asset('https://github.com/filebench/'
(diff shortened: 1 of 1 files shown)


fs/filebench.py
```diff
@@ -49,7 +49,7 @@ def setUp(self):
         self._testfile = self.params.get('testfile', default='fileserver.f')
 
         tarball = self.fetch_asset('https://github.com/filebench/'
-                                   'filebench/releases/ownload/1.5-alpha3/'
+                                   'filebench/releases/download/1.5-alpha3/'
```
PraveenPenguin: ok , one more help as you sending patch can you please remove  `expire`  pram as well as ` filebench-1.5-alpha3.tar.gz` not going to update so no need to download again

PR #299, opened 2017-03-02
Added thp defrag test

santhosh93:
This test turns off the defrag and fragments the memory using dd
workload and then turns on the defrag.And finally, checks whether
the defrag has occured.

### DIFF
--- a/memory/transparent_hugepages_defrag.py
+++ b/memory/transparent_hugepages_defrag.py
@@ -1,0 +1,121 @@
+#!/usr/bin/env python
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+# Test Inspired basically from
+# https://github.com/autotest/tp-qemu/blob/master/generic/tests/trans_hugepage_defrag.py
+#
+# Copyright: 2017 IBM
+# Author: Santhosh G <santhog4@linux.vnet.ibm.com>
+
+import os
+import time
+import mmap
+import avocado
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+from avocado.utils import memory
+from avocado.utils import disk
+from avocado.core import data_dir
+from avocado.utils.partition import Partition
+
+
+class Thp_Defrag(Test):
+
+    '''
+    Defrag test enables THP and fragments the system memory using dd load
+    and turns on THP defrag and checks whether defrag occured.
+    '''
+
+    def setUp(self):
+
+        '''
+        Sets required params for dd workload and mounts the tmpfs
+        '''
+
+        # Get required mem info
+        self.mem_path = os.path.join(data_dir.get_tmp_dir(), 'thp_space')
+        self.block_size = int(mmap.PAGESIZE) / 1024
+        self.dd_timeout = self.params.get('dd_timeout', default=900)
+        # add mount point
+        os.mkdir(self.mem_path)
+        self.device = Partition(device="none", mountpoint=self.mem_path)
+        self.device.mount(mountpoint=self.mem_path, fstype="tmpfs")
+        free_space = (disk.freespace(self.mem_path)) / 1024
+        # Leaving out some free space in tmpfs
+        self.count = (free_space / self.block_size) - 3
+
+    @avocado.fail_on
+    def test(self):
+
+        '''
+        Enables THP, Turns off the defrag and fragments the memory.
+        Once the memory gets fragmented turns on the defrag and checks
+        whether defrag happened.
+        '''
+
+        # Enables THP
+        memory.set_thp_value("enabled", "always")
+
+        # Turns off Defrag
+        memory.set_thp_value("khugepaged/defrag", "0")
+
+        # Fragments The memory
+        for iterator in range(self.count):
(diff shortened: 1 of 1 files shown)


memory/transparent_hugepages_defrag.py
```diff
@@ -0,0 +1,121 @@
...
+    def setUp(self):
+
+        '''
+        Sets required params for dd workload and mounts the tmpfs
+        '''
+
+        # Get required mem info
+        self.mem_path = os.path.join(data_dir.get_tmp_dir(), 'thp_space')
+        self.block_size = int(mmap.PAGESIZE) / 1024
+        self.dd_timeout = self.params.get('dd_timeout', default=900)
```
PraveenPenguin: `self.dd_timeout`  no need to fetch through `params.get` as test not have any input from yaml

memory/transparent_hugepages_defrag.py
```diff
@@ -0,0 +1,121 @@
...
+    def test(self):
+
+        '''
+        Enables THP, Turns off the defrag and fragments the memory.
+        Once the memory gets fragmented turns on the defrag and checks
+        whether defrag happened.
+        '''
+
+        # Enables THP
+        memory.set_thp_value("enabled", "always")
```
PraveenPenguin: it this section should be enclose with try/expect as if library fails with exception should be handle here

memory/transparent_hugepages_defrag.py
```diff
@@ -0,0 +1,121 @@
...
+
+        # Turns off Defrag
+        memory.set_thp_value("khugepaged/defrag", "0")
```
PraveenPenguin: here as well

PraveenPenguin: sample can be 

```
try:
 memory.set_thp_value("enabled", "always")
 memory.set_thp_value("khugepaged/defrag", "0")
except Exception as ex:
  self.fail(ex)
```

memory/transparent_hugepages_defrag.py
```diff
@@ -0,0 +1,121 @@
...
+
+        # Sets max possible hugepages before defrag on
+        memory.set_num_huge_pages(nr_full)
+        nr_hp_before = memory.get_num_huge_pages()
+
+        # Turns Defrag ON
+        memory.set_thp_value("khugepaged/defrag", "1")
+
+        sleep_time = 10
+        self.log.info("Sleeping %d seconds to settle out things", sleep_time)
```
PraveenPenguin: why this can be like 
```

self.log.info("Sleeping %d seconds to settle out things", 10)
 time.sleep(10)
```

memory/transparent_hugepages_defrag.py
```diff
@@ -0,0 +1,121 @@
...
+        time.sleep(sleep_time)
+
+        # Sets max hugepages after defrag on
+        memory.set_num_huge_pages(nr_full)
+        nr_hp_after = memory.get_num_huge_pages()
+
```
PraveenPenguin: why not this can be moved as helper function as it use two times ?

memory/transparent_hugepages_defrag.py
```diff
@@ -0,0 +1,121 @@
...
+        # Check for memory defragmentation
+        if nr_hp_before >= nr_hp_after:
+            e_msg = "No Memory Defragmentation\n"
+            e_msg += "%d hugepages before turning khugepaged on,\n"\
+                     "%d After it" % (nr_hp_before, nr_hp_after)
+            self.error(e_msg)
```
PraveenPenguin: i think this is kind of indication of test fail is it ?

memory/transparent_hugepages_defrag.py
```diff
@@ -0,0 +1,121 @@
...
+        total = memory.memtotal()
+        hugepagesize = memory.get_huge_page_size()
+        nr_full = int(0.8 * (total/hugepagesize))
+
+        # Sets max possible hugepages before defrag on
+        memory.set_num_huge_pages(nr_full)
+        nr_hp_before = memory.get_num_huge_pages()
+
+        # Turns Defrag ON
+        memory.set_thp_value("khugepaged/defrag", "1")
```
PraveenPenguin: here also try/expect enclose required

memory/transparent_hugepages_defrag.py
santhosh93: Handled using @avocado.fail_on ..
To avoid using multiple time try ..catch @avocado.fail_on decorator has been used.

memory/transparent_hugepages_defrag.py
santhosh93: Yes this can be done...

memory/transparent_hugepages_defrag.py
santhosh93: Yes this can be done...

memory/transparent_hugepages_defrag.py
santhosh93: Ok Sure...Let me work on it and add it to framework

memory/transparent_hugepages_defrag.py
santhosh93: Thanks for noticing it....it needs to be self.fail...

memory/transparent_hugepages_defrag.py
santhosh93: @PraveenPenguin So can we add a method to the framework like that it configures the max possible hugepages and returns the value how much it configured??

santhosh93: @PraveenPenguin : Got it....we can make it to helper function...thanks

memory/transparent_hugepages_defrag.py
PraveenPenguin: oh ok i miss that thanks for explanation

santhosh93:
Opened a new PR -https://github.com/avocado-framework-tests/avocado-misc-tests/pull/307

PR #313, opened 2017-03-08
Fix Xfstest

PraveenPenguin:
fix xfstest
```
1-enable xfstest to run fedora/centos/RHEL
2-enable local.config file to handle environment variable
3-Rectify README File
```

### DIFF
--- a/fs/xfstests.py
+++ b/fs/xfstests.py
@@ -27,7 +27,7 @@ import shutil
 
 from avocado import Test
 from avocado import main
-from avocado.utils import process, build, git
+from avocado.utils import process, build, git, distro
 from avocado.utils.software_manager import SoftwareManager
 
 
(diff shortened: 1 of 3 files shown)


PraveenPenguin:
Job log
[joblog.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/827092/joblog.txt)

fs/xfstests.py
```diff
@@ -39,14 +39,27 @@ def setUp(self):
...
+        detected_distro = distro.detect()
+
+        packages = ['e2fsprogs', 'automake', 'gcc', 'quota', 'attr',
+                    'make',  'xfsprogs',  'gawk', 'fio', 'dbench']
+
+        if 'Ubuntu' in detected_distro.name:
+            packages.extend(['xfslibs-dev', 'uuid-dev', 'libtool-bin', 'libuuid1',
+                             'libattr1-dev', 'libacl1-dev', 'libgdbm-dev', 'uuid-runtime'])
+
+        elif detected_distro.name in ['centos', 'fedora', 'redhat']:
```
narasimhan-v: Are we not considering suse here ?

PraveenPenguin: @narasimhan-v no

fs/xfstests.py.data/local.config
```diff
@@ -0,0 +1,4 @@
+export TEST_DEV=/dev/loop0
+export TEST_DIR=/mnt/test
+export SCRATCH_DEV=/dev/loop1
+export SCRATCH_MNT=/mnt/scratch
```
narasimhan-v: Does the user need to edit this file frequently ?
If yes, can it be handled with yaml ?

PraveenPenguin: yes it need to edit via user as it require valid input (Device) of test trigger , i think both way is fine as we need to have in  config  so i think editing in yaml than dump those entry config seems not good so it is better to edit config file directly  i think this approach looks me ok as xfstest follow this  ,

fs/xfstests.py
```diff
@@ -39,14 +39,27 @@ def setUp(self):
...
+                             'libaio-devel', 'libuuid-devel', 'openssl-devel', 'xfsprogs-devel',
+                             'btrfs-progs-devel'])
+        else:
+            self.skip("test not supported in %s" % detected_distro.name)
+
         for package in packages:
             if not sm.check_installed(package) and not sm.install(package):
-                self.error("Fail to install %s required for this test." %
-                           package)
+                self.skip("Fail to install %s required for this test." %
```
balamuruhans: shouldn't it be error, as we skip tests if package is not there or if env is unsupported, but when test tries to install some package and if it fails to install its better to error out.

fs/xfstests.py.data/local.config
narasimhan-v: @PraveenPenguin Okay, for now we can have this minimalistic version, with `TODO` added indicating creating FS, mounting the disk/loop devices, etc

fs/xfstests.py
PraveenPenguin: i think should be ok as if these package not installed as these package is not available so not seems me test error  , test is not able to run so i think we can skip the test and in report it kind of skip what you say ?

fs/xfstests.py.data/local.config
PraveenPenguin: @narasimhan-v sure thanks will add in TODO

fs/xfstests.py
balamuruhans: fine then, Looks good otherwise

PraveenPenguin: @balamuruhans thanks

PraveenPenguin:
closing in favour of #314

PR #314, opened 2017-03-08
FIx Xfstest v2

PraveenPenguin:
Fix xfstest

1- enable xfstest to run fedora/centos/RHEL
2- enable local.config file to handle environment variable
3-  Rectify README File


narasimhan-v:
Looks good to me

PR #319, opened 2017-03-10
Added memcached test version 2

santhosh93:
The test runs the memcached in the background
and runs the stress tool on it.

Made changes to work in rhel too.

### DIFF
--- a/memory/memcached.py
+++ b/memory/memcached.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2017 IBM
+# Author: Santhosh G <santhog4@linux.vnet.ibm.com>
+
+import time
+import getpass
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+from avocado.utils import memory
+from avocado.utils import distro
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Memcached(Test):
+
+    """
+    Memcached - High performance memory object caching system.
+    This test Runs the memcached server in the background, based
+    upon the memory param given. And it tries to run the memcslap
+    stress tool which generates a load against memcached server.
+    For more options on memcached:
+    Refer - https://linux.die.net/man/1/memcached
+    """
+
+    def setUp(self):
+
+        """
+        Sets up the args required to run the test.
+        """
+
+        sm = SoftwareManager()
+        detected_distro = distro.detect()
+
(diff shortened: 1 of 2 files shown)


memory/memcached.py
```diff
@@ -0,0 +1,120 @@
...
+        for package in deps:
+            if not sm.check_installed(package) and not sm.install(package):
+                self.error(' %s is needed for the test to be run' % package)
+
+        # Memcached Required Args
+        memory_to_use = self.params.get("memory_to_use",
+                                        default=(memory.freememtotal() / 1024))
+        port_no = self.params.get("port_no", default='12111')
+        memcached_args = self.params.get('memcached_args', default='')
+        self.memcached_cmd = 'memcached -u %s -p %s -m %d  %s &'\
```
nasastry: There are many instances of 'memcached' can it get a variable.

memory/memcached.py.data/1
```diff
@@ -0,0 +1,26 @@
+setup:
```
PraveenPenguin: i think this file is left over can you please remove this

memory/memcached.py.data/memcached.yaml
```diff
@@ -0,0 +1,26 @@
...
+            memcached_args: ''
+            system_ip: '127.0.0.1'
+            test_to_run: !mux
+                set:
+                    test_to_run: 'set'
+                get:
+                    test_to_run: 'get'
+            concurrency: '100'
+            stress_tool_args: ''
+
```
PraveenPenguin: please remove this white space

memory/memcached.py.data/memcached.yaml
```diff
@@ -0,0 +1,26 @@
+setup:
+    general: !mux
+        default:
```
PraveenPenguin: i still tab format is there can you  please convert tab to spcae

memory/memcached.py
```diff
@@ -0,0 +1,120 @@
...
+        detected_distro = distro.detect()
+
+        if detected_distro.name not in ['Ubuntu', 'redhat']:
+            self.skip('Test Not applicable')
+
+        if detected_distro.name == "Ubuntu":
+            deps = ['memcached', 'libmemcached-tools']
+            stress_tool = 'memcslap'
+
+        if detected_distro.name == "redhat":
```
PraveenPenguin: why not this can be run on fedora / centos ?

memory/memcached.py
PraveenPenguin: @santhosh93 have you like retain this ?

santhosh93: @nasastry : i did not get exactly the change you want me to do.

santhosh93: @PraveenPenguin : Yes . Even using an variable for memcached ...we need to use that variable through out the test again with diff name.so i think it is not reqd.

memory/memcached.py
santhosh93: As of now verified in redhat ...may be later support for the distros can be verfified and added.

memory/memcached.py
PraveenPenguin: sure

memory/memcached.py
PraveenPenguin: make sense

PraveenPenguin:
@nasastry thanks

PR #321, opened 2017-03-13
Fix valid smt values in ppc64_cpu

narasimhan-v:
ppc64_cpu supports smt values 0 to 8 for power8, and 0 to 4 for
other power processor levels.
Fixed that in the test.

### DIFF
--- a/cpu/ppc64_cpu_test.py
+++ b/cpu/ppc64_cpu_test.py
@@ -22,6 +22,7 @@ import os
 from avocado import Test
 from avocado import main
 from avocado.utils import process
+from avocado.utils import cpu
 from avocado.utils.software_manager import SoftwareManager
 
 
@@ -52,6 +53,9 @@ class PPC64Test(Test):
         self.smt_values = {1: "off"}
         self.key = 0
         self.value = ""
+        self.max_smt_value = 4
+        if cpu.get_cpu_arch().lower() == 'power8':
+            self.max_smt_value = 8
 
     def equality_check(self, test_name, cmd1, cmd2):
         """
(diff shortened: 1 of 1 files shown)


narasimhan-v:
[job_log.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/837286/job_log.txt)

cpu/ppc64_cpu_test.py
```diff
@@ -52,6 +53,9 @@ def setUp(self):
         self.smt_values = {1: "off"}
         self.key = 0
         self.value = ""
+        self.max_smt_value = 4
+        if cpu.get_cpu_arch().lower() == 'power8':
+            self.max_smt_value = 8
```
nasastry: Can you make this change to accommodate the next cpu architecture also.

PraveenPenguin: i think should be fine as  power7 and lower architecture  will not support smt level 8 it is just 4 , and other arch this test not supported

nasastry: What I mean of 'next cpu architecture' is for Power9. If this test case is run on Power9 it will take max_smt_value as '4'.

narasimhan-v: @nasastry IIRC, for power9, SMT values only upto 4 are supported. So this should be fine.

PraveenPenguin: @narasimhan-v thanks for clarification

PraveenPenguin:
@nasastry thanks

abdhaleegit:
Thanks @narasimhan-v  for the fix

PR #324, opened 2017-03-13
Added missing packages in tests

harish-24:
Unixbench test fails without patch command. Valgrind requires g++ package.

### DIFF
--- a/toolchain/valgrind.py
+++ b/toolchain/valgrind.py
@@ -18,6 +18,7 @@ from avocado import main
 from avocado.utils import process
 from avocado.utils import build
 from avocado.utils import archive
+from avocado.utils import distro
 from avocado.utils.software_manager import SoftwareManager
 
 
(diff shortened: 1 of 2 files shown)


toolchain/valgrind.py
```diff
@@ -34,6 +35,12 @@ def setUp(self):
         smm = SoftwareManager()
         self.failures = []
 
+        dist = distro.detect()
+        deps = ['gcc', 'make']
+        if dist.name == 'Ubuntu':
+            deps.extend(['g++'])
+        elif dist.name in ['SuSe', 'redhat', 'fedora']:
+            deps.extend(['gcc-c++'])
         for package in ['gcc', 'make']:
```
nasastry: I guess here you need to use "deps" instead of ['gcc', 'make']

nasastry: I guess here you need to use "deps" instead of ['gcc', 'make']

for package in deps:

PraveenPenguin: please consider centos as well

harish-24: @nasastry Sure, missed that.

PraveenPenguin:
@nasastry thanks

PR #325, opened 2017-03-14
Added skip check for powernv platform

harish-24:
Most test failed without support on PowerNV platform.This patch skips the unsupported tests.

### DIFF
--- a/generic/ras.py
+++ b/generic/ras.py
@@ -40,6 +40,14 @@ class RASTools(Test):
         architecture = os.uname()[4]
         if "ppc" not in architecture:
             self.skip("supported only on Power platform")
+        if 'PowerNV' in process.system_output('grep ^platform /proc/cpuinfo'):
+            nv_flag = False
+            for cmd in ['lsmcode', 'lsprop', 'nvram', 'ppc64_cpu']:
+                if cmd in self.name.name:
+                    nv_flag = True
+            if not nv_flag:
+                self.skip('Test not supported in PowerNV platform')
+
         sm = SoftwareManager()
         for package in ("ppc64-diag", "powerpc-utils", "lsvpd"):
             if not sm.check_installed(package) and not sm.install(package):


generic/ras.py
```diff
@@ -40,6 +40,14 @@ def setUp(self):
         architecture = os.uname()[4]
         if "ppc" not in architecture:
             self.skip("supported only on Power platform")
+        if 'PowerNV' in process.system_output('grep ^platform /proc/cpuinfo'):
+            nv_flag = False
+            for cmd in ['lsmcode', 'lsprop', 'nvram', 'ppc64_cpu']:
+                if cmd in self.name.name:
```
nasastry: self.name.name variable I didn't see any where in 'ras.py' 
is this a special variable or a typo?

generic/ras.py
```diff
@@ -40,6 +40,14 @@ def setUp(self):
         architecture = os.uname()[4]
         if "ppc" not in architecture:
             self.skip("supported only on Power platform")
+        if 'PowerNV' in process.system_output('grep ^platform /proc/cpuinfo'):
```
apahim: Could you please use the python's builtin `open()`? It's way more efficient for that operation:
```
$ time python -c "import avocado; print 'GenuineIntel' in avocado.utils.process.system_output('grep GenuineIntel /proc/cpuinfo')"
True

real    0m0.178s
user    0m0.155s
sys     0m0.023s
```
```
$ time python -c "print 'GenuineIntel' in open('/proc/cpuinfo', 'r').read()"
True

real    0m0.033s
user    0m0.027s
sys     0m0.006s
```

generic/ras.py
```diff
@@ -40,6 +40,14 @@ def setUp(self):
...
+            nv_flag = False
+            for cmd in ['lsmcode', 'lsprop', 'nvram', 'ppc64_cpu']:
```
apahim: Let me see if I got this correctly... It means that on `PowerNV` platform you will only run these tests: 'lsmcode', 'lsprop', 'nvram', 'ppc64_cpu'?

generic/ras.py
apahim: If that's the case, `self.name.name` will contain a long string with path, module name, class name and method name, like `/home/apahim/avocado/tests/test.py:MyTest.test`. Very prone to break this check.
On the other hand, `self._testMethodName` will have only the method name.

apahim: Anyway, if you need granularity here, I'd instead suggest you to use the skip decorators. Example:
```python
from avocado import Test
from avocado import skipUnless

IS_POWER_NV = 'PowerNV' in open('/proc/cpuinfo', 'r').read()

class MyTest(Test):
    def setUp(self):
        pass

    @skipUnless(IS_POWER_NV, "Skipping test in non-PowerNV platform")
    def test3_lsmcode(self):
        pass

    @skipUnless(IS_POWER_NV, "Skipping test in non-PowerNV platform")
    def test5_lsprop(self):
        pass
```

apahim: Oh, my logic might be broken... you will probably need something more like:

```python

from avocado import Test
from avocado import skipIf
from avocado import skipUnless

IS_POWER_NV = 'PowerNV' in open('/proc/cpuinfo', 'r').read()

class MyTest(Test):
    def setUp(self):
        pass

    @skipIf(IS_POWER_NV, "Skipping test in PowerNV platform")
    def test1_set_poweron_time(self):
        pass

    ...

    @skipUnless(IS_POWER_NV, "Skipping test in non-PowerNV platform")
    def test3_lsmcode(self):
        pass

    @skipUnless(IS_POWER_NV, "Skipping test in non-PowerNV platform")
    def test5_lsprop(self):
        pass
```

But you got my point :)

generic/ras.py
harish-24: Not sure there is not much of difference. 
```
$ time python -c "from avocado.utils import process"

real	0m0.175s
user	0m0.139s
sys	0m0.035s

$ time python -c "from avocado.utils import process; print 'GenuineIntel' in process.system_output('grep GenuineIntel /proc/cpuinfo')"
True

real	0m0.182s
user	0m0.151s
sys	0m0.030s
```
With the difference, both takes same time
```
$ time python -c "print 'GenuineIntel' in open('/proc/cpuinfo', 'r').read()"
True

real	0m0.010s
user	0m0.007s
sys	0m0.003s
```

generic/ras.py
harish-24: exactly

generic/ras.py
harish-24: sure, will replace this

generic/ras.py
PraveenPenguin: @harish-24 if i go through the logs you are getting performance improve

apahim: It makes sense. Anyway, we also have an `avocado.utils.cpu`. If I had to choose a way to handle this `/proc/cpuinfo` search, `process.system_output('grep ...')` would be my last option. But this, of course, is just a suggestion. Your code, your call :)

harish-24:
Closing this. Opened a new PR #327

PR #328, opened 2017-03-15
Added perf test and perf-test suite.Providing fix to install perf tool.

shriyak:
Perf test assorted sanity tests, initially through linked routines but also will look for a directory with more tests in the form of scripts.
perf-test suite contains series of test to validate that the perf_event subsystem is working.
Fix will install perf using linux-tools-common and linux-tools-<kernel-version>.

### DIFF
--- a/perf/perftool.py
+++ b/perf/perftool.py
@@ -20,7 +20,7 @@ import os
 
 from avocado import Test
 from avocado import main
-from avocado.utils import archive, build
+from avocado.utils import archive, build, process
 from avocado.utils.software_manager import SoftwareManager
 
 
@@ -77,8 +77,7 @@ class Perftool(Test):
         self.srcdir = os.path.join(self.srcdir, 'perf_event_tests-master')
         build.make(self.srcdir)
         os.chdir(self.srcdir)
-        process.system_output("echo -1 >/proc/sys/kernel/"
-                              "perf_event_paranoid")
+        os.system("echo -1 >/proc/sys/kernel/perf_event_paranoid")
         process.system_output("./run_tests.sh")
 
 
--- a/perf/perftool.py
+++ b/perf/perftool.py
@@ -35,13 +35,22 @@ class Perftool(Test):
         Build perftool Test
         Source:
         https://github.com/rfmvh/perftool-testsuite
+        Run perf test
+        Execute perf tests
+        Source : https://github.com/deater/perf_event_tests
+
         '''
 
         # Check for basic utilities
         smm = SoftwareManager()
-        for package in ['gcc', 'make', 'perf']:
+        for package in ['gcc', 'make']:
             if not smm.check_installed(package) and not smm.install(package):
                 self.error('%s is needed for the test to be run' % package)
+        if SoftwareManager().check_installed("linux-tools-common") is False:
+            if SoftwareManager().install("linux-tools-common") is False:
+                self.skip("linux-tools-common is not installing")
+        if SoftwareManager().install("linux-tools-$(uname-r)") is False:
+            self.skip("linux-tools-$(uname-r) is not installing")
 
         locations = ["https://github.com/rfmvh/perftool-testsuite/archive/"
                      "master.zip"]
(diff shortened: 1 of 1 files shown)


perf/perftool.py
```diff
@@ -35,13 +35,22 @@ def setUp(self):
...
+
         '''
 
         # Check for basic utilities
         smm = SoftwareManager()
-        for package in ['gcc', 'make', 'perf']:
+        for package in ['gcc', 'make']:
             if not smm.check_installed(package) and not smm.install(package):
                 self.error('%s is needed for the test to be run' % package)
+        if SoftwareManager().check_installed("linux-tools-common") is False:
```
nasastry: Please use smm.check_installed instead of SoftwareManager().check_installed

perf/perftool.py
```diff
@@ -35,13 +35,22 @@ def setUp(self):
...
+            if SoftwareManager().install("linux-tools-common") is False:
```
nasastry: same here

perf/perftool.py
```diff
@@ -35,13 +35,22 @@ def setUp(self):
...
+                self.skip("linux-tools-common is not installing")
+        if SoftwareManager().install("linux-tools-$(uname-r)") is False:
```
nasastry: same here

perf/perftool.py
```diff
@@ -57,10 +66,21 @@ def test(self):
             if '-- [ FAIL ] --' in line:
                 self.count += 1
                 self.log.info(line)
-
         if self.count > 0:
             self.fail("%s Test failed" % self.count)
 
+        process.system_output("perf test")
```
nasastry: process is not defined any where, please add import statement
from avocado.utils import process

perf/perftool.py
```diff
@@ -57,10 +66,21 @@ def test(self):
...
+        tarball = self.fetch_asset('https://github.com/deater/'
+                                   'perf_event_tests/archive/'
+                                   'master.zip', expire='7d')
+        archive.extract(tarball, self.srcdir)
+        self.srcdir = os.path.join(self.srcdir, 'perf_event_tests-master')
+        build.make(self.srcdir)
+        os.chdir(self.srcdir)
+        process.system_output("echo -1 >/proc/sys/kernel/"
+                              "perf_event_paranoid")
+        process.system_output("./run_tests.sh")
```
narasimhan-v: Wouldn't `process.run` here be better ?

perf/perftool.py
```diff
@@ -35,13 +35,22 @@ def setUp(self):
...
+        Run perf test
+        Execute perf tests
+        Source : https://github.com/deater/perf_event_tests
+
         '''
 
         # Check for basic utilities
         smm = SoftwareManager()
-        for package in ['gcc', 'make', 'perf']:
+        for package in ['gcc', 'make']:
```
PraveenPenguin: just add `linux-tools-common` in package than other code is required

perf/perftool.py
narasimhan-v: Please add `linux-tools-common` to `['gcc', 'make']`

perf/perftool.py
PraveenPenguin: can we validate  any output here

PraveenPenguin:
@shriyak please import process in test code it will make Travis happy :)

perf/perftool.py
```diff
@@ -57,10 +66,21 @@ def test(self):
...
+                                   'perf_event_tests/archive/'
+                                   'master.zip', expire='7d')
+        archive.extract(tarball, self.srcdir)
+        self.srcdir = os.path.join(self.srcdir, 'perf_event_tests-master')
+        build.make(self.srcdir)
+        os.chdir(self.srcdir)
+        process.system_output("echo -1 >/proc/sys/kernel/"
+                              "perf_event_paranoid")
+        process.system_output("./run_tests.sh")
+
```
narasimhan-v: You are running 2 tests here: perftool and perf_event_tests.
Make sure the failure/error are indicated better for each of them seperately, instead of generic "Test Failed".
I would suggest having seperate `test_` functions for each of them.

perf/perftool.py
PraveenPenguin: sorry my mistake just not i realize that linux-tools-common is required for ubutnu only and perf is required for all distro i think we should have check to handle that , as this won't work

shriyak: yes, I will make a check and include both.

perf/perftool.py
shriyak: Sure let me add these changes .

perf/perftool.py
shriyak: yes, we can validate the output praveen.
let me include process.run

perf/perftool.py
shriyak: yes, I missed out .

perf/perftool.py
shriyak: Instead of doing Software manager separately, I add it here

PR #337, opened 2017-03-17
Hwinfo 2

Pavithra1602:
fixed Pep8 compliant issue.

### DIFF
--- a/generic/hwinfo.py
+++ b/generic/hwinfo.py
@@ -1,0 +1,134 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Pavithra <pavrampu@linux.vnet.ibm.com>
+
+import os
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+from avocado.utils import distro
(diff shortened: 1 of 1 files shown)


generic/hwinfo.py
```diff
@@ -0,0 +1,134 @@
...
+        self.run_cmd("hwinfo --tv")
+        self.run_cmd("hwinfo --uml")
+        self.run_cmd("hwinfo --usb")
+        self.run_cmd("hwinfo --usb-ctrl")
+        self.run_cmd("hwinfo --vbe")
+        self.run_cmd("hwinfo --wlan")
+        self.run_cmd("hwinfo --xen")
+        self.run_cmd("hwinfo --zip")
+        self.run_cmd("hwinfo --short")
+        self.run_cmd("hwinfo --listmd")
```
abdhaleegit: a better way to handle all above options is to use list: like below

options = ['all', 'arch',  . . . . 'sys', 'usb', 'sound', 'tape', 'tv', 'scsi', 'listmd']
for ops in options:
    cmd = "hwinfo  --%s" % ops
    self.run_cmd(cmd)

generic/hwinfo.py
```diff
@@ -0,0 +1,134 @@
...
+        disk_name = process.system_output("df -h | egrep '(s|v)d[a-z][1-8]' | "
+                                          "tail -1 | cut -d' ' -f1",
```
abdhaleegit: above command has extra double qoute, which may fail.
Why don't we use 'disk' module in avocado utils to get the disk node ? Or
Pass the disk from yaml file, which is always good.  self.disk = self.params.get('disk')

generic/hwinfo.py
nasastry: Already given this feedback in other PR opened.

generic/hwinfo.py
nasastry: In my view `double-quote` will not be an issue.

Pavithra1602:
Raised new PR.
https://github.com/avocado-framework-tests/avocado-misc-tests/pull/338

PR #347, opened 2017-03-21
Update ppc64_cpu test with processor generation check.

sacsant:
On a POWER6 based system smt level upto 2 is supported, while on a POWER7
based system smt level upto 4 is supported. Depending on the processor
generation some of the smt tests fails.

This patch adds a check for smt level support and limits the runtime values.

### DIFF
--- a/cpu/ppc64_cpu_test.py
+++ b/cpu/ppc64_cpu_test.py
@@ -53,9 +53,11 @@ class PPC64Test(Test):
         self.smt_values = {1: "off"}
         self.key = 0
         self.value = ""
-        self.max_smt_value = 4
-        if cpu.get_cpu_arch().lower() == 'power8':
-            self.max_smt_value = 8
+        self.max_smt_value = 8
+        if cpu.get_cpu_arch().lower() == 'power7':
+            self.max_smt_value = 4
+        if cpu.get_cpu_arch().lower() == 'power6':
+            self.max_smt_value = 2
 
     def equality_check(self, test_name, cmd1, cmd2):
         """


cpu/ppc64_cpu_test.py
```diff
@@ -53,9 +53,11 @@ def setUp(self):
...
         self.key = 0
         self.value = ""
-        self.max_smt_value = 4
-        if cpu.get_cpu_arch().lower() == 'power8':
-            self.max_smt_value = 8
+        self.max_smt_value = 8
+        if cpu.get_cpu_arch().lower() == 'power7':
+            self.max_smt_value = 4
+        if cpu.get_cpu_arch().lower() == 'power6':
+            self.max_smt_value = 2
```
narasimhan-v: @sacsant I think we need to handle power9 as well here, where `max_smt_value` is not 8

PraveenPenguin: @narasimhan-v i think we an  push p9 support when it is required , till now should be fine

narasimhan-v: @sacsant @PraveenPenguin I am fine if we have a placeholder / skip for power9, and later add it when the hardware comes ?

sacsant: @narasimhan-v the current code should work on p9 as well.

narasimhan-v: @sacsant Fine then.

PraveenPenguin:
@narasimhan-v @nasastry thanks

PR #350, opened 2017-03-23
Added setUp to sensors test

harish-24:
Without setup test errors out in case of skip. Added setUp to address it.

### DIFF
--- a/cpu/sensors.py
+++ b/cpu/sensors.py
@@ -48,9 +48,9 @@ class Sensors(Test):
                 errs.append(error)
         return errs
 
-    def test(self):
+    def setUp(self):
         """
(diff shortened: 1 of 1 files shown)


cpu/sensors.py
```diff
@@ -48,10 +48,7 @@ def check_errors(cmd):
                 errs.append(error)
         return errs
 
-    def test(self):
-        """
-        Test for sensors command
-        """
+    def setUp(self):
```
narasimhan-v: Docstring missing

harish-24: Added

PraveenPenguin:
@narasimhan-v thanks

PR #354, opened 2017-03-23
Update tar for interbench

harish-24:
Tar is not accessible. Interbench tar is updated.

### DIFF
--- a/generic/interbench.py
+++ b/generic/interbench.py
@@ -47,9 +47,9 @@ class Interbench(Test):
         if (not sm_manager.check_installed("gcc") and not
                 sm_manager.install("gcc")):
             self.error("Gcc is needed for the test to be run")
-        tarball = self.fetch_asset('http://ck.kolivas.org'
-                                   '/apps/interbench/'
-                                   'interbench-0.31.tar.bz2')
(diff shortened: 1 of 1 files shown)


generic/interbench.py
```diff
@@ -47,9 +47,9 @@ def setUp(self):
         if (not sm_manager.check_installed("gcc") and not
                 sm_manager.install("gcc")):
             self.error("Gcc is needed for the test to be run")
-        tarball = self.fetch_asset('http://ck.kolivas.org'
-                                   '/apps/interbench/'
-                                   'interbench-0.31.tar.bz2')
+        tarball = self.fetch_asset('http://slackware.cs.utah.edu/pub/kernel'
+                                   '.org/pub/linux/kernel/people/ck/apps/'
+                                   'interbench/interbench-0.31.tar.gz')
```
PraveenPenguin: why this as i able to fetch package from both of URL , can you please help me about this

harish-24: The old link is not accessible across servers.

PraveenPenguin: oh my bad luck for my network it works so confused , any way both is working for me and new URL works for you so i think we are good to have

PR #355, opened 2017-03-24
Added perf test and providing fix to install perf packages.

shriyak:
Perf test is the built in tool which executes assorted sanity tests, initially through linked routines
but also will look for a directory with more tests in the form of scripts.
Fix will install perf using linux-tools-common and linux-tools for Ubuntu distro.

### DIFF
--- a/perf/perftool.py
+++ b/perf/perftool.py
@@ -17,10 +17,10 @@
 
 
 import os
-
+import platform
 from avocado import Test
 from avocado import main
-from avocado.utils import archive, build
+from avocado.utils import archive, build, distro, process
 from avocado.utils.software_manager import SoftwareManager
 
 
@@ -39,7 +39,15 @@ class Perftool(Test):
 
         # Check for basic utilities
         smm = SoftwareManager()
-        for package in ['gcc', 'make', 'perf']:
+        detected_distro = distro.detect()
+        kernel_ver = platform.uname()[2]
+        if 'Ubuntu' in detected_distro.name:
+            deps = ['gcc', 'make', 'linux-tools-common',
+                    'linux-tools-' + kernel_ver + '']
+        else:
+            deps = ['gcc', 'make', 'perf']
+
+        for package in deps:
             if not smm.check_installed(package) and not smm.install(package):
                 self.error('%s is needed for the test to be run' % package)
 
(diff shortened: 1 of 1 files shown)


perf/perftool.py
```diff
@@ -39,7 +39,15 @@ def setUp(self):
 
         # Check for basic utilities
         smm = SoftwareManager()
-        for package in ['gcc', 'make', 'perf']:
+        detected_distro = distro.detect()
+        kernel_ver = platform.uname()[2]
+        if 'Ubuntu' in detected_distro.name:
+            deps = ['gcc', 'make', 'linux-tools-common',
```
PraveenPenguin: can gcc, make package is kind pre defined in deps as it is kind of common 

ex: 
```
deps=[''gcc', 'make'']
if 'Ubuntu' in detected_distro.name:
 deps.extend(['linux-tools-common',  'linux-tools-%s' % kernel_ver ])
else:
   deps.append('perf')
```

perf/perftool.py
```diff
@@ -51,6 +59,19 @@ def setUp(self):
...
 
     def test(self):
+        '''
+        perf test :Does sanity tests and
+        execute the tests by calling each module
+        '''
+        # Built in perf test
+        result = process.run("perf test")
+        x = result.stderr.splitlines()
+        for string in x:
```
PraveenPenguin: can this be like 
```
    for failedString in process.run("perf test").stderr.splitlines():
           if 'FAILED' in str(failedString):
                   self.fail("Test case failed is %s", y.strip("[]")
```

just eliminate extra variables and self.log to self.fail as i think in this check we need to fail the test

PraveenPenguin: can this be like 
```
    for failedString in process.run("perf test").stderr.splitlines():
           if 'FAILED' in str(failedString):
                   self.fail("Test case failed is %s", y.strip("[]")
```

just eliminate extra variables and self.log to self.fail as i think in this check we need to fail the test

perf/perftool.py
```diff
@@ -39,7 +39,15 @@ def setUp(self):
...
+        kernel_ver = platform.uname()[2]
+        if 'Ubuntu' in detected_distro.name:
+            deps = ['gcc', 'make', 'linux-tools-common',
+                    'linux-tools-' + kernel_ver + '']
+        else:
+            deps = ['gcc', 'make', 'perf']
+
+        for package in deps:
             if not smm.check_installed(package) and not smm.install(package):
                 self.error('%s is needed for the test to be run' % package)
```
PraveenPenguin: one request as you submitting changes can you please change  `self.error` to` self.skip`

perf/perftool.py
shriyak: Sure, let me make these changes.

perf/perftool.py
shriyak: Yes, let me update it

shriyak:
Raised new PR

PR #356, opened 2017-03-24
Print error log properly

narasimhan-v:
Printing errors in self.fail, so that it is easier to view.

### DIFF
--- a/io/disk/multipath_test.py
+++ b/io/disk/multipath_test.py
@@ -119,8 +119,7 @@ class MultipathTest(Test):
         # Print errors
         if msg:
             msg = "Following Tests Failed\n" + msg
-            self.log.debug(msg)
-            self.fail("Some tests failed")
+            self.fail("Some tests failed. Find details below:\n%s", msg)
 
     def tearDown(self):
         """


io/disk/multipath_test.py
```diff
@@ -119,8 +119,7 @@ def test(self):
         # Print errors
         if msg:
             msg = "Following Tests Failed\n" + msg
-            self.log.debug(msg)
-            self.fail("Some tests failed")
+            self.fail("Some tests failed. Find detaild below:\n%s", msg)
```
balamuruhans: %s/detaild/details and `self.fail("Some tests failed. Find details below:\n%s" % msg)`

narasimhan-v: Typo will be corrected and will force push.
It is always preferred in logging to have `"some text %s", string`. So, that would stay.

balamuruhans: @narasimhan-v for `self.log` we use to follow as you say, but for `self.fail` I thought we will use %

PraveenPenguin: i think both way is fine , @balamuruhans are you ok with this ?

balamuruhans: @PraveenPenguin It got this suggestion from community which I shared here, I am okay.

PraveenPenguin: @balamuruhans yes thanks for sharing this that is very helpful ,again style issue can be depend on coder , if you are fine will accept this patch

PraveenPenguin:
@balamuruhans thanks for review

io/disk/multipath_test.py
narasimhan-v: @PraveenPenguin @balamuruhans pylint does not complain for both, so both are fine. And I do not prefer to push again ;-) .

PraveenPenguin: @narasimhan-v :)

PR #359, opened 2017-03-24
Added pthread deps to install

santhosh93:
And also changed self.error to self.skip for
package dependencies.

### DIFF
--- a/memory/libhugetlbfs.py
+++ b/memory/libhugetlbfs.py
@@ -41,19 +41,21 @@ class libhugetlbfs(Test):
         # Check for basic utilities
         sm = SoftwareManager()
         detected_distro = distro.detect()
-        if not sm.check_installed("gcc") and not sm.install("gcc"):
-            self.error('Gcc is needed for the test to be run')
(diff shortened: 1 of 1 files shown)


memory/libhugetlbfs.py
```diff
@@ -41,19 +41,23 @@ def setUp(self):
...
+            deps += ['libpthread-stubs0-dev']
+        elif detected_distro.name == "SuSE":
+            deps += ['glibc-devel-static']
         else:
-            if not sm.check_installed("git") and not sm.install("git"):
-                self.error('git is needed for the test to be run')
+            deps += ['glibc-static']
+
+        for package in deps:
+            if detected_distro.name == "SuSE" and package == 'git':
```
narasimhan-v: It is better to do this check before, where you are doing a distro check for pthread and glibc.

santhosh93: Ok will do. thanks..

PraveenPenguin:
@narasimhan-v thanks

PR #361, opened 2017-03-24
Perf test suite to validate the perf interface 

shriyak:
Perf test suite to validate the perf interface 

This is a series of test is meant to validate that the perf_event subsystem is working.
It supports features like perf fuzzer, that tries to hit the extremes of the perf_event interface and find bugs in the kernel perf_event_h, an archive of perf_event.h files over the history of the interface.Useful when figuring out when a feature was added. It also supports various perf_event related tools and also tests PAPI library if it is supported.

### DIFF
--- a/perf/perf_events_test.py
+++ b/perf/perf_events_test.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


PraveenPenguin:
@shriyak can you please squash three  commits  in two commits 

1- 	Added perf test and providing fix to install perf packages

2-  	Testsuite to validate perf_event_subsystem

but i prefer that raise to tow PR that it help to review

shriyak:
Raised new PR

PR #362, opened 2017-03-24
Removed the thp directory

santhosh93:
Did proper cleanup after execution of tests

### DIFF
--- a/memory/transparent_hugepages_defrag.py
+++ b/memory/transparent_hugepages_defrag.py
@@ -121,8 +121,8 @@ class Thp_Defrag(Test):
         if self.mem_path:
             self.log.info('Cleaning Up!!!')
             memory.set_num_huge_pages(0)
-            process.system('rm -rf %s/*' % self.mem_path, ignore_status=True)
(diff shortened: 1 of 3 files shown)


memory/transparent_hugepages_defrag.py
```diff
@@ -121,8 +121,8 @@ def tearDown(self):
         if self.mem_path:
             self.log.info('Cleaning Up!!!')
             memory.set_num_huge_pages(0)
-            process.system('rm -rf %s/*' % self.mem_path, ignore_status=True)
             self.device.unmount()
+            process.system('rm -rf %s' % self.mem_path, ignore_status=True)
```
PraveenPenguin: can you please help me to understand this why it is required ?

santhosh93: Yes...before the directory contents only were removed and the directory remained as such.
So if other thp tests ran with same job the tests were failing like directory present already
So removed directory too in this PR

santhosh93:
@PraveenPenguin : removed in all other tests as well

PR #368, opened 2017-03-27
Added perf test and providing fix to install perf packages.

shriyak:
Perf test is the built in tool which executes assorted sanity tests, initially through linked routines
but also will look for a directory with more tests in the form of scripts.
Fix will install perf using linux-tools-common and linux-tools for Ubuntu distro.

### DIFF
--- a/perf/perftool.py
+++ b/perf/perftool.py
@@ -15,12 +15,11 @@
 # Author:Praveen K Pandey <praveen@linux.vnet.ibm.com>
 #
 
-
 import os
-
+import platform
 from avocado import Test
 from avocado import main
-from avocado.utils import archive, build
+from avocado.utils import archive, build, distro, process
 from avocado.utils.software_manager import SoftwareManager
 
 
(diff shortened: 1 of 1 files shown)


perf/perftool.py
```diff
@@ -51,7 +60,19 @@ def setUp(self):
...
+        '''
+        perf test :Does sanity tests
+        Execute the tests by calling each module
+        '''
         self.count = 0
+        # Built in perf test
+        for string in process.run("perf test").stderr.splitlines():
+            if 'FAILED' in str(string.splitlines()):
+                self.count += 1
+                self.log.info("Test case failed is %s"
```
PraveenPenguin: i think you forgot to address previous review  comment here , self.fail() is appropriate here

perf/perftool.py
```diff
@@ -39,9 +38,19 @@ def setUp(self):
 
         # Check for basic utilities
         smm = SoftwareManager()
-        for package in ['gcc', 'make', 'perf']:
+        detected_distro = distro.detect()
+        kernel_ver = platform.uname()[2]
+        deps = ['gcc', 'make']
+        if 'Ubuntu' in detected_distro.name:
+            deps.extend(['linux-tools-common', 'linux-tools-%s' % kernel_ver])
+        elif detected_distro.name in ['redhat', 'SuSE', 'fedora']:
```
PraveenPenguin: i think this will work on centos as well can we have check for that as well or it can like 

it can be be only 
```
if ubuntu:

else :
```

perf/perftool.py
shriyak: I tried self.fail , but it will fail the test case at that point and execution was not able to proceed, so I included the count, which will display at the end

perf/perftool.py
shriyak: sure , let me include it.

PraveenPenguin:
@shriyak any update here ?

PraveenPenguin:
closing this PR as new version of this test raised as #448

PR #381, opened 2017-04-01
Fixed package install check in lvsetup

narasimhan-v:
pkg variable was initialized to " " instead of "".
Fixed that, and added that as a check in package installation.

### DIFF
--- a/io/disk/lvsetup.py
+++ b/io/disk/lvsetup.py
@@ -46,7 +46,7 @@ class Lvsetup(Test):
         """
         Check existence of input PV,VG, LV and snapshots prior to Test.
         """
-        pkg = " "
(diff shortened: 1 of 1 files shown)


PraveenPenguin:
@narasimhan-v can you please help me why this is required as it seems me test works perfect without this patch ?

narasimhan-v:
@PraveenPenguin `pkg` variable was initialized to " ", which is a wrong approach. In case of filesystems other than xfs/btrfs, it will try to install " ", and will fail.
I am not sure how/why it worked so long

PraveenPenguin:
@narasimhan-v oh ok sure did not realize that , thanks a lot for the patch

PR #386, opened 2017-04-04
sensors testcase fixups

sacsant:
Fix sensors test case to run only on bare-metal and a minor pep8 compliance fixup.

0fbc116: Add platform check for sensors testcase.
b21c214: Fix pep8 warning for sensors

### DIFF
--- a/cpu/sensors.py
+++ b/cpu/sensors.py
@@ -102,7 +102,8 @@ class Sensors(Test):
             except process.CmdError:
                 self.error(
                     'Starting Service Failed. Make sure module is loaded')
-        det_op = process.run('yes | sudo sensors-detect', shell=True, ignore_status=True).stdout
+        cmd = "yes | sudo sensors-detect"
+        det_op = process.run(cmd, shell=True, ignore_status=True).stdout
         if 'no sensors were detected' in det_op:
             self.skip('No sensors found to test !')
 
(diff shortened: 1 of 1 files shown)


cpu/sensors.py
```diff
@@ -51,7 +51,12 @@ def check_errors(cmd):
     def setUp(self):
         """
         Check pre-requisites before running sensors command
+        Testcase should be executed only on bare-metal environment.
         """
+        cmd = "cat /proc/cpuinfo | grep platform | awk -F':' '{print $NF}'\
+               | awk '{print $NF}'"
+        if not process.system_output(cmd, shell=True) == "PowerNV":
```
PraveenPenguin: please use this way` 'PowerNV' in open('/proc/cpuinfo', 'r').read() ` as invoking process.system() kind of performance over head , 

if not 'PowerNV' in open('/proc/cpuinfo', 'r').read()

PraveenPenguin: if you want to use avocado api than cpu.py can be used 

can be like 
```
from avocado.utils import cpu
-------------------------------------
-------------------------------------
if not 'PowerNV' in cpu._get_cpu_info():
```

sacsant:
Using avocado api is much simpler .. will resend the pull request with updated code

PR #389, opened 2017-04-04
Fixed path flush error in multupath_test

narasimhan-v:
When more than one multipath device was present, needed some
time after services are stable for flushing the paths. So, added
a sleep.


PraveenPenguin:
@narasimhan-v thanks for the patch

PR #391, opened 2017-04-05
Sensors package fix for suse

harish-24:
Fixed sensors package name for Suse and skip for other distros

### DIFF
--- a/cpu/sensors.py
+++ b/cpu/sensors.py
@@ -62,6 +62,10 @@ class Sensors(Test):
             if not s_mg.check_installed("lm-sensors") and not s_mg.install(
                     "lm-sensors"):
                 self.error('Need sensors to run the test')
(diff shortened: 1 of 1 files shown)


cpu/sensors.py
```diff
@@ -61,28 +61,32 @@ def setUp(self):
         if d_distro.name == "Ubuntu":
             if not s_mg.check_installed("lm-sensors") and not s_mg.install(
                     "lm-sensors"):
-                self.error('Need sensors to run the test')
+                self.skip('Need sensors to run the test')
+        elif d_distro.name == "SuSE":
```
PraveenPenguin: i think suse is not supported in BML so i don't think it is required

harish-24: Right. Sent PR #393 addressing this. Closing.

PR #397, opened 2017-04-05
Sensors fix1

PraveenPenguin:
1-remove not needed check
 a- remove ppc64 as distro check BML is always ppc64le
 b- remove powerkvm as distro check
2-Used avocado utility for config check

Signed-off-by: Praveen K Pandey praveen@linux.vnet.ibm.com

### DIFF
--- a/cpu/sensors.py
+++ b/cpu/sensors.py
@@ -18,11 +18,9 @@
 """
 Test for sensors command
 """
-import platform
-
 from avocado import Test
 from avocado import main
-from avocado.utils import process
+from avocado.utils import process, linux_modules
 from avocado.utils import distro
 from avocado.utils.software_manager import SoftwareManager
 from avocado.utils import cpu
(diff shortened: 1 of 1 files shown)


PraveenPenguin:
@harish-24 can you please have a look in to this

cpu/sensors.py
```diff
@@ -68,36 +67,23 @@ def setUp(self):
...
-            if 'platform\t: PowerNV\n' not in cpu._get_cpu_info():
-                self.skip('sensors test is applicable to bare-metal environment.')
-            kernel_ver = platform.uname()[2]
-            l_config = "CONFIG_SENSORS_IBMPOWERNV"
-            config_op = process.system_output(
-                'cat /boot/config-' + kernel_ver +
-                '| grep -i --color=never ' + l_config, shell=True)
-            if "=" not in config_op:
+
+        if 'platform\t: PowerNV\n' not in cpu._get_cpu_info():
```
harish-24: The PowerNV check has to be only for the arch `"ppc64", "ppc64le"`. Hence this would always skip on x86.
It should be
```
        if d_distro.arch in ["ppc64", "ppc64le"]:
            if 'platform\t: PowerNV\n' not in cpu._get_cpu_info():
                self.skip('sensors test is applicable to bare-metal environment.')
            else:
                config_check = linux_modules.check_kernel_config('CONFIG_SENSORS_IBMPOWERNV')
                if config_check == 0:
                ...
                ...
                else:
                    self.log.info('Module is Built In')
```
as all of this is specific to power architecture

cpu/sensors.py
```diff
@@ -68,36 +67,23 @@ def setUp(self):
...
-                if not c_val == "y":
-                    self.error('Config is not set properly')
-                else:
-                    self.log.info("Driver will be part of distro")
+            elif config_check == 1:
+                if linux_modules.load_module('ibmpowernv'):
+                    if linux_modules.module_is_loaded('ibmpowernv'):
+                        self.log.info('Module Loaded Successfully')
+                    else:
+                        self.error('Module Loading Failed')
```
harish-24: here as well

cpu/sensors.py
```diff
@@ -68,36 +67,23 @@ def setUp(self):
...
-                '| grep -i --color=never ' + l_config, shell=True)
-            if "=" not in config_op:
+
+        if 'platform\t: PowerNV\n' not in cpu._get_cpu_info():
+            self.skip('sensors test is applicable to bare-metal environment.')
+        else:
+            config_check = linux_modules.check_kernel_config(
+                'CONFIG_SENSORS_IBMPOWERNV')
+            if config_check == 0:
                 self.error('Config is not set')
```
harish-24: Can you change to self.skip() ?

cpu/sensors.py
PraveenPenguin: PowerNV is always ppc64le will check for ppc64le

cpu/sensors.py
PraveenPenguin: sure will do that

cpu/sensors.py
PraveenPenguin: sure

cpu/sensors.py
PraveenPenguin: my bad luck as for BE it script will not work will check for both

PraveenPenguin:
@harish-24 now pushed the requested change can please have a look ?

harish-24:
@PraveenPenguin Looks good.

PraveenPenguin:
@harish-24 thanks

apahim:
I see an ack, so merging straight away.

PR #399, opened 2017-04-06
Patch to fix ethtool test

hathyaga:
Fix will bring up interface after test completion

### DIFF
--- a/io/net/ethtool/ethtool_tests.sh
+++ b/io/net/ethtool/ethtool_tests.sh
@@ -64,3 +64,5 @@ for command in ifdown ifup; do
         exit $error
     fi
 done
+ifup $interface
+sleep 5


PraveenPenguin:
@hathyaga one question 5 seconds  is enough for interface bring up ? as sleep not looks good in tests

hathyaga:
@PraveenPenguin Usually a driver takes not more than 5 seconds to come up

hathyaga:
@PraveenPenguin  Is that ok?

PraveenPenguin:
@hathyaga sure my point is can we check interface state having timeout ?

io/net/ethtool/ethtool_tests.sh
```diff
@@ -64,3 +64,5 @@ for command in ifdown ifup; do
         exit $error
     fi
 done
+ifup $interface
+sleep 5
```
narasimhan-v: Something like this would do
```
for i in {1..10};
do
    ip link show ${interface} | grep "state UP"
    if [[ $? == 0 ]]; then
        break;
    fi
done
```

hathyaga:
Closing this PR as new PR https://github.com/avocado-framework-tests/avocado-misc-tests/pull/409 is merged

PR #400, opened 2017-04-06
Patch to fix net_data.py script

hathyaga:
Fix is for testinterface function to increase sleep time

### DIFF
--- a/io/net/net_data.py
+++ b/io/net/net_data.py
@@ -158,7 +158,7 @@ class NetDataTest(Test):
             self.fail("interface test failed")
         # up the interface
         process.system(if_up, shell=True)
-        time.sleep(4)
+        time.sleep(5)
         # check the status of interface through ethtool
         ret = process.system_output(self.eth, shell=True)
         if 'no' in ret:


hathyaga:
Failed logs
[net_data_fail.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/901479/net_data_fail.txt)

hathyaga:
Pass logs

hathyaga:
Pass logs
[net_data_pass.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/901483/net_data_pass.txt)

io/net/net_data.py
```diff
@@ -158,7 +158,7 @@ def testinterface(self):
             self.fail("interface test failed")
         # up the interface
         process.system(if_up, shell=True)
-        time.sleep(4)
+        time.sleep(5)
```
narasimhan-v: @hathyaga Are we sure this sleep is the best approach here ? Can't we check the interface state here with a timeout of 5 seconds ?

hathyaga: The operation performed here is ifup of the interface and then a sleep is done for the change to be reflected. The interface state is checked after that and then determines PASS/FAIl.

it only fails because the sleep time is less and the state (even though it is up) is collected as fail

narasimhan-v: Something like this should work
```
for i in range (0, 600, 60):
    if 'yes' in process.system_output(self.eth, shell=True):
        break
    time.sleep(60)
```

hathyaga:
New PR: https://github.com/avocado-framework-tests/avocado-misc-tests/pull/403

Closing this

PR #401, opened 2017-04-06
Added ignore_status in user creation.

santhosh93:
The test failed already if a group and username trinity already exists.
So added ignore_status in group and user creation step.

### DIFF
--- a/fuzz/trinity.py
+++ b/fuzz/trinity.py
@@ -40,9 +40,10 @@ class Trinity(Test):
         """
         Add not root user
         """
-        process.run('groupadd trinity', sudo=True)
+        process.run('groupadd trinity', sudo=True, ignore_status=True)
         process.run(
-            'useradd -g trinity  -m -d /home/trinity  trinity', sudo=True)
+            'useradd -g trinity  -m -d /home/trinity  trinity',
+            ignore_status=True, sudo=True)
         process.run('usermod -a -G trinity  trinity', sudo=True)
 
         smm = SoftwareManager()


fuzz/trinity.py
```diff
@@ -40,9 +40,10 @@ def setUp(self):
         """
         Add not root user
         """
-        process.run('groupadd trinity', sudo=True)
+        process.run('groupadd trinity', sudo=True, ignore_status=True)
         process.run(
-            'useradd -g trinity  -m -d /home/trinity  trinity', sudo=True)
+            'useradd -g trinity  -m -d /home/trinity  trinity',
+            ignore_status=True, sudo=True)
```
narasimhan-v: So, would you continue even if it fails ?
Also, is it not needed to set `ignore_status=True` to `usermod` in the next line ?

PraveenPenguin:
@santhosh93 i think better approach is check if user is already there skip the user creation step ,

check can be :

```
if process.system('getent passwd trinity',ignore_status=True):
    process.run(useradd -g trinity  -m -d /home/trinity  trinity',sudo=True)

```

santhosh93:
@narasimhan-v @PraveenPenguin : may be we can check whether both group/user if already exists and we can create a group / user.
@narasimhan-v : in usermod we are adding that user to that group..in the above statements if we are sure that both user and group exists ..then usermod should not give a prob.

santhosh93:
Opened a new PR - https://github.com/avocado-framework-tests/avocado-misc-tests/pull/410

PR #404, opened 2017-04-06
Skip test if disk space is insufficient

harish-24:
Test skips if disk space is less than total memory of the system

### DIFF
--- a/generic/interbench.py
+++ b/generic/interbench.py
@@ -24,7 +24,7 @@ from avocado import Test
 from avocado import main
 from avocado.utils import archive
 from avocado.utils import process
-from avocado.utils import build
+from avocado.utils import build, disk, memory
 from avocado.utils.software_manager import SoftwareManager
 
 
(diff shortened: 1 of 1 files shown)


generic/interbench.py
```diff
@@ -24,7 +24,8 @@
 from avocado import main
 from avocado.utils import archive
 from avocado.utils import process
-from avocado.utils import build
+from avocado.utils import build, disk, memory
+from avocado.core import data_dir as datadir
```
narasimhan-v: Are we allowed to use core modules in tests ?

narasimhan-v:
Job log please

generic/interbench.py
harish-24: It is being used in other tests

harish-24:
[inter_log.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/902623/inter_log.txt)

PraveenPenguin:
@harish-24 Please don't use internal APIs. They have no guarantee of stability and are intended for internal usage of avocado , can `self.teststmpdir` not be used here , let me check in  repo  where we used internal API may be we need to fix those places as i pointed there no  guarantee of stability

harish-24:
@PraveenPenguin Ok. `self.teststmpdir` can be used in this case. Changed accordingly.

PraveenPenguin:

PR #403, opened 2017-04-06
patch for net-data script

hathyaga:
Patch fixes testinterface execution for interface check based
on timeout

Signed-off-by: Harsha Thyagaraja <harshkid@linux.vnet.ibm.com>

PR #406, opened 2017-04-06
Add SMT on/off loop subtest to ppc64_cpu_test

sacsant:
This patch adds a sub test to turn off/on SMT capabilities in a loop
with count of 100.
Additionally the test collects dmesg output at the end of the test.

### DIFF
--- a/cpu/ppc64_cpu_test.py
+++ b/cpu/ppc64_cpu_test.py
@@ -91,10 +91,15 @@ class PPC64Test(Test):
             self.threads_per_core()
             self.smt_snoozedelay()
             self.dscr()
+
+        self.smt_loop()
+
         if self.failures > 0:
             self.log.debug("Number of failures is %s" % self.failures)
             self.log.debug(self.failure_message)
             self.fail()
+
+        process.system_output("dmesg")
 
     def smt(self):
         """
@@ -146,6 +151,14 @@ class PPC64Test(Test):
         command1 = "ppc64_cpu --dscr | awk '{print $NF}'"
         command2 = "cat /sys/devices/system/cpu/dscr_default"
         self.equality_check("DSCR", command1, command2)
+
+    def smt_loop(self):
+        """
+        Tests smt on/off in a loop
+        """
+        for _ in range(1, 100):
+            process.run("ppc64_cpu --smt=off && ppc64_cpu --smt=on",
+                        shell=True)
 
     def tearDown(self):
         """


cpu/ppc64_cpu_test.py
```diff
@@ -91,11 +91,16 @@ def test(self):
...
             self.dscr()
+
+        self.smt_loop()
+
         if self.failures > 0:
             self.log.debug("Number of failures is %s" % self.failures)
             self.log.debug(self.failure_message)
             self.fail()
 
+        process.system_output("dmesg")
```
narasimhan-v: Is this needed, since avocado profilers are getting them anyway ?

cpu/ppc64_cpu_test.py
```diff
@@ -147,6 +152,14 @@ def dscr(self):
         command2 = "cat /sys/devices/system/cpu/dscr_default"
         self.equality_check("DSCR", command1, command2)
 
+    def smt_loop(self):
+        """
+        Tests smt on/off in a loop
+        """
+        for i in range(1, 100):
```
narasimhan-v: `for _ in range(1, 100):`

cpu/ppc64_cpu_test.py
```diff
@@ -147,6 +152,14 @@ def dscr(self):
...
+            process.system_output("ppc64_cpu --smt=off && ppc64_cpu --smt=on",
```
narasimhan-v: Any reason as to why not running them in 2 separate `process.system_output`

narasimhan-v: Also, process.run would be better, since it gives the return status as well

narasimhan-v:
@sacsant job log please

cpu/ppc64_cpu_test.py
sacsant: smt off operation takes longer time. The time depends on number of cpus. Running them together will ensure that smt on will not be executed before smt off is completed. This will avoid any timing errors

sacsant: yes, will change.

cpu/ppc64_cpu_test.py
sacsant: I wanted to collect the dmesg log so that if that I can capture any errors due to this test. I infact had added a dmesg -c in the setup but then decided against it as it might affect logs when this test case is run with other avocado tests in a batch.

cpu/ppc64_cpu_test.py
sacsant: done

sacsant:
[job.log.zip](https://github.com/avocado-framework-tests/avocado-misc-tests/files/902646/job.log.zip)

cpu/ppc64_cpu_test.py
```diff
@@ -147,6 +152,14 @@ def dscr(self):
         command2 = "cat /sys/devices/system/cpu/dscr_default"
         self.equality_check("DSCR", command1, command2)
 
+    def smt_loop(self):
+        """
+        Tests smt on/off in a loop
+        """
+        for _ in range(1, 100):
+            process.run("ppc64_cpu --smt=off && ppc64_cpu --smt=on",
+                        shell=True)
```
PraveenPenguin: is there any failure condition can be catch to fail the test just a thought

PraveenPenguin: if it is ok to test catch failure one way to check return status 

can be like :
```
for _ in range(1, 100):
    if process.system("ppc64_cpu --smt=off && ppc64_cpu --smt=on", shell=True, ignore_status=True):
         self.fail("msg")
```

sacsant: SMT on/off command probably will not fail as far as the tool is concerned. If there are any failures it probably will be kernel failures. That was one of the reason I added the dmesg log collection after the loop test so that any kernel side failures/trace messages can be captured.

cpu/ppc64_cpu_test.py
narasimhan-v: I would recommend writing a to change smt state and see if smt changes state by giving a timeout, and breaking once it changes, and report a failure if it does not. And call this function a 100 times.

sacsant: On a LPAR I have seen the SMT off operation takes roughly a minute .. while on a bare metal box it might take up 2 1/2 minutes. I am sure this value will vary depending on machine and resources.  I would like to avoid adding such a code.

narasimhan-v: @sacsant can we follow an approach similar to https://github.com/avocado-framework-tests/avocado-misc-tests/pull/403/files

cpu/ppc64_cpu_test.py
PraveenPenguin: sure make sense

PraveenPenguin:

PR #410, opened 2017-04-07
Check for trinity group/user before creation

santhosh93:
Used getent command to check whether the user and group
'trinity' exists before creating it.


PraveenPenguin:

PR #415, opened 2017-04-10
Changed the default mountpoint to teststmpdir

narasimhan-v:
Some tests were using self.srcdir for multiple purposes.
Changed the tests which mount a disk as well as download a
tarball, to use default mountpoint as self.teststmpdir.


PraveenPenguin:
@nasastry thanks

PR #421, opened 2017-04-12
Linux scheduler testsuite

PraveenPenguin:
added linux-scheduler-testing as automative way test linux-scheduler


PraveenPenguin:
job log : [job.log.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/916260/job.log.txt)

cpu/linsched.py
```diff
@@ -0,0 +1,70 @@
...
+            os.path.join(self.datadir, 'fix.patch'))
+        process.system(fix_patch, shell=True, ignore_status=True)
+
+        build.make(self.srcdir)
+
+    def test(self):
+
+        os.chdir(self.srcdir)
+
+        if process.system('./%s' % self.args, ignore_status=True, shell=True):
```
narasimhan-v: I think `process.run()` will be better here, since we need failure output as well.

cpu/linsched.py
```diff
@@ -0,0 +1,70 @@
...
+        for package in deps:
+            if not smm.check_installed(package) and not smm.install(package):
+                self.cancel(
+                    "Fail to install %s required for this test." % package)
+        git.get_repo('https://github.com/thejinxters/linux-scheduler-testing',
+                     destination_dir=self.srcdir)
+        os.chdir(self.srcdir)
+        fix_patch = 'patch -p1 < %s' % (
+            os.path.join(self.datadir, 'fix.patch'))
+        process.system(fix_patch, shell=True, ignore_status=True)
```
narasimhan-v: And I suggest `process.run()` here as well

cpu/linsched.py
PraveenPenguin: @narasimhan-v i think should be fine as it is scheduler test so not sure here we can get any failure condition as after test run if command executed successfully user can verify out put of time take by scheduler

cpu/linsched.py
PraveenPenguin: this can be done but not thinking about benefit of using `process.run` instead of `process.system` are you pointing we can check out put of patch applied ?

narasimhan-v: Yes, we need output of failure case

cpu/linsched.py
```diff
@@ -0,0 +1,70 @@
...
+        '''
+
+        # Check for basic utilities
+        smm = SoftwareManager()
+        deps = ['gcc', 'make', 'patch']
+        if distro.detect().name == "SuSE":
+            deps.append('git-core')
+        else:
+            deps.append('git')
+        self.args = self.params.get('args', default='pi 100')
```
apahim: this line should be after the `self.cancel()`

cpu/linsched.py
```diff
@@ -0,0 +1,70 @@
...
+        process.run(fix_patch, shell=True)
+
+        build.make(self.srcdir)
+
+    def test(self):
+
+        os.chdir(self.srcdir)
+
+        if process.system('./%s' % self.args, ignore_status=True, shell=True):
+            self.fail("test run fails of  %s" % self.args)
```
apahim: I'd suggest to rephrase this to something like:
```python
        cmd = './%s' % self.args
            self.fail('Command [%s] failed.' % cmd)
```

apahim: or just...
```python
        if process.system('./%s' % self.args, ignore_status=True, shell=True):
            self.fail('Test [%s] failed.' % self.args)
```

cpu/linsched.py
PraveenPenguin: sure will do

cpu/linsched.py
PraveenPenguin: make sense will change

PR #427, opened 2017-04-14
Fixed exfiotest output dir parameter

narasimhan-v:
Output dir was having ';'.
-o option was taking the directory without "", and thus was
parsing them as 2 commands.
Fixed it.


PraveenPenguin:
@narasimhan-v thanks

PR #429, opened 2017-04-14
Added test for sosreport - v2

Pavithra1602:
sosreport generates an archive of configuration and diagnostic information from the running system.

### DIFF
--- a/generic/sosreport.py
+++ b/generic/sosreport.py
@@ -1,0 +1,149 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2017 IBM
+# Author: Pavithra <pavrampu@linux.vnet.ibm.com>
+
+import os
+import tempfile
+import shutil
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+from avocado.utils import distro
+from avocado.utils.software_manager import SoftwareManager
+
+
+class sosreport_test(Test):
+
+    is_fail = 0
+
+    def run_cmd(self, cmd):
+        self.log.info("executing ============== %s =================" % cmd)
+        if process.system(cmd, ignore_status=True, sudo=True):
+            self.log.info("%s command failed" % cmd)
+            self.is_fail += 1
+        return
+
+    def run_cmd_out(self, cmd):
+        return process.system_output(cmd, shell=True, ignore_status=True, sudo=True)
+
+    def setUp(self):
+        dist = distro.detect()
+        sm = SoftwareManager()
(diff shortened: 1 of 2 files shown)


generic/sosreport.py
```diff
@@ -0,0 +1,149 @@
...
+            self.log.info("sosreport is installed")
+        elif dist.name == 'Ubuntu':
+            if not sm.install("sosreport"):
+                self.skip("Fail to install sosreport required for this test.")
+        elif dist.name == 'redhat':
+            if not sm.install("sos"):
+                self.skip("Fail to install sos package required for this test.")
+        else:
+            self.skip("sosreport is not supported on this distro "
+                      "or package is not installed")
```
PraveenPenguin: why not it can be done in this way 
```

sos_pkg=""
if dist.name == 'Ubuntu':
  sos_pkg='sosreport'
elif dist.name == 'redhat':
  sos_pkg='sos'
else:
  self.skip("sosreport is not supported on this distro ")

if sos_pkg and not smm.check_installed(sos_pkg) and not smm.install(sos_pkg):
   self.skip("Package %s is missing and could not be installed" % pkg)
```

generic/sosreport.py
```diff
@@ -0,0 +1,149 @@
...
+        if not os.path.isdir(dir_name):
+            self.is_fail += 1
+            self.log.info("--build option failed")
+        self.run_cmd("sosreport --batch --tmp-dir=%s -v" % directory_name)
+        self.run_cmd("sosreport --batch --tmp-dir=%s --verify" % directory_name)
+        if self.run_cmd_out("sosreport --batch --tmp-dir=%s --quiet -e ntp,numa1" % directory_name):
+            self.is_fail += 1
+            self.log.info("--quiet option failed")
+        self.run_cmd("sosreport --batch --tmp-dir=%s --debug" % directory_name)
+        if 'test123' not in self.run_cmd_out("sosreport --batch --ticket-number=test123  | "
```
PraveenPenguin: one help can test123 also be a yaml param as i think ticket id and case id can change what you think ?

generic/sosreport.py
```diff
@@ -0,0 +1,149 @@
...
+                            "grep libraries" % directory_name):
+            self.is_fail += 1
+            self.log.info("--skip-plugins option failed")
+        self.run_cmd("sosreport --batch --tmp-dir=%s -e ntp,numa,snmp" % directory_name)
+        if 'sendmail' not in self.run_cmd_out("sosreport --batch --tmp-dir=%s -e "
+                                              "sendmail | grep sendmail" % directory_name):
+            self.is_fail += 1
+            self.log.info("--enable-plugins option failed")
+        self.run_cmd("sosreport --batch --tmp-dir=%s -o "
+                     "pci,powerpc,procenv,process,processor,kdump" % directory_name)
```
PraveenPenguin: this command only run in power pc if yes please have check of pate form before run this test as in other arch it kind of misleading

generic/sosreport.py
Naresh-ibm: line31-line35
     def run_cmd(self, cmd):
        self.log.info("executing ============== %s =================" % cmd)
        if process.system(cmd, ignore_status=True, sudo=True):
            self.log.info("%s command failed" % cmd)
            self.is_fail += 1


according to avocado usage this can be
self.log.info("executing ============== %s =================", cmd)
self.log.info("%s command failed", cmd)

Pavithra1602:
Raised new PR.

PR #431, opened 2017-04-14
Added test for RAS tools -v2

Pavithra1602:
This test tests various options of RAS tools.

### DIFF
--- a/generic/ras_extended.py
+++ b/generic/ras_extended.py
@@ -1,0 +1,237 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2017 IBM
+# Author: Pavithra <pavrampu@linux.vnet.ibm.com>
+
+import os
+from shutil import copyfile
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+from avocado import skipIf
+from avocado.utils.software_manager import SoftwareManager
+
+IS_POWER_NV = 'PowerNV' in open('/proc/cpuinfo', 'r').read()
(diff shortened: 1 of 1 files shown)


generic/ras_extended.py
```diff
@@ -0,0 +1,237 @@
...
+    This test checks various RAS tools:
+    """
+    is_fail = 0
+
+    def run_cmd(self, cmd):
+        cmd_result = process.run(cmd, ignore_status=True, sudo=True,
+                                 shell=True)
+        if cmd_result.exit_status != 0:
+            self.is_fail += 1
+        return
```
PraveenPenguin: it can be like 

```
if process.system(cmd, ignore_status=True, sudo=True, shell=True):
   self.is_fail += 1
 return 
```

generic/ras_extended.py
```diff
@@ -0,0 +1,237 @@
...
+
+    def setUp(self):
+        architecture = os.uname()[4]
+        if "ppc" not in architecture:
```
PraveenPenguin: i think it can be like 
`if "ppc" not in os.uname()[4]:` want to eliminate extra variable

generic/ras_extended.py
```diff
@@ -0,0 +1,237 @@
...
+            self.skip("supported only on Power platform")
+        sm = SoftwareManager()
+        for package in ("ppc64-diag", "powerpc-utils", "lsvpd", "sysfsutils"):
+            if not sm.check_installed(package) and not sm.install(package):
+                self.error("Fail to install %s required for this"
```
PraveenPenguin: please use `self.skip` or `self.cancel`  instead of `self.error `

generic/ras_extended.py
```diff
@@ -0,0 +1,237 @@
...
+        if self.is_fail >= 1:
+            self.fail("%s command(s) failed in ls-vdev tool "
+                      "verification" % self.is_fail)
+
+    @skipIf(IS_POWER_NV or IS_KVM_GUEST, "This test is not supported on KVM guest or PowerNV platform")
+    def test1_lsdevinfo(self):
+        self.log.info("===============Executing lsdevinfo tool test===="
+                      "===========")
+        self.run_cmd("lsdevinfo")
+        list = ['-h', '-V', '-c', '-R', '-F name,type']
```
PraveenPenguin: just a thought can we pass list  as  yaml param ?

Pavithra1602:
Raised new PR

PR #436, opened 2017-04-18
Added check for VCPU,also included failure info for all failures,opti…

shriyak:
…mized the code

and fixed all pep8 issues.

24*7 performance  counters is supported on Power VM platform.
Script will cover basic testing of events in 24x7 counters that belong to different
domains like core,chip,virtual CPU home core, etc.To monitor an event, the user must
specify the required parameters for the event,which include its name,
the chip/core on which to monitor the event as well as the domain.


PraveenPenguin:
please rebase as remove `Unused import sys` it make travis happy :)

PR #441, opened 2017-04-19
Added the multipath package and service name for Sles OS.

Naresh-ibm:
The multipath package and service names are different for rhel
and sles. hence script was skipping while running. so added
the same to the script and now it is working fine.


PraveenPenguin:
@narasimhan-v thanks

PR #448, opened 2017-04-20
Added perf test and provided fix to install packages for Ubuntu , hen…

shriyak:
…ce included the check for distros.

Script executes perf test and perftool testsuite.
Perf test : does sanity tests and execute the tests by calling each module.
Perftool testsuite : This testsuite will cover all the perf's built in comands, their options and
use-cases in various situations.It should be run against the perftool installed from the packages or just built

### DIFF
--- a/perf/perftool.py
+++ b/perf/perftool.py
@@ -13,14 +13,12 @@
 #
 # Copyright: 2016 IBM
 # Author:Praveen K Pandey <praveen@linux.vnet.ibm.com>
-#
-
 
 import os
-
+import platform
 from avocado import Test
 from avocado import main
-from avocado.utils import archive, build
+from avocado.utils import archive, build, distro, process
 from avocado.utils.software_manager import SoftwareManager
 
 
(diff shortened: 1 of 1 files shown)


perf/perftool.py
```diff
@@ -32,16 +30,24 @@ class Perftool(Test):
...
         '''
 
         # Check for basic utilities
         smm = SoftwareManager()
-        for package in ['gcc', 'make', 'perf']:
+        detected_distro = distro.detect()
+        kernel_ver = platform.uname()[2]
+        deps = ['gcc', 'make']
+        if 'Ubuntu' in detected_distro.name:
+            deps.extend(['linux-tools-common', 'linux-tools-%s' % kernel_ver])
```
PraveenPenguin: one help for ubuntu also i think perf package is required ?

shriyak: For ubuntu, perf is installed using linux-tools-common, perf package is not needed.

PraveenPenguin: oh ok sure , thanks for clarification

PR #457, opened 2017-04-25
Use root space by default to create loop device

harish-24:
Test now runs with loopdevice creation under root if base disk is not provided.

### DIFF
--- a/fs/xfstests.py
+++ b/fs/xfstests.py
@@ -28,7 +28,7 @@ import shutil
 
 from avocado import Test
 from avocado import main
-from avocado.utils import process, build, git, distro, partition
+from avocado.utils import process, build, git, distro, partition, disk
 from avocado.utils.software_manager import SoftwareManager
 
 
@@ -73,6 +73,9 @@ class Xfstests(Test):
         self.disk_mnt = self.params.get('disk_mnt', default='/mnt/loop_device')
         self.dev_type = self.params.get('type', default='loop')
         self.fs_to_test = self.params.get('fs', default='ext4')
+        if process.system('which mkfs.%s' % self.fs_to_test, ignore_status=True):
+            self.skip('Unknown filesystem %s' % self.fs_to_test)
+        mount = True
         self.devices = []
         shutil.copyfile(os.path.join(self.datadir, 'local.config'),
                         os.path.join(self.srcdir, 'local.config'))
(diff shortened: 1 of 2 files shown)


fs/xfstests.py
```diff
@@ -83,8 +84,13 @@ def setUp(self):
             base_disk = self.params.get('disk', default=None)
             loop_size = self.params.get('loop_size', default='9GiB')
             if not base_disk:
-                self.skip('Please provide a base disk to create loop device')
-            self._create_loop_device(base_disk, loop_size)
+                # Using root for file creation by default
+                if disk.freespace('/')/1024/1024/1024 > 15:
```
narasimhan-v: `if disk.freespace('/')/1073741824`

fs/xfstests.py
```diff
@@ -161,14 +167,16 @@ def tearDown(self):
...
                                sudo=True, ignore_status=True)
-            self.part.unmount()
-
-    def _create_loop_device(self, base_disk, loop_size):
-        self.part = partition.Partition(
-            base_disk, mountpoint=self.disk_mnt)
-        self.part.mount()
-        if process.system('which mkfs.%s' % self.fs_to_test, ignore_status=True):
-            self.skip('Unknown filesystem %s' % self.fs_to_test)
+            if not self.disk_mnt == '':
```
narasimhan-v: Will `if not self.disk_mnt:` not work here ?

fs/xfstests.py
```diff
@@ -161,14 +167,16 @@ def tearDown(self):
...
+                self.part.unmount()
+
+    def _create_loop_device(self, base_disk, loop_size, mount=True):
+        if mount:
+            self.part = partition.Partition(
+                base_disk, mountpoint=self.disk_mnt)
+            self.part.mount()
+            if process.system('which mkfs.%s' % self.fs_to_test, ignore_status=True):
```
narasimhan-v: Suggestion:
Something like this should also work, your take
```
from distutils.spawn import find_executable
if find_executable("mkfs.%s" % self.fs_to_test):
```

fs/xfstests.py.data/xfstests.yaml
```diff
@@ -9,8 +8,15 @@ setup:
     # Run with either loop_type (or) disk_type
     loop_type: !mux
         type: 'loop'
-        loop_size: '5GiB'
-        disk: /dev/sdj2
+        loop_size: '7GiB'
+        # Option to provide disk for loop device creation,
+        # Uses '/' by default for file creation
+        # disk: /dev/sdj2
```
narasimhan-v: `disk` can be added as a parameter here, without any value to it, so that it is easier to edit the value for automated runs later. Example below:
`disk: `

narasimhan-v:
@harish-24 Job log please

fs/xfstests.py
harish-24: If we can have a utility, we can change across tests later. So for now i am opting to old one.

narasimhan-v: Another suggestion: Can we have this check immediately after reading the parameter `self.fs_to_test` ?

harish-24:
@narasimhan-v Pushed with changes. Thanks.
Job log: [xfs_base_disk.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/957590/xfs_base_disk.txt)

PraveenPenguin:
@narasimhan-v thanks

PR #458, opened 2017-04-26
Usage of srcdir and teststmpdir in a better way

narasimhan-v:
teststmpdir is common for all variants of a test. So, it is the
ideal location to extract repos / tar-balls, for reuse by other
variants of the test.
So, changed the usage of some IO tests accordingly.

This is in line with the discussion at https://github.com/avocado-framework/avocado/issues/1924

### DIFF
--- a/io/disk/lvsetup.py
+++ b/io/disk/lvsetup.py
@@ -27,7 +27,6 @@ using a given policy.
 
 For details about the policy see README.
 """
-import os
 import avocado
 from avocado import Test
 from avocado import main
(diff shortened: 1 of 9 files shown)


io/disk/lvsetup.py
```diff
@@ -82,8 +81,7 @@ def setUp(self):
         self.lv_name = lv_name
         if lv_utils.lv_check(vg_name, lv_snapshot_name):
             self.skip('Snapshot %s already exists' % lv_snapshot_name)
-        self.mount_loc = 'avocado_%s' % self.fs_name
-        os.mkdir(self.mount_loc)
+        self.mount_loc = self.srcdir
```
PraveenPenguin: i think earlier one is ok what u say but no hard feeling ?

narasimhan-v: Thought it would be better to use `self.srcdir` as uniform across all tests for mounting.

PraveenPenguin: sure , thanks for help , looks me good :)

narasimhan-v: Thank you

PR #462, opened 2017-05-02
Test to verify servicelog - v5

Pavithra1602:
The servicelog packages provides a library for logging service-related
events to the servicelog database, and a number of command-line
utilities for viewing the contents of the database.

### DIFF
--- a/generic/servicelog.py
+++ b/generic/servicelog.py
@@ -1,0 +1,240 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


generic/servicelog.py
```diff
@@ -0,0 +1,240 @@
...
+from avocado import main
+from avocado.utils import process
+from avocado.utils.service import ServiceManager
+from avocado.utils.software_manager import SoftwareManager
+
+
+class servicelog(Test):
+
+    is_fail = 0
+    url = "http://ausgsa.ibm.com/projects/p/plx/FVT/rtas-events"
```
PraveenPenguin: i think this is internal link not able to access out side

Pavithra1602: could not find these files on any other links. need to keep in data directory.

PraveenPenguin: @Pavithra1602 if there is no other way than i think we can kept it on data dir , thanks for helping

PR #472, opened 2017-05-04
Enhanced PowerVMEEH script

vrbagalkote:
Now script will be able to inject one error more than the
eeh_max-freeze bit, so that we can test adapter failure.
Also, if there is EEH miss for more than 5 times for a perticular
function, it continues with next adapter throwing a warning

### DIFF
--- a/io/pci/PowerVMEEH.py
+++ b/io/pci/PowerVMEEH.py
@@ -60,7 +60,7 @@ class PowerVMEEH(Test):
         if output != '0x1':
             self.skip("EEH is not enabled, please enable via FSP")
             sys.exit(1)
-        self.max_freeze = int(self.params.get('max_Freeze', default='1'))
+        self.max_freeze = int(self.params.get('max_freeze', default='1'))
         cmd = "echo %d > /sys/kernel/debug/powerpc/eeh_max_freezes"\
             % self.max_freeze
         process.system(cmd, ignore_status=True, shell=True)
@@ -73,6 +73,9 @@ class PowerVMEEH(Test):
         Test to execute basic error injection on PE
         """
         for self.addr in pci.get_pci_addresses():
+            enter_loop = True
+            num_of_miss = 0
+            num_of_hit = 0
             self.pci_mem_addr = pci.get_memory_address(self.addr)
             self.pci_mask = pci.get_mask(self.addr)
             self.pci_class_name = pci.get_pci_class_name(self.addr)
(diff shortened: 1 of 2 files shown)


vrbagalkote:
[job.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/975961/job.txt)

i have included time.sleep(10), because command lspci -D from avocado.utils.pci.py fails if we execute lspci -D as soon as the adapter is removed. So this is like a work around.

PraveenPenguin:
@narasimhan-v can we tweak library it self if sleep is really needed for lspci command , please suggest

io/pci/PowerVMEEH.py
```diff
@@ -60,11 +60,13 @@ def setUp(self):
         if output != '0x1':
             self.skip("EEH is not enabled, please enable via FSP")
             sys.exit(1)
-        self.max_freeze = int(self.params.get('max_Freeze', default='1'))
+        self.max_freeze = int(self.params.get('max_freeze', default='1'))
```
PraveenPenguin: i think it should be` maxFreeze` as yaml entry show this , please rectify in any one of the place

io/pci/PowerVMEEH.py
```diff
@@ -83,21 +87,46 @@ def test_eeh_basic_pe(self):
...
+                                    self.log.info("PE recovered successfully")
+                    else:
+                        self.log.warning("EEH inject failed for 5 times with\
+                                   function %s" % func)
+                        self.enter_loop = False
                         break
-                    self.log.info("PE recovered successfully")
+                if not self.enter_loop:
+                    break
+            if self.num_of_hit > self.max_freeze:
```
narasimhan-v: `else:` instead of `if self.num_of_hit > self.max_freeze:`

io/pci/PowerVMEEH.py
```diff
@@ -73,6 +75,8 @@ def test_eeh_basic_pe(self):
         Test to execute basic error injection on PE
         """
         for self.addr in pci.get_pci_addresses():
+            self.num_of_miss = 0
+            self.num_of_hit = 0
```
narasimhan-v: This was already defined in `setUp()`. Is it needed here as well ?

io/pci/PowerVMEEH.py
```diff
@@ -83,21 +87,46 @@ def test_eeh_basic_pe(self):
...
-                else:
-                    if not self.check_eeh_pe_recovery(self.addr):
-                        self.fail("PE %s recovery failed after first EEH"
-                                  % self.addr)
+            while self.num_of_hit <= self.max_freeze:
+                for func in self.function:
+                    self.log.info("Running error inject on pe %s function %s"
+                                  % (self.addr, func))
+                    if self.num_of_miss < 5:
+                        self.return_code = self.basic_eeh(func,
```
PraveenPenguin: lot of global variable as i think it requires only in sub routine scope please have a look

io/pci/PowerVMEEH.py
narasimhan-v: yaml file has `maxFreeze`

io/pci/PowerVMEEH.py
```diff
@@ -83,21 +87,46 @@ def test_eeh_basic_pe(self):
...
-                    continue
-                else:
-                    if not self.check_eeh_pe_recovery(self.addr):
-                        self.fail("PE %s recovery failed after first EEH"
-                                  % self.addr)
+            while self.num_of_hit <= self.max_freeze:
+                for func in self.function:
+                    self.log.info("Running error inject on pe %s function %s"
+                                  % (self.addr, func))
+                    if self.num_of_miss < 5:
```
narasimhan-v: Why comparing against 5 here ?

io/pci/PowerVMEEH.py
```diff
@@ -83,21 +87,46 @@ def test_eeh_basic_pe(self):
...
+                                   function %s" % func)
+                        self.enter_loop = False
                         break
-                    self.log.info("PE recovered successfully")
+                if not self.enter_loop:
+                    break
+            if self.num_of_hit > self.max_freeze:
+                if self.check_eeh_removed():
+                    self.log.info("PE %s removed successfully" % self.addr)
+                    time.sleep(10)
```
narasimhan-v: I also doubt if this sleep is really needed.
If needed, is it working as designed ?

io/pci/PowerVMEEH.py
vrbagalkote: yes, because for each and every adapter, hit and miss to be set as zero before starting the test.

io/pci/PowerVMEEH.py
vrbagalkote: 5 is just a random number. I have seen many times where eeh_hit is missed for first three times, and then is successful after that, and also once its not hit even after 5 times, its safe to assume to try with next function. Even in the attached job.log file, EEH is miss for first three times, but then later its successful for first adapter.

io/pci/PowerVMEEH.py
vrbagalkote: Its working as designed. But I will share the job,log file without sleep(), please suggest. Even I dont want sleep function here.:)

vrbagalkote: Discussed with @narasimhan-v , and thought its better to put sleep() in eeh_removed(), its better suited there.

vrbagalkote:
@PraveenPenguin @narasimhan-v addressed the comments.

PraveenPenguin:
@narasimhan-v thanks

PR #477, opened 2017-05-09
SMT Changes in HTX now runs once every hour

narasimhan-v:
Enhanced HTX test to run SMT changes once every hour,
since running it very frequently is not a proper use case.

### DIFF
--- a/io/disk/htx_block_devices.py
+++ b/io/disk/htx_block_devices.py
@@ -35,7 +35,7 @@ class HtxTest(Test):
     :see:https://github.com/open-power/HTX.git
     :param block_devices: names of block_devices on which you want to run HTX
     :param mdt_file: mdt file used to trigger HTX
-    :params time_limit: how much time you want to run this stress.
+    :params time_limit: how much time(hours) you want to run this stress.
     """
 
     def setUp(self):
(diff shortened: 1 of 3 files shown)


io/disk/htx_block_devices.py
```diff
@@ -118,8 +118,10 @@ def test(self):
         self.log.info("Running the HTX on %s", self.block_device)
         cmd = "htxcmdline -run %s -mdt %s" % (self.block_device, self.mdt_file)
         process.system(cmd, ignore_status=True)
-        for _ in range(0, self.time_limit, 60):
-            self.run_smt()
+        for time_loop in range(0, self.time_limit, 60):
+            # Running SMT changes every hour
+            if time_loop % 3600 == 0:
```
PraveenPenguin: i am not sure after a hour smt change will really effect the test so i suggested can we get some approx time of SMT change may be 5 min should be ok , what you think

narasimhan-v: I had a discussion with developers, and it is not advisable to change SMT values so frequently when a workload ia running.
So, a sufficient time interval (1 hour in this case) is necessary.

PraveenPenguin: @narasimhan-v sure make sense , thanks for helping me here

PR #480, opened 2017-05-09
Enhanced avago9361.py script

vrbagalkote:
Added extensive set of adapter functinality, for which
values can be set=<ON|OFF>

### DIFF
--- a/io/disk/Avago_storage_adapter/avago9361.py
+++ b/io/disk/Avago_storage_adapter/avago9361.py
@@ -76,7 +76,16 @@ class Avago9361(Test):
         """
         adjust_on_off = ['restorehotspare', 'autorebuild', 'copyback', 'eghs',
                          'alarm', 'foreignautoimport', 'maintainpdfailhistory',
-                         'ocr']
+                         'ocr', 'immediateio', 'largeQD',
+                         'driveactivityled', 'flushwriteverify',
+                         'limitMaxRateSATA', 'supportssdpatrolread',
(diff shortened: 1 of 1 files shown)


vrbagalkote:
[job.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/986381/job.txt)
 Attached is job.log

io/disk/Avago_storage_adapter/avago9361.py
```diff
@@ -76,7 +76,16 @@ def test_set_on_off(self):
         """
         adjust_on_off = ['restorehotspare', 'autorebuild', 'copyback', 'eghs',
                          'alarm', 'foreignautoimport', 'maintainpdfailhistory',
-                         'ocr']
+                         'ocr', 'immediateio', 'largeQD',
+                         'driveactivityled', 'flushwriteverify',
+                         'limitMaxRateSATA', 'supportssdpatrolread',
```
PraveenPenguin: can these a yaml param as it help to user to on/off which he/she wants please suggest

vrbagalkote: These are basically MFC(megaraid feature controller) values and its better to test all these values anyway. So I have given in the script instead of in yaml.

PraveenPenguin: sure make sense yes default it run all the option question is like we can enable custom as well?

PraveenPenguin:
@vrbagalkote thanks

PR #483, opened 2017-05-11
Bonding script enhancement to choose creation of bonding  interface in Peer

hathyaga:
interface in peer

Also, bond_fail function is enhanced to consider various
failover conditions

### DIFF
--- a/io/net/bonding.py
+++ b/io/net/bonding.py
@@ -45,10 +45,7 @@ class Bonding(Test):
         depends = []
         if detected_distro.name == "Ubuntu":
             depends.append("openssh-client")
-        # FIXME: "redhat" as the distro name for RHEL is deprecated
-        # on Avocado versions >= 50.0.  This is a temporary compatibility
-        # enabler for older runners, but should be removed soon
-        if detected_distro.name in ["rhel", "fedora", "centos", "redhat"]:
+        if detected_distro.name in ["redhat", "fedora", "centos"]:
             depends.append("openssh-clients")
         if detected_distro.name == "SuSE":
             depends.append("openssh")
@@ -110,13 +107,16 @@ class Bonding(Test):
             self.net_mask.append(mask)
         self.bonding_slave_file = "/sys/class/net/%s/bonding/slaves"\
                                   % self.bond_name
+        self.peer_bond_needed = self.params.get("peer_bond_needed",
+                                                default="False")
+        self.wait_time = self.params.get("wait_time", default="5")
 
     def bond_remove(self, arg1):
         '''
         bond_remove
         '''
         if arg1 == "local":
-            self.log.info("Bonding configuration removed on laocal")
+            self.log.info("Removing Bonding configuration on local machine")
             self.log.info("------------------------------------------------")
             for ifs in self.host_interfaces:
                 cmd = "ifconfig %s down" % ifs
@@ -130,7 +130,7 @@ class Bonding(Test):
                 self.log.info("bond removing command failed in local machine")
             time.sleep(5)
         else:
-            self.log.info("Bonding configuration removed on Peer machine")
+            self.log.info("Removing Bonding configuration on Peer machine")
             self.log.info("------------------------------------------------")
             cmd = ''
             cmd += 'ifconfig %s down;' % self.bond_name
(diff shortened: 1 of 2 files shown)


io/net/bonding.py
```diff
@@ -45,10 +45,7 @@ def setUp(self):
         depends = []
         if detected_distro.name == "Ubuntu":
             depends.append("openssh-client")
-        # FIXME: "redhat" as the distro name for RHEL is deprecated
-        # on Avocado versions >= 50.0.  This is a temporary compatibility
-        # enabler for older runners, but should be removed soon
-        if detected_distro.name in ["rhel", "fedora", "centos", "redhat"]:
+        if detected_distro.name in ["redhat", "fedora", "centos"]:
```
narasimhan-v: Please do not delete this. Have it as it was before.

io/net/bonding.py
```diff
@@ -110,13 +107,16 @@ def setUp(self):
             self.net_mask.append(mask)
         self.bonding_slave_file = "/sys/class/net/%s/bonding/slaves"\
                                   % self.bond_name
+        self.peer_bond_needed = self.params.get("peer_bond_needed",
+                                                default="False")
```
narasimhan-v: default=False

io/net/bonding.py
```diff
@@ -165,20 +165,47 @@ def bond_fail(self, arg1):
...
+                cmd = "cat /proc/net/bonding/%s" % self.bond_name
+                process.system_output(cmd, shell=True, verbose=True)
+                cmd = "ifconfig %s up" % interface
+                time.sleep(10)
+                if process.system(cmd, shell=True, ignore_status=True) != 0:
+                    self.fail("bonding not working when trying to up the\
+                              interface %s" % interface)
+                time.sleep(5)
+            else:
+                self.fail("Need a min of 2 host interfaces to to test\
```
narasimhan-v: Should not fail here. Can just print a warn message or info

io/net/bonding.py
```diff
@@ -165,20 +165,47 @@ def bond_fail(self, arg1):
...
+        process.system_output(cmd, shell=True, verbose=True)
+        if process.system(ping_cmd, shell=True, ignore_status=True) != 0:
+            self.log.info("Ping test failed which is expected ")
+        for interface in self.host_interfaces:
             cmd = "ifconfig %s up" % interface
+            time.sleep(10)
             if process.system(cmd, shell=True, ignore_status=True) != 0:
-                self.fail("bonding not working when trying to up the\
-                          interface %s" % interface)
+                self.log.info("Ping to the bonding interface failed which\
```
narasimhan-v: There is no ping tried here, but the info says ping failed

io/net/bonding.py.data/bonding.yaml
```diff
@@ -18,3 +18,5 @@ peer_ip: "85.85.85.85"
 peer_interfaces: "eth3"
 bond_name: "bond13"
 user_name: "root"
+peer_bond_needed: False
+wait_time: "10"
```
narasimhan-v: This should ideally be used in all time.sleep()

io/net/bonding.py
```diff
@@ -165,20 +165,47 @@ def bond_fail(self, arg1):
...
-            cmd = "cat /proc/net/bonding/%s" % self.bond_name
-            process.system_output(cmd, shell=True, verbose=True)
+                self.fail("Could not bring down the interface %s " % interface)
+            time.sleep(10)
+        ping_cmd = "ping -I %s %s -c 5"\
+            % (self.bond_name, self.peer_first_ipinterface)
+        cmd = "cat /proc/net/bonding/%s" % self.bond_name
+        process.system_output(cmd, shell=True, verbose=True)
+        if process.system(ping_cmd, shell=True, ignore_status=True) != 0:
+            self.log.info("Ping test failed which is expected ")
```
narasimhan-v: This should ideally use the `ping_check()`.

io/net/bonding.py.data/bonding.yaml
hathyaga: can it be used only where it is needed? The intention to use this is to wait for the Bond interface to come up in the peer. If this is set to any large value, then the entire script will take a long time to finish

io/net/bonding.py
vrbagalkote: Can all three 'if' be a part of if-else case?

io/net/bonding.py.data/bonding.yaml
narasimhan-v: @hathyaga If it is for peer only, please indicate that. The parameter name can be `peer_wait_time`.
Regarding script taking long time, there are ways to implement this with wait time :-)

io/net/bonding.py
vrbagalkote: @narasimhan-v If warn or info is given, test will still continue ahead right? If minimum number of interfaces are not given, test should not continue, So I fell fail is right. Plz let me know your thoughts.

narasimhan-v: Theoretically bonding will work even if it has one slave only.
So, we should save the failed case, and fail in the last.

hathyaga:
Nrew PR having changes suggested: https://github.com/avocado-framework-tests/avocado-misc-tests/pull/484

Closing this PR.

PR #489, opened 2017-05-12
Handling package dependencies in infiniband tests

narasimhan-v:
Handled dependencies for ssh and iputils packages for different
distros for all infiniband tests.

### DIFF
--- a/io/net/infiniband/ib_bw_perf.py
+++ b/io/net/infiniband/ib_bw_perf.py
@@ -23,8 +23,8 @@
 
 
 import time
-from avocado import main
 import netifaces
+from avocado import main
 from avocado import Test
 from avocado.utils.software_manager import SoftwareManager
 from avocado.utils import process, distro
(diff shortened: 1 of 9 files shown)


io/net/infiniband/ib_bw_perf.py
```diff
@@ -40,10 +40,17 @@ def setUp(self):
...
-            if not sm.check_installed(pkg) and not sm.install(pkg):
+        smm = SoftwareManager()
+        detected_distro = distro.detect()
+        pkgs = ["perftest"]
+        if detected_distro.name == "Ubuntu":
+            pkgs.append('openssh-client')
+        elif detected_distro.name == "SuSE":
+            pkgs.append('openssh')
+        else:
+            pkgs.append('openssh-clients')
```
vrbagalkote: is it not better to check specifically with Red-Hat/Rhel, rather than executing in else.

narasimhan-v: Yes, but the package name in `rhel` is also `openssh-clients`.

PraveenPenguin:
@vrbagalkote thanks for review

narasimhan-v:
@PraveenPenguin Thanks

PR #492, opened 2017-05-15
Use whiteboard for performance data

clebergnu:
The following five tests use a custom file to record performance
data.  Avocado test API makes an attribute available that will
automatically preserve test data in any arbitraty format, the
whiteboard, documented here:

  http://avocado-framework.readthedocs.io/en/49.0/WritingTests.html#saving-test-generated-custom-data

This patch is a suggestion/RFC to just use the whiteboard instead of
yet another external file.  The upside is less code and possible
future goodies that the framework may bring (for instance, we may show
the whiteboard by default on job reports).  The downside of this
change is that, if there are scripts consuming this performance data,
they'd have to be adjusted to use the whiteboard file path.

### DIFF
--- a/cpu/ebizzy.py
+++ b/cpu/ebizzy.py
@@ -87,12 +87,10 @@ class Ebizzy(Test):
         usr_time = pattern.findall(results)[0]
         pattern = re.compile(r"sys (.*?) s")
         sys_time = pattern.findall(results)[0]
-
-        perf_json = {'records': records, 'real_time':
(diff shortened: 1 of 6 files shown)


narasimhan-v:
@clebergnu Thanks for the PR. IMO, the upside outweighs the downside.

Just a quick question on the upside you mentioned: if decided to show the whiteboard by default on job reports, and there are more than 1 files in whiteboard, will that be handled too ?

clebergnu:
@narasimhan-v sure, sending a v2.

PR #495, opened 2017-05-16
Fix Pmqa test to fail for non root users

harish-24:
fix pmqa test to fail for non root users. Previously it was a false PASS.


PraveenPenguin:
@narasimhan-v thanks

PR #497, opened 2017-05-17
Check the Arcconf package and remove the http repo path in yaml file

naveedus:
This patch will check the Arcconf package with correct case letter,
also will remove the default tool and firmware path for the adapter.

Signed-off-by: Naveed Ahmed U S <naveedus@linux.vnet.ibm.com>


PraveenPenguin:
@naveedus @narasimhan-v thanks

PR #498, opened 2017-05-17
Use whiteboard for performance data [v2]

clebergnu:
The following five tests use a custom file to record performance
data.  Avocado test API makes an attribute available that will
automatically preserve test data in any arbitraty format, the
whiteboard, documented here:

  http://avocado-framework.readthedocs.io/en/49.0/WritingTests.html#saving-test-generated-custom-data

This patch is a suggestion/RFC to just use the whiteboard instead of
yet another external file.  The upside is less code and possible
future goodies that the framework may bring (for instance, we may show
the whiteboard by default on job reports).  The downside of this
change is that, if there are scripts consuming this performance data,
they'd have to be adjusted to use the whiteboard file path.

---

Changes from v1 (#492):
 * Style fix

### DIFF
--- a/cpu/ebizzy.py
+++ b/cpu/ebizzy.py
@@ -87,11 +87,10 @@ class Ebizzy(Test):
         usr_time = pattern.findall(results)[0]
         pattern = re.compile(r"sys (.*?) s")
         sys_time = pattern.findall(results)[0]
-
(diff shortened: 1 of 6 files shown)


PraveenPenguin:
@narasimhan-v thanks

PR #499, opened 2017-05-18
Fix cpupower cpu_freq_path function return value

narasimhan-v:
scaling_available_frequencies file had a space in the end, which
when split, returned an empty element in array.
Fixed that with returning after stripping the space in the end.


PraveenPenguin:
@abdhaleegit can you please verify as this fix working for you ?

PraveenPenguin:
@abdhaleegit thanks for ack

PR #500, opened 2017-05-18
fuzz/fsfuzzer : Add dependency packages

abdhaleegit:
With out these packages test will not continue
so added in the list.


PraveenPenguin:
@abdhaleegit @narasimhan-v thanks

PR #513, opened 2017-05-25
Bonding script enhancement

hathyaga:
The enhancement contains an option to create a Bonding Peer
interface if needed and failure check added when all interfaces
are down

### DIFF
--- a/io/net/bonding.py
+++ b/io/net/bonding.py
@@ -43,15 +43,15 @@ class Bonding(Test):
         detected_distro = distro.detect()
         sm = SoftwareManager()
         depends = []
-        if detected_distro.name == "Ubuntu":
-            depends.append("openssh-client")
         # FIXME: "redhat" as the distro name for RHEL is deprecated
         # on Avocado versions >= 50.0.  This is a temporary compatibility
         # enabler for older runners, but should be removed soon
-        if detected_distro.name in ["rhel", "fedora", "centos", "redhat"]:
-            depends.append("openssh-clients")
-        if detected_distro.name == "SuSE":
-            depends.append("openssh")
+        if detected_distro.name == "Ubuntu":
+            depends.extend(["openssh-client", "iputils-ping"])
+        elif detected_distro.name in ["rhel", "fedora", "centos", "redhat"]:
+            depends.extend(["openssh-clients", "iputils"])
+        else:
+            depends.extend(["openssh", "iputils"])
         for pkg in depends:
             if not sm.check_installed(pkg) and not sm.install(pkg):
                 self.skip("%s package is need to test" % pkg)
@@ -77,6 +77,7 @@ class Bonding(Test):
         if self.peer_first_interface == "":
             self.fail("test failed because peer interface can not retrieved")
         self.bond_name = self.params.get("bond_name", default="tempbond")
+        self.bond_status = "cat /proc/net/bonding/%s" % self.bond_name
         self.mode = self.params.get("bonding_mode", default="")
         if self.mode == "":
             self.skip("test skipped because mode not specified")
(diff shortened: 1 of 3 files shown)


hathyaga:
Job log run for all modes with a peer interface created
[job_log.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1028445/job_log.txt)

io/net/bonding.py
```diff
@@ -221,9 +253,16 @@ def bond_setup(self, arg1, arg2):
...
+            process.system(cmd, shell=True, ignore_status=True)
+            for i in range(0, 600, 60):
+                if 'state UP' in process.system_output("ip link \
+                     show %s" % self.bond_name, shell=True):
+                    self.log.info("Bonding setup is successful on\
+                                  local machine")
+                    break
+                else:
+                    self.fail("Bonding setup on local machine has failed")
+                time.sleep(60)
```
narasimhan-v:
```
            for i in range(0, 600, 60):
                if 'state UP' in process.system_output("ip link \
                    show %s" % self.bond_name, shell=True):
                    self.log.info("Bonding setup is successful on\
                                  local machine")
                    break
                time.sleep(60)
            else:
                self.fail("Bonding setup on local machine has failed")
```

io/net/bonding.py
```diff
@@ -283,19 +323,27 @@ def tearDown(self):
...
-            time.sleep(5)
+            process.system(cmd, shell=True, ignore_status=True)
+            for i in range(0, 600, 60):
+                if 'state UP' in process.system_output("ip link \
+                     show %s" % val1, shell=True):
+                    self.log.info("Interface %s is up" % val1)
+                    break
+                else:
+                    self.log.info("Interface %s in not up\
+                                   in the host machine" % val1)
```
narasimhan-v:
```
            for i in range(0, 600, 60):
                if 'state UP' in process.system_output("ip link \
                     show %s" % val1, shell=True):
                    self.log.info("Interface %s is up" % val1)
                    break
                time.sleep(60)
            else:
                self.log.info("Interface %s in not up\
                              in the host machine" % val1)
```

io/net/bonding.py
```diff
@@ -157,29 +162,57 @@ def ping_check(self, arg1):
...
+                process.system_output(self.bond_status, shell=True,
+                                      verbose=True)
+                cmd = "ifconfig %s up" % interface
+                time.sleep(self.sleep_time)
+                if process.system(cmd, shell=True, ignore_status=True) != 0:
+                    self.fail("Not able to bring up the slave\
+                                    interface %s" % interface)
+                time.sleep(self.sleep_time)
+        else:
+            self.log.warning("Need a min of 2 host interfaces to test\
```
narasimhan-v: Should this be a debug or warn ?
When I run with only one interface I get a WARN, but would like to see a PASS instead

`RESULTS    : PASS 0 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 1 | INTERRUPT 0 | CANCEL 0`

narasimhan-v:
@hathyaga It is almost done, except for the few changes suggested.
With the suggested changes the test works fine for an `single interface` that `does not come up immediately` (which is the issue this PR handles)

io/net/bonding.py
hathyaga: I agree. It should be as "Debug" otherwise it does not look good that the test shows a warning even when it has passed.

io/net/bonding.py
hathyaga: Will be making this change here and above and will send a PR momentarily

hathyaga:
Closing this PR as new PR has been sent: https://github.com/avocado-framework-tests/avocado-misc-tests/pull/516

PR #517, opened 2017-05-26
Extend test category using tags

clebergnu:
The tests on this repo are already well categorized by the directory
structure, but that can only give so much info about them. Let's also
make use of some basic (initially) tags.

Some exampels of the immediate benefits include listing/running only
tests that can run as regular users:

 $ avocado list --filter-by-tags=-privileged cpu/
 INSTRUMENTED cpu/ebizzy.py:Ebizzy.test
 INSTRUMENTED cpu/linsched.py:Linsched.test

Listing/running tests that require a Power architecture:

 $ avocado list --filter-by-tags=power cpu/
 INSTRUMENTED cpu/cpupower.py:cpupower.test
 INSTRUMENTED cpu/ppc64_cpu_test.py:PPC64Test.test

This is a *partial* implementation (only includes tests under the
"cpu" directory) of the extended information using tags, intended as
an RFC (or initial step).  This is related to an RFC posted on
the mailing list.

Reference: https://www.redhat.com/archives/avocado-devel/2017-May/msg00007.html

### DIFF
--- a/cpu/sensors.py
+++ b/cpu/sensors.py
@@ -33,6 +33,8 @@ class Sensors(Test):
 
     """
     Run sensors command on ppc architectures.
+
+    :avocado: tags=cpu,power,privileged
     """
 
     @staticmethod
--- a/README.rst
+++ b/README.rst
@@ -5,9 +5,6 @@ This repository is dedicated to host any tests written using the avocado [1]
 API. It is being initially populated with tests ported from autotest
 client tests repository, but it's not limited by that.
 
-Tests will be organized per category basis, each category  with its own
-directory.
-
 Once you have the avocado installed, you can run the tests like below::
 
     $ avocado run avocado-misc-tests/perf/stress.py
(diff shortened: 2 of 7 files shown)


cpu/sensors.py
```diff
@@ -33,6 +33,8 @@ class Sensors(Test):
 
     """
     Run sensors command on ppc architectures.
+
+    :avocado: tags=cpu,power,privileged
```
harish-24: sensors test can ran on x86 as well. I think `power` tag can be removed here.

clebergnu: @harish-24 right, but it looks like *this* test is specific to power.  A few facts:
 * docstring says it: "Run sensors command on ppc architectures."
 * there's a specific check for `CONFIG_SENSORS_IBMPOWERNV` and the test is skipped if it's not set

Can you please double check your suggestion in light of those? Thanks.

PraveenPenguin: @clebergnu sorry may be doc string having issue will fix ,
*  checking `CONFIG_SENSORS_IBMPOWERNV` config only i think when distro match with ppc please have a look in this 

https://github.com/avocado-framework-tests/avocado-misc-tests/blob/master/cpu/sensors.py#L69

harish-24: @clebergnu @PraveenPenguin  is right. I guess docstring needs a change. This is not a ppc specific test. Thanks.

clebergnu: Right, I missed the conditional on the distro.  So, let's fix the docstring and the tag.  Thanks!

PraveenPenguin: @clebergnu we fixed the doc string can you please have look and possible raise a rectified PR

PR #520, opened 2017-05-31
Improvements to avago3008.py script

vrbagalkote:
With this patch, user can just pass the tool location in yaml
file, instead of making it mandatory to place the script in a
particular folder

### DIFF
--- a/io/disk/Avago_storage_adapter/avago3008.py
+++ b/io/disk/Avago_storage_adapter/avago3008.py
@@ -21,6 +21,7 @@ This script will list all the adapter connected to the system.
 """
 
 
+import os
 import time
 from avocado import Test
 from avocado.utils import process
@@ -44,9 +45,13 @@ class Avago3008(Test):
         self.disk = str(self.params.get('disk')).split(" ")
         self.spare = str(self.params.get('spare'))
         self.size = int(self.params.get('size', default='max'))
+        self.tool_location = self.params.get('tool_location')
         if not self.disk:
             self.skip("Please provide disks to run the tests")
         self.number_of_disk = len(self.disk)
+        if (not os.path.isfile(self.tool_location) or not
+           os.access(self.tool_location, os.X_OK)):
+            self.skip()
 
         self.dict_raid = {'raid0': [2, None, None], 'raid1': [2, 2, None],
                           'raid1e': [3, None, None],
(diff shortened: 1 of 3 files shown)


io/disk/Avago_storage_adapter/avago3008.py
```diff
@@ -44,6 +44,7 @@ def setUp(self):
         self.disk = str(self.params.get('disk')).split(" ")
         self.spare = str(self.params.get('spare'))
         self.size = int(self.params.get('size', default='max'))
+        self.tool_location = self.params.get('tool_location')
```
PraveenPenguin: can this will be a default value ?

vrbagalkote: User is expected to to give this, and we can't have default values, as if the tool is not present in that location, script will error out.

narasimhan-v: @vrbagalkote I would suggest to skip the test in case the tool is not present.

```if not os.path.isfile(self.tool_location) or not os.access(self.tool_location, os.X_OK):
    self.skip()```

File: io/disk/Avago_storage_adapter/avago3008.py.data/avago3008.yaml [hunk 2]
```diff
@@ -1,8 +1,9 @@
 cenario:
     contorller: 0
-    disk: 1:4 1:5 1:6
-    spare: "1:7"
+    disk: 2:1 2:4 2:5 2:6
```
narasimhan-v: Can we leave `disk` and `spare` empty, so that users who run by mistake do not suffer from loss of disk data.

vrbagalkote:
@PraveenPenguin @narasimhan-v addressed ur comments.

io/disk/Avago_storage_adapter/avago3008.py
```diff
@@ -44,9 +45,13 @@ def setUp(self):
...
         if not self.disk:
             self.skip("Please provide disks to run the tests")
         self.number_of_disk = len(self.disk)
+        if (not os.path.isfile(self.tool_location) or not
```
narasimhan-v: @vrbagalkote we can have this check immediately after reading the yaml parameter `self.tool_location`. Except that, looks good.

vrbagalkote: If its upto me to decide, then the current version looks good to me.

PraveenPenguin:

PR #456, opened 2017-04-24
Added the check for VCPU, added failure info for all failures, optimi…

shriyak:
…zed the code and also fixed pep8 issues.

24*7 performance counters is supported on Power VM platform.
Script will cover basic testing of events in 24x7 counters that belong to different
domains like core,chip,virtual CPU home core, etc.To monitor an event, the user must
specify the required parameters for the event,which include its name,
the chip/core on which to monitor the event as well as the domain.

Signed-off-by: Shriya <shriyak@linux.vnet.ibm.com>

PR #526, opened 2017-06-06
Test to verify 24x7 hardware counters 

shriyak:
24*7 performance counters is supported on Power VM platform.
Script will cover basic testing of events in 24x7 counters that belong to different
domains like core,chip,virtual CPU home core, etc.To monitor an event, the user must
specify the required parameters for the event,which include its name,
the chip/core on which to monitor the event as well as the domain.
Added the check for VCPU, added failure info for all failures, optimized the code.


shriyak:
Previous PR : https://github.com/avocado-framework-tests/avocado-misc-tests/pull/456

PR #528, opened 2017-06-06
Test to verify raw event code for each generic event

shriyak:
Performance Monitoring Unit (PMU) provides a list of events to measure micro-architectural events such as the number of cycles,instructions retired, L1 cache misses and so on.Those events are called PMU hardware events or hardware events.The generic hardware and cache perf events for Power are exported to sysfs,so users can precisely determine the PMU event monitored by the generic event.
The script tests Display event codes for Generic HW (PMU) events.It will read content of each file from
/sys/bus/event_source/devices/cpu/events and compare the raw event code for each generic event.

### DIFF
--- a/perf/perf_genericevents.py
+++ b/perf/perf_genericevents.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


PraveenPenguin:
@shriyak please rebase with remove  `Unused import re` it make tarvis happy

PR #534, opened 2017-06-09
Energy mangement: Tuned tests

shriyak:
“Tuned” in RHEL is tuning daemon for automatically tuning the system via the use of tuning profile. It can also be configured
to react to changes to improve performance of the server and also to make system settings persistent.
tuned-adm is a command line tool that provides a number of different profiles to improve performance in a number of specific use cases
tuned-adm includes the ability to run any command as part of enabling or disabling a tuning profile.
Several predefined profiles are already included. You can even create your own profile, either based on one of the existing ones by copying it or make a completely new one. The distribution provided profiles are stored in subdirectories below /usr/lib/tuned
and the user defined profiles in subdirectories below /etc/tuned. If there are same profiles on both places, user defined profiles have precedence.

### DIFF
--- a/cpu/em_tuned.py
+++ b/cpu/em_tuned.py
@@ -1,0 +1,213 @@
+#!/usr/bin/env python
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
+# Author: Shriya Kulkarni <shriyak@linux.vnet.ibm.com>
+# Author: Satheesh Rajendran <sathnaga@linux.vnet.ibm.com>
+
+from avocado import Test
+from avocado import main
+from avocado.utils import process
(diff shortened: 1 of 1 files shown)


cpu/em_tuned.py
```diff
@@ -0,0 +1,213 @@
...
+        if status != 0:
+            self.error_count += 1
+            self.log.info("Error setting Profile %s" % profile)
+
+    def get_available_profiles(self):
+        """
+        To get all available profiles
+        """
+        cmd_profile_list = "tuned-adm list|grep '^\-'|\
+                           sed 's/^- //' | awk '{print $1}'"
```
abdhaleegit: Can we use python alternative for shell command? like
`
cmd_profile_list = []
cmd_profile = commands.getoutput('tuned-adm list').split('\n')
for line in cmd_profile:
        if line startswith("- "):
               cmd_profile.append[line]
`

cpu/em_tuned.py
```diff
@@ -0,0 +1,213 @@
...
+
+class em_tuned(Test):
+
+    def setUp(self):
+        detected_distro = distro.detect()
+        if detected_distro.name == 'rhel':
+            self.log.info("Tuned-adm service is supported on rhel")
+        else:
+            self.cancel(" Tuned-adm service is not supported "
+                        "on %s" % detected_distro.name)
```
shriyak: sure , I will make the changes

cpu/em_tuned.py
shriyak: sure let me try it.

shriyak:
I will open new PR with the changes requested.

PR #539, opened 2017-06-15
generic/rcutorture: process.system() did not run the command

abdhaleegit:
CPU off-on commands from process.system() function never got
executed, replace with os.system. Also added tags to be a kernel
tests and replace skip with cancel

### DIFF
--- a/generic/rcutorture.py
+++ b/generic/rcutorture.py
@@ -33,6 +33,8 @@ class Rcutorture(Test):
     CONFIG_RCU_TORTURE_TEST enables an intense torture test of the RCU
     infratructure. It creates an rcutorture kernel module that can be
     loaded to run a torture test.
+
+    :avocado: tags=kernel,privileged
     """
 
     def setUp(self):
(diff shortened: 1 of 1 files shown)


generic/rcutorture.py
```diff
@@ -62,37 +64,31 @@ def cpus_toggle(self):
...
             time.sleep(10)
-            online = 'echo 1 > /sys/devices/system/cpu/cpu%s/online' % cpu
-            process.system(online)
+            os.system('echo 1 > /sys/devices/system/cpu/cpu%s/online' % cpu)
 
         self.log.info("Offline and online second half cpus %s\n", scpu)
         for cpu in range(shalf_count, full_count):
-            offline = 'echo 0 > /sys/devices/system/cpu/cpu%s/online' % cpu
-            process.system(offline)
+            os.system('echo 0 > /sys/devices/system/cpu/cpu%s/online' % cpu)
```
narasimhan-v: Any idea why `process.system()` did not run ? Was it because `shell=True` was not set ?

PraveenPenguin: strange for me it always works

abdhaleegit: Yes, shell=True works fine. Thanks
But, how about adding cpu_online() and cpu_offline(0 to utils/cpu.py  ? as it will be used in many tests ?

narasimhan-v: @abdhaleegit Yes, makes sense.

abdhaleegit: @narasimhan-v @PraveenPenguin 

Once this is merged https://github.com/avocado-framework/avocado/pull/2074
I will rework on this patch and sent the latest code

PraveenPenguin: @abdhaleegit i think we can wait for it as Lukáš already start review this patch

PraveenPenguin:
@abdhaleegit closing this PR as new version is raised

PR #544, opened 2017-06-20
Added bridge interface test

harish-24:
Test the ethernet bridge configuration in the linux kernel using brctl command

### DIFF
--- a/io/net/bridge.py
+++ b/io/net/bridge.py
@@ -1,0 +1,112 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2017 IBM
+# Author: Harish Sriram <harish@linux.vnet.ibm.com>
+# Bridge interface test
+
+
+import netifaces
+from avocado import main
+from avocado import Test
+from avocado.utils.software_manager import SoftwareManager
+from avocado.utils import process
+
+
+class Bridging(Test):
+    '''
+    Test bridge interface
+    '''
+
+    def check_failure(self, cmd):
+        if process.system(cmd, sudo=True, shell=True, ignore_status=True):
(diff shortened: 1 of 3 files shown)


io/net/bridge.py
```diff
@@ -0,0 +1,98 @@
...
+
+        cidr_cmd = "ip addr show %s | sed -nr 's/.*inet ([^ ]+)."\
+                   "*/\\1/p'" % self.host_interface
+        self.cidr = process.system_output(
+            '%s' % cidr_cmd, shell=True)
+        route_cmd = "route -n | grep %s | grep -w UG | awk "\
+                    "'{ print $2 }'" % self.host_interface
+        self.gateway = process.system_output(
+            '%s' % route_cmd, shell=True)
+        brd_cmd = "ip addr show %s | grep inet | grep brd | "\
```
narasimhan-v: `cidr_cmd`, `route_cmd`, etc variables can be `cmd`

io/net/bridge.py.data/bridge.yaml
```diff
@@ -0,0 +1,2 @@
+host_interface: "eth0"
+peer_ip: "9.114.15.234"
```
narasimhan-v: Please leave these empty

harish-24:
@narasimhan-v Pushed. Thanks

io/net/bridge.py
```diff
@@ -0,0 +1,99 @@
...
+            "awk '{ print $4 }'" % self.host_interface
+        self.broadcast = process.system_output(
+            '%s' % cmd, shell=True)
+
+    def test(self):
+        '''
+        Bridge Test
+        Set up the ethernet bridge configuration in the linux kernel
+        '''
+        process.system('brctl addbr br0', sudo=True)
```
PraveenPenguin: can we have a pram as `ignore_status=True` it help in case when command failed

PraveenPenguin: can we need check as bridge added or not

PraveenPenguin: this applicable all the places where `process.system` used

harish-24: @PraveenPenguin I think each step is as important as any other and if it fails, we should not be going to next stepand better to raise the `CmdError` exception.

harish-24: sure, will add it

PraveenPenguin: raising CmdError i think not good i think we have proper validation for every command if command output is failed (not matched with desired  output ) need to fail the test

harish-24:
@PraveenPenguin Pushed with the changes. Thanks !

PraveenPenguin:
@narasimhan-v thanks

PR #554, opened 2017-07-03
Portbounce testcase for fc switches.

pridhiviraj:
This testcase executes the portbounce test for adapters which
are under fc switch ports.

portbounce test is basically enable/disabling fc port for short
& long durations.

Currently full testcase is implemented for fc switches as setup
available.

TODO: Update fcoe switch login and run command API's. And also
update portdisable test for fcoe switch

Signed-off-by: Pridhiviraj Paidipeddi <ppaidipe@linux.vnet.ibm.com>

### DIFF
--- a/io/net/port_bounce.py
+++ b/io/net/port_bounce.py
@@ -1,0 +1,199 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+# Copyright: 2017 IBM
+# Author: Pridhiviraj Paidipeddi <ppaidipe@linux.vnet.ibm.com>
+# this script runs portbounce test on different ports of fc or fcoe switches.
+
+import re
+import time
+import telnetlib
+
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+
+
+class CommandFailed(Exception):
+    def __init__(self, command, output, exitcode):
+        self.command = command
+        self.output = output
+        self.exitcode = exitcode
+
+    def __str__(self):
+        return "Command '%s' exited with %d.\nOutput:\n%s" \
+               % (self.command, self.exitcode, self.output)
+
+
+class PortBounceTest(Test):
+
+    """
+    :param type: <fc/fcoe> : type of switch fc/fcoe <in small case>
+    :param fcoe_fc: <yes/no> : If port is an FC port in FCOE switch
+    :param switch_name: FC Switch name/ip
+    :param userid: FC switch user name to login
+    :param password: FC switch password to login
+    :param port_ids: FC switch port ids where port needs to disable/enable
+    :param sbt: short bounce time in seconds
+    :param lbt: long bounce time in seconds
+    :param count: Number of times test to run
+    """
+
+    def setUp(self):
+        """
+        test parameters
+        """
+        self.parameters()
+
+    def parameters(self):
+        self.fc_type = self.params.get("type", '*', default=None)
+        self.fcoe_fc = self.params.get("fcoe_fc", '*', default=None)
+        self.switch_name = self.params.get("switch_name", '*', default=None)
+        self.userid = self.params.get("userid", '*', default=None)
+        self.password = self.params.get("password", '*', default=None)
+        self.port_ids = self.params.get("port_ids",
+                                        '*', default=None).split(",")
+        self.sbt = int(self.params.get("sbt", '*', default=5))
+        self.lbt = int(self.params.get("lbt", '*', default=250))
+        self.count = int(self.params.get("count", '*', default="2"))
+        self.prompt = ">"
+
+    def fc_login(self, ip, username, password):
+        '''
+        telnet Login method for remote fc switch
+        '''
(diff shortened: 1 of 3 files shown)


io/net/port_bounce.py
```diff
@@ -0,0 +1,156 @@
...
+        self.output = output
+        self.exitcode = exitcode
+
+    def __str__(self):
+        return "Command '%s' exited with %d.\nOutput:\n%s" \
+               % (self.command, self.exitcode, self.output)
+
+
+SHORT_BOUNCE_TIME = 5
+LONG_BOUNCE_TIME = 250
```
narasimhan-v: Better to take it as arguments from yaml file

io/net/port_bounce.py.data/README.txt
```diff
@@ -0,0 +1,13 @@
+Port bounce Testcase:
+
+this testcase mainly depends on the setup of fc or fcoe switch to
+connect different adapters to different ports in respective switches.
+
```
narasimhan-v: Also mention, this test is applicable only for Brocade Switches

io/net/port_bounce.py
```diff
@@ -0,0 +1,156 @@
...
+        for port in self.port_ids:
+            self.log.info("Portbounce test for port %s for %s times",
+                          port, self.count)
+            for i in range(self.count+1):
+                self.log.info("Port %s, Short Portbounce test", port)
+                test(self, port, SHORT_BOUNCE_TIME)
+                self.log.info("Port %s, Long Portbounce test", port)
+                test(self, port, LONG_BOUNCE_TIME)
+                self.log.info("%s portdisable/portenable tests completed for \
+                              Port %s", i+1, port)
```
narasimhan-v: There can be a scenario of disabling all ports at once, for a user specified time.

io/net/port_bounce.py
```diff
@@ -0,0 +1,156 @@
...
+
+    @staticmethod
+    def portdisable(self, port, sleep_time):
```
abdhaleegit: This function does both disable and enable. can we rename it to porttoggle() ?

io/net/port_bounce.py
```diff
@@ -0,0 +1,156 @@
...
+        except CommandFailed as cf:
+            self.log.info("port disable failed for port %s, details: %s",
+                          port, str(cf))
+        time.sleep(sleep_time)
+        self.log.info("Enable port %s", port)
+        try:
+            self.fc_run_command("portenable %s" % port)
+        except CommandFailed as cf:
+            self.log.info("port enable failed for port %s, details: %s",
+                          port, str(cf))
```
abdhaleegit: We can make sure port is really disabled/enabled by sending 'port info <portid>' or from 'switchshow' command

abdhaleegit: Also we expect 'remote port timeout' messages  or 'No light siganl' at linux side, can we check that ?

io/net/port_bounce.py
```diff
@@ -0,0 +1,156 @@
...
+    def portdisable_fcoe_fc(self, port, sleep_time):
+        pass
+
+    def tearDown(self):
+        output = process.system_output("dmesg -T --level=alert,crit,err,warn",
+                                       ignore_status=True,
+                                       shell=True, sudo=True)
+
+        self.log.debug("Kernel Errors: %s", output)
+
```
abdhaleegit: as part of cleanup, Make sure all ports are connected again. 'switchshow' ?

io/net/port_bounce.py
abdhaleegit: you can do multiple port disable/enable : portdisable B11 B14  or portenable B10 B20  etc.

io/net/port_bounce.py.data/README.txt
abdhaleegit: Test can be improved to also do a storage side port off on ?

[Server] <  p1- [switch] - p4 >   [storage]

io/net/port_bounce.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+        self.parameters()
+
+    def parameters(self):
+        self.fc_type = self.params.get("type", '*', default=None)
+        self.fcoe_fc = self.params.get("fcoe_fc", '*', default=None)
+        self.switch_name = self.params.get("switch_name", '*', default=None)
+        self.userid = self.params.get("userid", '*', default=None)
+        self.password = self.params.get("password", '*', default=None)
+        port_ids = self.params.get("port_ids", '*', default=None)
```
PraveenPenguin: can this be like 
`self.port_ids =  self.params.get("port_ids", '*', default=None).split(",")` want to eliminate extra variable as it make GC more work to do

pridhiviraj:
@narasimhan-v @abdhaleegit @PraveenPenguin  Thanks for the review. I will work on the comments and sent a V2.

io/net/port_bounce.py.data/README.txt
narasimhan-v: @abdhaleegit 2 things to note here:

1. The current implementation will take care of this, if given storage port ids instead of host port ids.
2.  The problem with doing that is, all the other FC hosts will be affected by disabling/enabling storage ports. So, that should be handled with caution.

narasimhan-v:
@pridhiviraj I would suggest to get the multipath device name as input, and after each and every portbounce, check if IO still happens via that multipath (ex: iostat)

PraveenPenguin:
@pridhiviraj any update here ?

narasimhan-v:
@pridhiviraj That is a seperate test, as it needs to consider timeouts of individual paths as well. So, we can take care of that as an enhancement.

@PraveenPenguin Test looks good to me

PraveenPenguin:
@pridhiviraj thanks for the patch

PraveenPenguin:
@narasimhan-v @abdhaleegit thanks

PR #560, opened 2017-07-06
Improvements for HTX on NIC Interface Stress test.

pridhiviraj:
HTX Changes the test interface ips before starting the
test, due to which the commands to be executed in peer
will no more functional.
So this patch has below improvements.

1. Falling back to password based ssh to the public IP
which is not going to be effected by HTX.
2. Imprved pingum test work flow. If N/W interfaces are
in different topology N/W ping test takes some time.
So running the pingum test in a loop for 10 times until
it passes.
3. As HTX changing test interface ips, we need to restore
back to original ips. So in tearDown handling this.

Signed-off-by: Pridhiviraj Paidipeddi <ppaidipe@linux.vnet.ibm.com>

### DIFF
--- a/io/net/htx_nic_devices.py
+++ b/io/net/htx_nic_devices.py
@@ -17,9 +17,13 @@
 import os
 import re
 import time
-import pexpect
-import sys
+try:
+    import pxssh
+except ImportError:
+    from pexpect import pxssh
+
 from avocado import Test
+from avocado.utils import distro
 from avocado import main
 from avocado.utils import process
 from avocado.utils.process import CmdError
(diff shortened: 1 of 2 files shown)


pridhiviraj:
@PraveenPenguin Hold this for some time, We want to implement restoring back the test ips on ubuntu as well and also on cross OS setup as well.(For ex: Host is Ubuntu & Peer is SuSE).

PraveenPenguin:
@pridhiviraj sure

pridhiviraj:
@PraveenPenguin Please merge this. we have implemented this test to work on cross OS distro setup. But still we need to implement this for ubuntu.(This can be done later.)

PraveenPenguin:
@pridhiviraj thanks for the patch

PraveenPenguin:
@narasimhan-v thanks for ACK

PR #568, opened 2017-07-12
Fix enviroinment check for power systems

harish-24:
cpu._get_cpu_info returns only 1st cpu details and hence fails to check bare-metal enviroinment. This fixes it.


vrbagalkote:
It gives the desired o/p, I tried it on PowerNV machine

>>> from avocado.utils import cpu
>>> if not cpu._list_matches(open('/proc/cpuinfo').readlines(),'platform\t: PowerNV\n'):
...     print "PowerVM"
... else:
...     print "PowerNv"
... 
PowerNv
>>>

apahim:
Looks good. Thanks.

PR #573, opened 2017-07-19
Added new version of fio tarball link.

harihare:
Added link to latest version of fio tarball.

### DIFF
--- a/io/disk/fiotest.py.data/fio.yaml
+++ b/io/disk/fiotest.py.data/fio.yaml
@@ -2,7 +2,7 @@ parameters:
     disk: '/dev/sdd'  
     dir: '/mnt'
     fio_job: 'fio-simple.job'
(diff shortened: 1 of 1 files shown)


io/disk/fiotest.py.data/fio.yaml
```diff
@@ -2,7 +2,7 @@ parameters:
     disk: '/dev/sdd'  
     dir: '/mnt'
     fio_job: 'fio-simple.job'
-    fio_tool_url: 'http://brick.kernel.dk/snaps/fio-2.1.10.tar.gz'
+    fio_tool_url: 'http://brick.kernel.dk/snaps/fio-2.99.tar.gz'
```
PraveenPenguin: i think this change is not required as it is user input and can be modified on the fly what you say ?

PR #575, opened 2017-07-24
Patch to enhance Bonding script

hathyaga:
patch will replace all ifconfig related commands used
with ip commands and will also unload bonding module
while removing the bonding configuration in Host and
Peer machines


hathyaga:
Job log for Bonding interface created in Peer and tested all modes.

PR #576, opened 2017-07-25
Test to verify pstore - v2

Pavithra1602:
This script triggers crash on the machine configured in config.yaml and verifies files under /sys/fs/pstore.

### DIFF
--- a/generic/pstore.py
+++ b/generic/pstore.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2017 IBM
+# Author: Pavithra <pavrampu@linux.vnet.ibm.com>
+
+import time
+import os
+from avocado import Test
+from avocado import main
+from virttest import remote
+from avocado.utils import process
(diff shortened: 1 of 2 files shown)


generic/pstore.py
```diff
@@ -0,0 +1,75 @@
...
+        file_list = ['common-nvram', 'dmesg-nvram']
+        for files in file_list:
+            if files not in open(log_file).read():
+                self.fail("%s is not saved" % files)
+        process.run("echo "" > %s" % log_file, ignore_status=True, sudo=True, shell=True)
+        session2.run("cat /etc/os-release", 600, "True")
+        if "rhel" in open(log_file).read():
+            session2.run("yum install sos", 600, "True")
+        if "Ubuntu" in open(log_file).read():
+            session2.run("yum install sosreport", 600, "True")
```
PraveenPenguin: i think you need to use apt installer better to use  `SoftwareManager` for that

generic/pstore.py
```diff
@@ -0,0 +1,75 @@
...
+        session2 = remote.RemoteRunner("ssh", self.ip, 22, self.user_name, self.password,
+                                       self.prompt, "\n", log_file, 100, 10, None)
+        session2.run("ls -lrt /sys/fs/pstore", 100, "True")
+        file_list = ['common-nvram', 'dmesg-nvram']
+        for files in file_list:
+            if files not in open(log_file).read():
+                self.fail("%s is not saved" % files)
+        process.run("echo "" > %s" % log_file, ignore_status=True, sudo=True, shell=True)
+        session2.run("cat /etc/os-release", 600, "True")
+        if "rhel" in open(log_file).read():
```
PraveenPenguin: i think in fedora and centos also it work what you say ?

generic/pstore.py
PraveenPenguin: @Pavithra1602 thanks for helping on this as these command you want to run in target system (avocado not installed in remote machine ) may be we can have as that enhancement

PR #597, opened 2017-08-09
Don't overwrite properties

apahim:
Properties are not supposed to be overwritten, but since some of them
were still not protected, we ended up abusing them. Now, with the recent
PR https://github.com/avocado-framework/avocado/pull/2129 they are about
to be protected and this commit fixes the tests before they break. No
actual changes in functionality though.

Signed-off-by: Amador Pahim <apahim@redhat.com>


PraveenPenguin:

PR #598, opened 2017-08-09
Patch for htx_nic_devices.py script

hathyaga:
This patch has a fix which takes care of pingum failure
for Infiniband interfaces

### DIFF
--- a/io/net/htx_nic_devices.py.data/README.txt
+++ b/io/net/htx_nic_devices.py.data/README.txt
@@ -5,10 +5,21 @@ ready for the execution and also updates bpt file with necessary
 data. Once configuration is done test starts executing the stress
 test on all the interfaces provided in input yaml file
 
-User can mention the time limit in hours to execute.
+User can mention the time limit to be executed in hours.
 
 This test assumes below packages to be installed on both host & peer
 	pexpect
 	htx
 
 The net_id should be >= 100 and  <= 223
+
+Hostname for the host and the peer servers must always be set to the
+fully qualified domain name of the server IP. If not, set it using
+hostnamectl command
+
+If the test interfaces are the ones with Public IP's, they cannot be
+used to run this script as this script will change the IP's for them
+based on the net_id value. If this happens, none of the Peer commands
+will be executed as ssh will not happen and the test will fail.
+Make sure that there is a separate interface configured with Public IP
+and is left untouched.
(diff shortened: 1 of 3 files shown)


hathyaga:
Job log for HTX run on IB interfaces
[HTX_IB_interfaces_PASS.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1211526/HTX_IB_interfaces_PASS.txt)

hathyaga:
Job log for HTX run  on NIC interfaces
[HTX_NIC_interfaces_PASS.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1211530/HTX_NIC_interfaces_PASS.txt)

io/net/htx_nic_devices.py.data/README.txt
```diff
@@ -5,10 +5,14 @@ ready for the execution and also updates bpt file with necessary
 data. Once configuration is done test starts executing the stress
 test on all the interfaces provided in input yaml file
 
-User can mention the time limit in hours to execute.
+User can mention the time limit to be calculated in one tenth of an
+hour. For eg, value 1 is 60 minutes
```
narasimhan-v: I guess the previous one was clear enough ?

io/net/htx_nic_devices.py.data/htx_nic_devices.yaml
```diff
@@ -1,9 +1,9 @@
 mdt_file: 'net.mdt'
-host_ip: x.xx.xxx.xx
-peer_ip: x.xx.xxx.xy
+host_public_ip: x.xx.xxx.xx
```
narasimhan-v: What if the test interfaces are public interfaces ? Will the test run cleanly ?

io/net/htx_nic_devices.py.data/README.txt
hathyaga: Yes it was. That could have been not changed and left as is but just thought I'll add a more detailed description to what was existing.

io/net/htx_nic_devices.py.data/htx_nic_devices.yaml
hathyaga: That's a no because ssh happens through Public interfaces and that's how the commands on the peer gets executed. If we consider test interfaces with Public IP's, the script will change the IP for them due to net_id parameter in bpt file. This will fail ssh to the peer. 
So we'll have to make sure that while running HTX, we have a separate interface set aside which has a Public IP and that should be left untouched.

narasimhan-v: @hathyaga Sure, can we have this documented here.

hathyaga: Sure. I'll add it.

PraveenPenguin:
@hathyaga thanks for the patch , @narasimhan-v thanks for ACK

PR #599, opened 2017-08-10
Patch for enhancement of multiport test

hathyaga:
This patch will run a flood ping test parallely on the
test interfaces


io/net/multiport_stress.py
```diff
@@ -50,7 +50,7 @@ def setUp(self):
                 self.cancel("interface is not available")
         self.count = self.params.get("count", default="1000")
 
-    def test_multiport_stress(self):
+    def test_multiport_ping(self):
```
PraveenPenguin: i think earlier function function definition seems me more relevant

PR #607, opened 2017-08-18
Add with-nvmf to mofed install options

narasimhan-v:
Add with-nvmf to mofed install options.
This option tries to install mofed with nvme over fabrics support,
if the mofed version has support for it.


PraveenPenguin:
@narasimhan-v i think this option is user centric (Test choice ), can you please help me to understand as it will help for larger audience .may be please enhance commit message

narasimhan-v:
@PraveenPenguin Yes, it is user centric.

Users having latest mofed with nvmf, this will pass. Else, this case will fail.
It will have no effect in case of usability of larger audience, IMO.

PraveenPenguin:
@narasimhan-v thanks for helping , looks good

PR #610, opened 2017-08-24
Patch to inject EEH on Specified PE on PowerNV platform

vrbagalkote:
This sets max_eeh_freeze, and injects EEH error on specified PE
checks for recovery of the adapter.

### DIFF
--- a/io/pci/PowerNVEEH.py
+++ b/io/pci/PowerNVEEH.py
@@ -1,0 +1,215 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+#
+# Copyright: 2017 IBM
+# Author: Venkat Rao B <vrbagal1@linux.vnet.ibm.com>
+
+"""
+This scripts basic EEH tests on all PCI device
+"""
+
+import time
+from avocado import main
+from avocado import Test
+from avocado.utils import process
+from avocado.utils import pci
+from avocado.utils import genio
+
+EEH_HIT = 0
+EEH_MISS = 1
+
+
+class EEHRecoveryFailed(Exception):
+    """
+    Exception class, if EEH fails to recover
+    """
+    def __init__(self, thing, dev, log=None):
+        self.thing = thing
+        self.dev = dev
+        self.log = log
+
+    def __str__(self):
+        return "%s %s recovery failed: %s" % (self.thing, self.dev, self.log)
+
+
+class PowerNVEEH(Test):
+
+    """
+    This class contains functions for listing domains
+    forming EEH command
+    """
+    def setUp(self):
+
+        """
+        Gets the console and set-up the machine for test
+        """
+        output = genio.read_file("/sys/kernel/debug/powerpc/eeh_enable")\
+            .strip()
+        if output != '0x1':
+            self.cancel("EEH is not enabled, please enable via FSP")
+        self.max_freeze = int(self.params.get('max_freeze', default='1'))
+        cmd = "echo %d > /sys/kernel/debug/powerpc/eeh_max_freezes"\
+            % self.max_freeze
+        process.system(cmd, ignore_status=True, shell=True)
+        self.function = str(self.params.get('function'))
+        self.err = str(self.params.get('err'))
+        self.pci_device = str(self.params.get('pci_device', default=' '))
+        self.phb = self.pci_device.split(":", 1)[0]
+        self.addr = genio.read_file("/sys/bus/pci/devices/%s/"
+                                    "eeh_pe_config_addr" % self.pci_device)
+        self.addr = str(self.addr).rstrip()
+        self.log.info("Test-----> %s", self.addr)
+        self.log.info("===============Testing EEH Frozen PE==================")
+
+    def test_eeh_basic_pe(self):
+
+        """
+        Test to execute basic error injection on PE
+        """
+        enter_loop = True
+        num_of_miss = 0
+        num_of_hit = 0
+        while num_of_hit <= self.max_freeze:
+            for func in self.function:
+                self.log.info("Running error inject on pe %s function %s"
+                              % (self.pci_device, func))
+                if num_of_miss < 5:
+                    return_code = self.basic_eeh(self.addr, self.err, func,
+                                                 self.phb, self.pci_device)
+                    if return_code == EEH_MISS:
+                        num_of_miss += 1
+                        self.log.info("number of miss is %d"
+                                      % num_of_miss)
+                        continue
+                    else:
+                        num_of_hit += 1
+                        self.log.info("number of hit is %d"
+                                      % num_of_hit)
+                        if num_of_hit <= self.max_freeze:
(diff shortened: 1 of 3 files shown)


vrbagalkote:
[job.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1248761/job.txt)

Attached is the job.log file

io/pci/PowerNVEEH.py.data/PowerNVEEH.yaml
```diff
@@ -0,0 +1,6 @@
+scenario:
+            #max_freeze: 5
```
narasimhan-v: why all parameters are commented ?

io/pci/PowerNVEEH.py.data/PowerNVEEH.yaml
```diff
@@ -0,0 +1,6 @@
...
+            #function: 4
+            #err: 1
+            #pe: "0001:0c:00.0"
```
narasimhan-v: this can be `pci_device` as in other tests

io/pci/PowerNVEEH.py.data/PowerNVEEH.yaml
```diff
@@ -0,0 +1,6 @@
...
+            #phb: "0001"
```
narasimhan-v: can't this be computed from pe / pci_device ?

io/pci/PowerNVEEH.py
```diff
@@ -0,0 +1,218 @@
...
+        cmd = "dmesg -C"
+        process.system(cmd, ignore_status=True, shell=True)
+
+    @classmethod
+    def check_eeh_hit(cls):
+
+        """
+        Function to check if EEH is successfully hit
+        """
+        tries = 10
```
narasimhan-v: Isn't 10 sec too less time here ? Can't it be 30 sec at least ?
And also in PowerVMEEH.

io/pci/PowerNVEEH.py
```diff
@@ -0,0 +1,218 @@
...
+
+    @classmethod
+    def check_eeh_removed(cls):
+
+        """
+        Function to check if PE is recovered successfully
+        """
+        tries = 30
+        for _ in range(0, tries):
+            cmd = "(dmesg | grep 'permanently disabled'; echo $?)"
```
narasimhan-v: Why is `()` needed here ?

io/pci/PowerNVEEH.py.data/PowerNVEEH.yaml
vrbagalkote: I remember from one of previous comments, that this is better, to avoid undesirable effect if any tester runs as is. So all are commented.

io/pci/PowerNVEEH.py.data/PowerNVEEH.yaml
vrbagalkote: I think pe is right abbreviation here.

io/pci/PowerNVEEH.py
vrbagalkote: I didnt get u

io/pci/PowerNVEEH.py.data/PowerNVEEH.yaml
narasimhan-v: Correct. But that can be achieved if the values to these parameters are left empty.

```
max_freeze: 5
function: 4
err:
pe:
```

narasimhan-v: The problem with having these commented is, difficult to uncomment and change them automatically. Only changing can be done easily

io/pci/PowerNVEEH.py.data/PowerNVEEH.yaml
narasimhan-v: @vrbagalkote Agreed. But in other tests such as hotplug or driver bind, the parameter name is `pci_device`.

We want all names to be same, so that they can be changed automatically in one shot. So, either:
1. change this to `pci_device`
2. change in all other tests `pci_device` to `pe`

io/pci/PowerNVEEH.py
narasimhan-v: It can be `cmd = "dmesg | grep 'permanently disabled'; echo $?"`

io/pci/PowerNVEEH.py
```diff
@@ -0,0 +1,218 @@
...
+    """
+    def setUp(self):
+
+        """
+        Gets the console and set-up the machine for test
+        """
+        output = genio.read_file("/sys/kernel/debug/powerpc/eeh_enable")\
+            .strip()
+        if output != '0x1':
+            self.skip("EEH is not enabled, please enable via FSP")
```
PraveenPenguin: i think self.cancel is good option here as skip is going to deprecate soon

io/pci/PowerNVEEH.py
```diff
@@ -0,0 +1,218 @@
...
+            sys.exit(1)
```
PraveenPenguin: redundant logic self.cancel or self.skip does same thing  , please remove this

io/pci/PowerNVEEH.py
```diff
@@ -0,0 +1,218 @@
...
+        cmd = "echo %d > /sys/kernel/debug/powerpc/eeh_max_freezes"\
+            % self.max_freeze
+        process.system(cmd, ignore_status=True, shell=True)
+        self.function = str(self.params.get('function'))
+        self.err = str(self.params.get('err'))
+        self.pci_device = str(self.params.get('pci_device', default=' '))
+        self.phb = self.pci_device.split(":", 1)[0]
+        self.addr = genio.read_file("/sys/bus/pci/devices/%s/"
+                                    "eeh_pe_config_addr" % self.pci_device)
+        print type(self.addr)
```
PraveenPenguin: is this left over , if you want to print the value so i think please use `self.log.info `

io/pci/PowerNVEEH.py
```diff
@@ -0,0 +1,218 @@
...
+        self.addr = str(self.addr).rstrip()
+        print "Test----->", self.addr
```
PraveenPenguin: here as well

io/pci/PowerNVEEH.py
```diff
@@ -0,0 +1,218 @@
...
+    @classmethod
+    def error_inject(cls, addr, err, func, phb, pci_device):
+
+        """
+        Form a command to inject the error
+        """
+        cmd = "echo %s:%s:%s:0:0 > /sys/kernel/debug/powerpc/PCI%s/err_injct \
+               && lspci -ns %s" % (addr, err, func, phb, pci_device)
+        res = process.system_output(cmd, ignore_status=True, shell=True)
+        return int(res[-1])
```
PraveenPenguin: `return int(process.system_output(cmd, ignore_status=True, shell=True)[-1])  ` will be used just want to eliminate extra variable

io/pci/PowerNVEEH.py
```diff
@@ -0,0 +1,218 @@
...
+    def check_eeh_removed(cls):
+
+        """
+        Function to check if PE is recovered successfully
+        """
+        tries = 30
+        for _ in range(0, tries):
+            cmd = "dmesg | grep 'permanently disabled'; echo $?"
+            res = process.system_output(cmd, ignore_status=True, shell=True)
+            if int(res[-1]) == 0:
```
PraveenPenguin: here  also 
`
int( process.system_output(cmd, ignore_status=True, shell=True)[-1]) == 0:`

vrbagalkote:
[job.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1329147/job.txt)

Attached is the job.log file. @PraveenPenguin @narasimhan-v

PraveenPenguin:

PR #621, opened 2017-09-05
Added bpfcc/bcc test

harish-24:
Runs bcc self test. Added support only for Ubuntu.

### DIFF
--- a/toolchain/bcc.py
+++ b/toolchain/bcc.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2017 IBM
+# Author:Harish Sriram <harish@linux.vnet.ibm.com>
+#
+
+
+import os
+
+from avocado import Test
+from avocado import main
+from avocado.utils import archive, build, process, distro
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Bcc(Test):
+
+    """
+    Bcc is a compiler that understands traditional K&R C with just the
+    restriction that bit fields are mapped to one of the other integer types.
+    This is the self test of bcc package.
+    """
+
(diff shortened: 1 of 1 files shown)


generic/bcc.py
```diff
@@ -0,0 +1,76 @@
...
+                            'libedit-dev', 'libllvm3.8', 'llvm-3.8-dev',
+                            'libclang-3.8-dev', 'python', 'zlib1g-dev',
+                            'libelf-dev', 'clang-format-3.8', 'python-netaddr',
+                            'python-pyroute2', 'arping', 'iperf', 'netperf',
+                            'ethtool']:
+                if not smm.check_installed(package) and not smm.install(package):
+                    self.cancel("Failed to install %s, which is needed for"
+                                "the test to be run" % package)
+        else:
+            self.cancel("Test is yet to be supported for other distro")
```
PraveenPenguin: i think this decision will take on  starting of test as line of 43
can be like 
```
        if not distro.detect().name.lower() == 'ubuntu':
               self.cancel('Upsupported OS %s' % distro.detect().name.lower())
```

harish-24:
[bcc_test.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1276547/bcc_test.txt)

generic/bcc.py
```diff
@@ -0,0 +1,76 @@
...
+        os.makedirs('%s/build' % self.sourcedir)
+        self.builddir = '%s/build' % self.sourcedir
+        os.chdir(self.builddir)
+        process.run('cmake .. -DCMAKE_INSTALL_PREFIX=/usr', shell=True)
+
+        build.make(self.builddir)
+
+    def test(self):
+
+        if build.make(self.builddir, extra_args='test', ignore_status=True):
```
PraveenPenguin: are we need to parse debug file to make more pointed about failure

generic/bcc.py
```diff
@@ -0,0 +1,76 @@
...
+from avocado import Test
+from avocado import main
+from avocado.utils import archive, build, process, distro
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Bcc(Test):
+
+    """
+    Self test case of bcc
```
PraveenPenguin: can you please add some more info as what this test does

harish-24: Sure

generic/bcc.py
harish-24: sure, will make the change.

generic/bcc.py
harish-24: I think it is not required as it is selftest will point out the failures.

harish-24:
@PraveenPenguin Made the changes. Thanks for the review.

PR #626, opened 2017-09-12
Added numatop test

harish-24:
Test for numatop command using a self built mgen application. Currently supports only Ubuntu.

### DIFF
--- a/cpu/numatop.py
+++ b/cpu/numatop.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2017 IBM
+# Author: Harish Sriram <harish@linux.vnet.ibm.com>
+#
+
+
+import os
+
+from avocado import Test
+from avocado import main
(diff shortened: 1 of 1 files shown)


harish-24:
[numatop.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1295588/numatop.txt)

cpu/numatop.py
```diff
@@ -0,0 +1,86 @@
...
+        mgen = os.path.join(self.sourcedir, 'test/mgen/mgen')
+        self.numa_pid = process.SubProcess(
+            'numatop -d result_file', shell=True).start()
+
+        # Run mgen for 5 seconds to generate a single snapshot of numatop
+        process.run('%s -a 0 -c %s -t 5' %
+                    (mgen, cpu.cpu_online_list()[0]), shell=True, sudo=True)
+
+        # Kill numatop recording after running mgen
+        process.run('pkill -P %s' % self.numa_pid, ignore_status=True)
```
narasimhan-v: use `kill()` or `terminate()` from SubProcess Class.
https://github.com/avocado-framework/avocado/blob/master/avocado/utils/process.py#L555

cpu/numatop.py
```diff
@@ -0,0 +1,86 @@
...
+        process.run('kill -9 %s' % self.numa_pid, ignore_status=True)
+
+        # Analyse record file for mgen record
+        result = process.system_output(
+            'grep -A1 PID %s/result_file' % self.sourcedir,
```
narasimhan-v: It would be better to process the result_file's contents in python instead of shell by grepping.

cpu/numatop.py
harish-24: Thanks for the suggestion. Will change it accordingly.

cpu/numatop.py
harish-24: Agreed.

harish-24:
@narasimhan-v Updated with the changes. Thanks.

PR #628, opened 2017-09-14
Enhancement in port_bounce.py with multipath checks

Naresh-ibm:
when ports bounce is done we are checking the respective multipath
status, both when disable and enable of switch ports

Signed-off-by: Naresh Bannoth <nbannoth@linux.vnet.ibm.com>


Naresh-ibm:
[port_bounce_job_log_1.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1301949/port_bounce_job_log_1.txt)

io/disk/port_bounce.py
```diff
@@ -191,23 +199,49 @@ def verify_port_enable(self, test_ports):
                 msg = "Port %s is failed to enable" % port
                 self.failure_list[port] = msg
 
-    def verify_port_toggle_host(self, status):
```
narasimhan-v: Any reason this is removed ?

io/disk/port_bounce.py
```diff
@@ -146,25 +152,27 @@ def porttoggle(self, test_ports, sleep_time):
         except CommandFailed as cf:
             self.log.info("port disable failed for port(s) %s, details: %s",
                           test_ports, str(cf))
-        time.sleep(sleep_time)
```
narasimhan-v: Why not use this sleep time everywhere ?

io/disk/port_bounce.py
```diff
@@ -191,23 +199,49 @@ def verify_port_enable(self, test_ports):
...
+        err_paths = []
+        for wwid in wwids:
+            if wwid not in process.system_output('multipath -ll',
+                                                 ignore_status=True,
+                                                 shell=True):
+                continue
+            mpaths = multipath.get_paths(wwid)
+            for path in mpaths:
+                path_stat = multipath.get_path_status(path)
+                if path_stat[0] == 'failed' or path_stat[2] == 'faulty':
```
narasimhan-v: Why checking 'failed' and 'faulty' in both functions ?

io/disk/port_bounce.py
Naresh-ibm: @narasimhan-v this sleep time is defer for different scenarios.

narasimhan-v: @Naresh-ibm having different sleep times everywhere is not ideal.

1. We should try to eliminate the use of sleep. probably you can explore the wait_for in utils
2. If 1 does not work good for a situation, have sleep time but uniform everywhere.
3. If 2 also does not work good for a situation, specify why a different sleep time is needed. I am against 3, since it would be different for different infrastructure (ex: different OS, different number of multipath devices, etc)

Naresh-ibm: @narasimhan-v  this sleep time is given intentionally for port bounce, (for short port bounce and for long port bounce), which will be in minutes and hence we cannot use same variable as other sleep time. hope you got it.

io/disk/port_bounce.py
Naresh-ibm: this was mistake while editing the file will add it again. :-)

io/disk/port_bounce.py
Naresh-ibm: each has some different usage in the if. So this is also needed  in both the places.

io/disk/port_bounce.py
narasimhan-v: Ok, Sorry. I got confused with port bounce time with sleep time.
Can you make all the other sleep times uniform and use it via a variable please

io/disk/port_bounce.py
narasimhan-v: Need to check the same thing in both `mpath_pass_check()` and `mpath_fail_check()` ? Please clarify

io/disk/port_bounce.py
```diff
@@ -191,23 +199,49 @@ def verify_port_enable(self, test_ports):
...
-                          actual_state: %s", status, state)
+    def mpath_fail_check(self, test_ports):
+        '''
+        checking mpath disk status after disabling the switch port
+        '''
+        total_mpaths = 0
+        count = 0
+        wwids = multipath.get_multipath_wwids()
+        for wwid in wwids:
+            if wwid not in process.system_output('multipath -ll',
```
vrbagalkote: @Naresh-ibm I think its better to run command outside of loop and just compare it in the loop.

io/disk/port_bounce.py
```diff
@@ -191,23 +199,49 @@ def verify_port_enable(self, test_ports):
...
-                              bus_id, state)
-            else:
-                self.fail("port state not changed in host expected state: %s, \
-                          actual_state: %s", status, state)
+    def mpath_fail_check(self, test_ports):
+        '''
+        checking mpath disk status after disabling the switch port
+        '''
+        total_mpaths = 0
+        count = 0
```
vrbagalkote: Any reason why there variables are declared?

io/disk/port_bounce.py
```diff
@@ -191,23 +199,49 @@ def verify_port_enable(self, test_ports):
...
+                                                 ignore_status=True,
+                                                 shell=True):
+                continue
+            mpaths = multipath.get_paths(wwid)
+            total_mpaths += len(mpaths)
+            for path in mpaths:
+                path_stat = multipath.get_path_status(path)
+                if path_stat[0] == 'failed' or path_stat[2] == 'faulty':
+                    count += 1
+        self.log.info("total mpaths = %s", total_mpaths)
```
vrbagalkote: there is no need of variable total_mpaths

io/disk/port_bounce.py
```diff
@@ -191,23 +199,49 @@ def verify_port_enable(self, test_ports):
...
+            self.error("all paths did not failed on port disable")
+
+    def mpath_pass_check(self):
+        '''
+        checking mpath disk status after enabling the switch port
+        '''
+        wwids = multipath.get_multipath_wwids()
+        err_paths = []
+        for wwid in wwids:
+            if wwid not in process.system_output('multipath -ll',
```
vrbagalkote: same here

io/disk/port_bounce.py
Naresh-ibm: @vrbagalkote to make sure,no junk values are stored in them.

io/disk/port_bounce.py
vrbagalkote: better to have a method to check wwid

Naresh-ibm: closing this in favour of #680

Naresh-ibm:
closing this in favour of #680

PR #636, opened 2017-09-18
Added new services.

Pavithra1602:
Added opal_errd, opal_prd and rtas_errd.


PraveenPenguin:
@Pavithra1602 i think we can check  all the service available in the system , any way merging this , will take that as enhancement

PR #639, opened 2017-09-21
Removed multipath flush from teardown

narasimhan-v:
Multipath test was flushing all the paths in teardown.
This will cause all multipath devices to stop, which is not
expected. The system state should be retained after the test

### DIFF
--- a/io/disk/multipath_test.py
+++ b/io/disk/multipath_test.py
@@ -140,7 +140,6 @@ class MultipathTest(Test):
 
         # Need to wait for some time to make sure multipaths are loaded.
         time.sleep(5)
-        process.run('multipath -F')
 
 
 if __name__ == "__main__":


io/disk/multipath_test.py
```diff
@@ -140,7 +140,6 @@ def tearDown(self):
 
         # Need to wait for some time to make sure multipaths are loaded.
         time.sleep(5)
```
PraveenPenguin: is this sleep can also be removed as we are not flushing multipath device

narasimhan-v: @PraveenPenguin No, this sleep is needed still. since we need to wait for multipath to be stable before test exits.

PraveenPenguin: aah ok , thanks

PR #644, opened 2017-09-26
adding 6 yaml and 6 job files for KVM CI perf tests

Venutiwa:
multiple files are needed as the requirement is to have profilers running for each test individually and not only for the entire job.

Signed-off-by: Venu <venutiwa@in.ibm.com>

### DIFF
--- a/io/disk/fiotest.py.data/fio_rand_RW.yaml
+++ b/io/disk/fiotest.py.data/fio_rand_RW.yaml
@@ -1,0 +1,9 @@
+fio_randRW.yaml:
+parameters:
+    disk: '/dev/sdd'
+    dir: '/mnt'
+    fio_job: 'fio-rand-RW.job'
+    fio_tool_url: 'http://brick.kernel.dk/snaps/fio-2.99.tar.gz'
+filesystem: !mux
+    ext4:
+        fs: 'ext4'
--- a/io/disk/fiotest.py.data/fio_rand_RW.yaml
+++ b/io/disk/fiotest.py.data/fio_rand_RW.yaml
@@ -1,6 +1,5 @@
 fio_randRW.yaml:
 parameters:
-    disk: '/dev/sdd'
     dir: '/mnt'
     fio_job: 'fio-rand-RW.job'
     fio_tool_url: 'http://brick.kernel.dk/snaps/fio-2.99.tar.gz'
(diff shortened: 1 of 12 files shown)


io/disk/fiotest.py.data/fio_rand_RW.yaml
```diff
@@ -0,0 +1,9 @@
+fio_randRW.yaml:
+parameters:
+    disk: '/dev/sdd'
```
narasimhan-v: I would prefer to have value of `disk` empty, since when run unintentionally, test should do no harm to the disk.

PraveenPenguin:
@Venutiwa @narasimhan-v   is there is not other way to have a single yaml file as this approach not looks good , also one more point can these job file sent to fio source code , please suggest

narasimhan-v:
@Venutiwa @PraveenPenguin Sure, that can be tried definitely.
These job profiles can be sent to https://github.com/axboe/fio/tree/master/examples

I think multiple yaml files is because of https://github.com/avocado-framework/avocado/issues/2169

Venutiwa:
@PraveenPenguin  as @narasimhan-v  mentioned, multiple yaml files are because of avocado-framework/avocado#2169 . If you remember, we had a discussion about this. 
About sending the job files to https://github.com/axboe/fio/tree/master/examples , Sure will look into it.
@narasimhan-v regarding the disk value, I did not make any changes in that parameter from the original yaml file that was already existing in the avocado-misc-tests.

narasimhan-v:
@Venutiwa I understand that. We actually need to be changing the default values in yaml everywhere, since we keep on evolving with best practises

PraveenPenguin:
@Venutiwa ah sure i miss that context

io/disk/fiotest.py.data/fio_rand_read.yaml
```diff
@@ -0,0 +1,9 @@
+fio_randread.yaml:
+parameters:
+    disk: '/dev/sdd'
```
sathnaga: same as above and rest of places, a readme file explaining the values or a comment here should help

sathnaga:
@Venutiwa Commit msg can be changed something like
```
Add support for multiple jobs for fio test

<Explain about the jobs in commit msg>
```
would be good for later read

PraveenPenguin:
@Venutiwa can you please wrap in single commit , and any luck to send job file to fio source ?

Venutiwa:
@PraveenPenguin It was not accepted in the fio source, so continuing here https://github.com/axboe/fio/pull/476

Venutiwa:
new PR has been raised with single commit: https://github.com/avocado-framework-tests/avocado-misc-tests/pull/653

PR #645, opened 2017-09-26
Fix ppc test

harish-24:
Use avocado library function for processor check and Handle SMT mix state in ppc smt test.

### DIFF
--- a/cpu/ppc64_cpu_test.py
+++ b/cpu/ppc64_cpu_test.py
@@ -43,8 +43,12 @@ class PPC64Test(Test):
         if SoftwareManager().check_installed("powerpc-utils") is False:
             if SoftwareManager().install("powerpc-utils") is False:
                 self.cancel("powerpc-utils is not installing")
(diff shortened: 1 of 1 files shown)


cpu/ppc64_cpu_test.py
```diff
@@ -37,14 +38,18 @@ def setUp(self):
         """
         Verifies if powerpc-utils is installed, and gets current SMT value.
         """
-        command = "uname -p"
-        if 'ppc' not in process.system_output(command, ignore_status=True):
+        dist = distro.detect()
+        if 'ppc' not in dist.arch:
```
PraveenPenguin: can it be like 

if 'ppc' not in distro.detect().arch

harish-24:
@PraveenPenguin Thanks. Pushed.

cpu/ppc64_cpu_test.py
harish-24: sure

PraveenPenguin:
@sathnaga thanks for ACK

PR #649, opened 2017-10-03
Fix compiling error Valgrind test (ppc64el)

rrosatti:
There is a compiling error when running valgrind testcase on ppc64el.

m_debuglog.c: In function ‘local_sys_write_stderr’:
m_debuglog.c:203:4: error: PIC register clobbered by ‘r2’ in ‘asm’
    __asm__ volatile (
    ^~~~~~~
m_debuglog.c: In function ‘vgPlain_debugLog’:
m_debuglog.c:228:4: error: PIC register clobbered by ‘r2’ in ‘asm’
    __asm__ volatile (
    ^~~~~~~

In the upstream version (v3.13) it's already fixed. So, by just using
the upstream version instead of v3.12, this error does not happen.

Signed-off-by: Rodrigo R. Galvao <rosattig@br.ibm.com>

### DIFF
--- a/toolchain/valgrind.py
+++ b/toolchain/valgrind.py
@@ -48,7 +48,7 @@ class Valgrind(Test):
             if not smm.check_installed(package) and not smm.install(package):
                 self.cancel('%s is needed for the test to be run' % package)
         tarball = self.fetch_asset(
-            "http://valgrind.org/downloads/valgrind-3.12.0.tar.bz2")
+                "ftp://sourceware.org/pub/valgrind/valgrind-3.13.0.tar.bz2")
         archive.extract(tarball, self.srcdir)
         version = os.path.basename(tarball.split('.tar.')[0])
         self.sourcedir = os.path.join(self.srcdir, version)


rrosatti:
The upstream commit that fix this error in the Valgrind package is this one -> https://sourceware.org/git/?p=valgrind.git;a=commit;h=41308849efb1f35816d409855314184e852c742d

rrosatti:
@PraveenPenguin you're welcome!!

toolchain/valgrind.py
```diff
@@ -48,7 +48,7 @@ def setUp(self):
             if not smm.check_installed(package) and not smm.install(package):
                 self.cancel('%s is needed for the test to be run' % package)
         tarball = self.fetch_asset(
-            "http://valgrind.org/downloads/valgrind-3.12.0.tar.bz2")
+                "ftp://sourceware.org/pub/valgrind/valgrind-3.13.0.tar.bz2")
```
apahim: It would be nice if we had a lib to probe the latest version of a given asset in a given location so we could call `fetch_asset()` always with the latest version.

PraveenPenguin: @apahim seems me good idea  , i will send a patch for it

apahim: @PraveenPenguin, the new `avocado.utils.vmimage` library has something like that to get the latest distro and image versions. Feel free to use it as reference ;)

PraveenPenguin: @apahim sure , thanks

PR #653, opened 2017-10-09
adding support for multiple fio jobs for KVM perf CI

Venutiwa:
multiple yaml files are needed because the requirement is to have profilers run individually for each test

Signed-off-by: Venu <venutiwa@in.ibm.com>

With reference to https://github.com/avocado-framework-tests/avocado-misc-tests/pull/644: as per the comment on this PR, adding the changes in single commit.

### DIFF
--- a/io/disk/fiotest.py.data/fio-rand-RW.job
+++ b/io/disk/fiotest.py.data/fio-rand-RW.job
@@ -1,0 +1,18 @@
+; fio-rand-RW.job for fiotest
+
+[global]
+name=fio-rand-RW
+filename=fio-rand-RW
+rw=randrw
+rwmixread=60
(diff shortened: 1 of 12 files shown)


PraveenPenguin:
@narasimhan-v what you think about this PR please suggest

narasimhan-v:
@PraveenPenguin The only enhancement possible here is to eliminate multiple yaml files. But since the need here is separate profilers, I think we can merge this one. 

@Venutiwa https://github.com/avocado-framework/avocado/issues/2169 needs to be followed and if fixed, this commit has to be fixed here as well.

PraveenPenguin:
@narasimhan-v thanks

PR #655, opened 2017-10-10
Fix numatop test

harish-24:
mgen application does not necessarily to be on first line. Patch checks for its existence in the file rather than the first line.

### DIFF
--- a/cpu/numatop.py
+++ b/cpu/numatop.py
@@ -64,6 +64,7 @@ class Numatop(Test):
 
     def test(self):
 
+        mgen_flag = False
         mgen = os.path.join(self.sourcedir, 'test/mgen/mgen')
         self.numa_pid = process.SubProcess(
             'numatop -d result_file', shell=True)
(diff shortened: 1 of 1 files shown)


cpu/numatop.py
```diff
@@ -79,11 +80,12 @@ def test(self):
         # Analyse record file for mgen record
         with open('%s/result_file' % self.sourcedir, 'r') as f_read:
             lines = f_read.readlines()
-            if 'PID' in lines[2]:
-                if 'mgen' not in lines[3]:
-                    self.fail(
-                        'Numatop failed to record mgen latency. Please check '
-                        'the record file: %s/result_file' % self.sourcedir)
+            for line in lines:
+                if 'mgen' in line.split():
```
narasimhan-v: is `split()` needed here ?

harish-24: not necessarily, but thought it looks better to have check for the application. than the entire string.

harish-24: @narasimhan-v right. I think it does not make sense to split at all. I ll make change as you have suggested.

harish-24:
@narasimhan-v Done. Thanks.

PR #656, opened 2017-10-11
Test to verify packages to be installed by default - v3

Pavithra1602:
This test checks whether powerpc-utils, ppc64-diag and lsvpd are installed by default. Additional packages can be mentioned in yaml file.

### DIFF
--- a/generic/packages.py
+++ b/generic/packages.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2017 IBM
+# Author: Pavithra <pavrampu@linux.vnet.ibm.com>
+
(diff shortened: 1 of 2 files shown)


generic/packages.py
```diff
@@ -0,0 +1,65 @@
...
+        self.sm = SoftwareManager()
+        self.packages = self.params.get('packages', default=['powerpc-utils', 'ppc64-diag', 'lsvpd'])
+        if 'PowerNV' in open('/proc/cpuinfo', 'r').read():
+            self.packages.extend(['opal-prd'])
+
+    def test(self):
+        dist = distro.detect()
+        is_fail = 0
+        if dist.name == 'rhel':
+            self.packages_rhel = self.params.get('packages_rhel', default=['lshw', 'librtas'])
```
PraveenPenguin: this can be local variable as i did not see it used outside of test subroutine

generic/packages.py
```diff
@@ -0,0 +1,65 @@
...
+            self.packages.extend(self.packages_rhel)
+            for package in self.packages:
+                if "anaconda" in process.system_output("yum list installed | "
+                                                       "grep %s | tail -1" % package, shell=True):
+                    self.log.info("%s package is installed by default" % package)
+                else:
+                    self.log.info("%s package is not installed by default" % package)
+                    is_fail += 1
+        elif dist.name == 'Ubuntu':
+            self.packages_ubuntu = self.params.get('packages_ubuntu', default=['librtas2'])
```
PraveenPenguin: this as well

PraveenPenguin:
@Pavithra1602 thanks for the patch , only very minor issue like pylint and two variable can be in local scope , i already taken care those while merging

PR #661, opened 2017-10-15
memhotplug: hot unplug for free memory available

abdhaleegit:
The script unplugs all the memory making the machine go hung,
So make sure we use only portion of free memory by passing memcount

if memory count is not specified whole free memory is being hotplugged

### DIFF
--- a/memory/memhotplug.py
+++ b/memory/memhotplug.py
@@ -38,6 +38,15 @@ errorlog = ['WARNING: CPU:', 'Oops',
             'double fault:', 'BUG: Bad page state in']
 
 
+def get_mem_count():
+    with open('/sys/devices/system/memory/block_size_bytes', 'r') as block_file:
+        blocksize = int(block_file.read()) / 10
+    with open('/sys/devices/system/node/has_normal_memory', 'r') as node_file:
+        nodes = node_file.read()
+    node_mem = (memory.meminfo.MemFree.mb) / len(nodes.split('-'))
+    return ((node_mem * 1024) / blocksize)
+
+
 def online(block):
     try:
         memory.hotplug(block)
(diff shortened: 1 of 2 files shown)


memory/memhotplug.py
```diff
@@ -38,6 +38,15 @@
             'double fault:', 'BUG: Bad page state in']
 
 
+def get_mem_count():
+    with open('/sys/devices/system/memory/block_size_bytes', 'r') as block_file:
+        blocksize = int(block_file.read()) / 10
```
PraveenPenguin: can you please help me to understand by only divide by 10 as here converting hex value to int than divide by 10 , can we right sift the value which is appropriate (divide by 2)   than assign to blocksize

memory/memhotplug.py
```diff
@@ -38,6 +38,15 @@
...
+    with open('/sys/devices/system/node/has_normal_memory', 'r') as node_file:
+        nodes = node_file.read()
+    node_mem = (memory.meminfo.MemFree.mb) / len(nodes.split('-'))
```
PraveenPenguin: in my system i get `has_normal_memory ` value is `0,8`  so this split logic not works for me can you please more elaborate why it needed

memory/memhotplug.py
```diff
@@ -97,9 +106,16 @@ def setUp(self):
         self.stresstime = int(self.params.get('stresstime', default='10'))
         self.vmcount = int(self.params.get('vmcount', default='4'))
         self.iocount = int(self.params.get('iocount', default='4'))
+        self.hotplug_count = self.params.get('memcount', default='3')
```
PraveenPenguin: i think `self.hotplug_count` can be only zero or none zero value as i did not see any effect in code as it has any numeric value

abdhaleegit:
Here is the summary for the above comments:

_1. why divide by 10 at  blocksize = int(block_file.read()) / 10_
The wrong assumption I had while calculating:

Each memory block is of size 10000000 bytes
i.e # cat /sys/devices/system/memory/block_size_bytes
10000000
convert it to KB i.e blcoksize = 10000000 / 1024 = 9765 KB
** here I did divide by 10 instead of 1024 ^^

So to get total blocks:
Total free memory is in KB: MemFree:        63606464 kB
And the machine have 2 nodes. each node takes : MemFree:        31905216 kB
To get the total memory blocks on each node = 31905216 kB / 9765 KB = 3275 blocks

3275 blocks are too many compare to 255 memory blocks in sysfs at /sys/devices/system/memory/
i.e 128  blocks for each node.

Out of 255 memory blocks not all of them are hotpluggable. so only way to check is the removable parameter.

So the memcount calculation is unnecessary here because always there are 255 memory blocks at any time, instead I will let user to pass self.hotplug_count as number of memory blocks to remove (0 to 255)

The code will hotplug/unplug based on the given data.

2. has_normal_memory has 0,8 in your system : I will take care of that in my new code

A new patch will followup.

Thanks

PraveenPenguin:
@abdhaleegit thanks for explanation , waiting for V2 :)

PraveenPenguin:
@abdhaleegit any update on this PR , if this PR not pursue  any more can we close this

PR #667, opened 2017-10-24
Use dd for loop device creation on ext3 root fs and change default disk device

harish-24:
xfstests checks for 15GB of availabe space but default sizes are more than that and hence test fails to create loop devices. So default sizes are reduced.
As fallocate is not supported on ext3 root file system, it is replaced with dd command for loop device creation for ext3 fs.

### DIFF
--- a/fs/xfstests.py
+++ b/fs/xfstests.py
@@ -45,8 +45,9 @@ class Xfstests(Test):
         Build xfstest
         Source: git://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git
         """
+        self.use_dd = False
         if process.system_output("df -T / | awk 'END {print $2}'", shell=True) == 'ext3':
-            self.cancel('Test does not support ext3 root file system')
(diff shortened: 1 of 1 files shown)


fs/xfstests.py
```diff
@@ -218,8 +219,13 @@ def _create_loop_device(self, base_disk, loop_size, mount=True):
             self.part.mount()
         # Creating two loop devices
         for i in range(2):
-            process.run('fallocate -o 0 -l %s %s/file-%s.img' %
-                        (loop_size, self.disk_mnt, i), shell=True, sudo=True)
+            if self.use_dd:
+                dd_count = int(loop_size.split('GiB')[0]) * 1024
```
narasimhan-v: bs can be made 1G instead

harish-24: sure

narasimhan-v:
@harish-24 looks good.
@harihare Thanks for the review.

PR #668, opened 2017-10-24
Fix kernbench test

harish-24:
URL failed to fetch with location of fetch_asset provided as a string rather than a list. This patch fixes it.

### DIFF
--- a/kernel/kernbench.py
+++ b/kernel/kernbench.py
@@ -118,7 +118,7 @@ class Kernbench(Test):
         """
         # Setting the kernel
         # Uncompress the kernel archive to the work directory
-        tarball = self.fetch_asset("kernbench.zip", locations=self.location,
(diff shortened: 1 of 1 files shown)


kernel/kernbench.py
```diff
@@ -118,7 +118,7 @@ def test(self):
         """
         # Setting the kernel
         # Uncompress the kernel archive to the work directory
-        tarball = self.fetch_asset("kernbench.zip", locations=self.location,
+        tarball = self.fetch_asset("kernbench.zip", locations=[self.location],
```
narasimhan-v: @harish-24 looks good, but would extending the expiry also be good ?

harish-24: It can be increased, but the idea was to run daily so that latest kernel would be fetched everyday.

narasimhan-v: Oh ok

PR #670, opened 2017-10-26
Script to add Uperf which is a network performance benchmark

hathyaga:
tool

This script will calculate the performance of the Network
adapter

### DIFF
--- a/io/net/uperf_test.py
+++ b/io/net/uperf_test.py
@@ -1,0 +1,127 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2017 IBM
+# Author: Harsha Thygaraaja <harshkid@linux.vnet.ibm.com>
+
+"""
+Unified Performance Tool or uperf for short, is a network
+performance measurement tool that supports execution of
+workload profiles
+"""
+
+import os
+import netifaces
+from avocado import main
+from avocado import Test
+from avocado.utils.software_manager import SoftwareManager
+from avocado.utils import distro
+from avocado.utils import build
+from avocado.utils import archive
+from avocado.utils import process
+from avocado.utils.genio import read_file
+
+
+class Uperf(Test):
+    """
+    Uperf Test
+    """
+
+    def setUp(self):
+        """
+        To check and install dependencies for the test
+        """
+        smm = SoftwareManager()
+        detected_distro = distro.detect()
+        pkgs = ["gcc", "autoconf", "perl", "m4", "git-core"]
+        if detected_distro.name == "Ubuntu":
+            pkgs.extend(["libsctp1", "libsctp-dev", "lksctp-tools"])
+        else:
+            pkgs.extend(["lksctp-tools", "lksctp-tools-devel"])
+        for pkg in pkgs:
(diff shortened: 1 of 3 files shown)


vrbagalkote:
@hathyaga Remove wait from import statement and resend the patch.

hathyaga:
@vrbagalkote Yup. I already knew that was causing the check to fail. I only need to re-send it again.

io/net/uperf_test.py
```diff
@@ -0,0 +1,129 @@
...
+        result = process.run(cmd, shell=True, ignore_status=True)
+        if result.exit_status:
+            self.fail("FAIL: Uperf Run failed")
+        for line in result.stdout.splitlines():
+            if self.peer_ip in line:
+                if 'Mb/s' in line:
+                    tput = int(line.split()[3].split('.')[0])
+                else:
+                    # Converting the throughput calculated in Gb to Mb
+                    tput = int(line.split()[3].split('.')[0]) * 1000
```
vrbagalkote: what if through put returned in some other metrics, other than Mbps and Gbps

io/net/uperf_test.py
```diff
@@ -0,0 +1,129 @@
...
+                                         "//github.com/uperf/uperf/"
+                                         "archive/master.zip")
+        tarball = self.fetch_asset(uperf_download, expire='7d')
+        archive.extract(tarball, self.teststmpdir)
+        self.uperf_dir = os.path.join(self.teststmpdir, "uperf-master")
+        cmd = "scp -r %s %s@%s:/tmp" % (self.uperf_dir, self.peer_user,
+                                        self.peer_ip)
+        if process.system(cmd, shell=True, ignore_status=True) != 0:
+            self.cancel("unable to copy the uperf into peer machine")
+        tmp = "cd /tmp/uperf-master;./configure ppc64le;make"
```
vrbagalkote: I think no need to use tmp here, you can use cmd only.

io/net/uperf_test.py
hathyaga: Ethtool and sysfs prints the throughput in MBps. 
Considering throughput for a network card will be in MB or GB, the uperf will also calculate performance in those metrics.  I have no other metric to be considered. 

Also, I've seen a crad operating in KB speed.

io/net/uperf_test.py
```diff
@@ -0,0 +1,129 @@
...
+        transmitting (or receiving) data from a client. This transmit large
+        messages using multiple threads or processes.
+        """
+        if self.uperf_run == '1':
+            tmp = "cd /tmp/uperf-master/src;./uperf -s"
+            cmd = "ssh %s@%s \"%s\"" % (self.peer_user, self.peer_ip, tmp)
+            obj = process.SubProcess(cmd, verbose=False, shell=True)
+            obj.start()
+        speed = int(read_file("/sys/class/net/%s/speed" % self.iface))
+        self.expected_tp = int(self.expected_tp) * speed / 100
```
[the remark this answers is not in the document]
hathyaga: This can be considered.

io/net/uperf_test.py
hathyaga: This is purely cosmetic. Rather than making the command line seem longer, I'd rather use variable "tmp".

narasimhan-v: @hathyaga What @vrbagalkote meant was, use `cmd` instead of `tmp`

io/net/uperf_test.py
```diff
@@ -0,0 +1,129 @@
...
+        os.chdir(self.uperf)
+        time.sleep(1)
```
narasimhan-v: Do add a comment here for the sleep

hathyaga:
@narasimhan-v All changes suggested is taken care of. Please check.  Attached is the job log. 
[Uperf_job_log.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1421868/Uperf_job_log.txt)

PraveenPenguin:
@hathyaga thanks for the patch , merging as already one ACK 
@narasimhan-v thanks for ACK

PR #671, opened 2017-10-26
Restoring blacklisted multipaths

narasimhan-v:
After blacklisting multipaths via conf file, they were not restored.
This would cause them to be blacklisted for operations after that.
Fixed it by restoring it immediately after blacklisting.


narasimhan-v:
Closing since this is already taken care in #672

PR #681, opened 2017-11-03
Fixed checking for ERROR in dmesg

harihare:
the code was spliting the dmesg due to which multi-word match will fail. This
was a critical issue were we may mix the bugs. the same is fixed with this patch.

Signed-off-by: Hariharan T.S <hari@linux.vnet.ibm.com>

### DIFF
--- a/cpu/cpustress.py
+++ b/cpu/cpustress.py
@@ -82,14 +82,13 @@ class cpustresstest(Test):
 
     @staticmethod
     def __error_check():
-        ERROR = None
-        logs = (process.system_output(
-            "dmesg -T --level=alert,crit,err,warn", ignore_status=True, shell=True, sudo=True)).split()
-        for log in logs:
-            for error in errorlog:
+        ERROR = []
+        logs = process.system_output("dmesg -Txl 1,2,3,4").splitlines()
+        for error in errorlog:
+            for log in logs:
                 if error in log:
-                    ERROR = 'Test resulted %s in syslogs"' % error
-        return ERROR
+                    ERROR.append(log)
+        return "\n".join(ERROR)
 
     @staticmethod
     def __isSMT():
--- a/generic/stress-ng.py
+++ b/generic/stress-ng.py
@@ -122,12 +122,16 @@ class stressng(Test):
         cmd = 'stress-ng %s' % " ".join(args)
         process.run(cmd, ignore_status=True, sudo=True)
         collect_dmesg(self)
+        ERROR = []
         pattern = ['WARNING: CPU:', 'Oops',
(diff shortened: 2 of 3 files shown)


cpu/cpustress.py
```diff
@@ -83,12 +83,10 @@ def __clear_dmesg():
     @staticmethod
     def __error_check():
         ERROR = None
-        logs = (process.system_output(
-            "dmesg -T --level=alert,crit,err,warn", ignore_status=True, shell=True, sudo=True)).split()
-        for log in logs:
-            for error in errorlog:
-                if error in log:
-                    ERROR = 'Test resulted %s in syslogs"' % error
+        logs = process.system_output("dmesg -T --level=alert,crit,err,warn")
```
narasimhan-v: Suggestion: `dmesg -Txl 1,2,3,4`

PraveenPenguin:
@abdhaleegit please review this

cpu/cpustress.py
harihare: I don't see any value add, Please let me know if any. 

However both the command gives same output, Hence I have modified it. Still I am interested to know how different is this from the command with actual names.

narasimhan-v: @harihare There is no difference, except for the addition of `x`, which specifies the level.

cpu/cpustress.py
```diff
@@ -83,12 +83,10 @@ def __clear_dmesg():
...
-        logs = (process.system_output(
-            "dmesg -T --level=alert,crit,err,warn", ignore_status=True, shell=True, sudo=True)).split()
-        for log in logs:
-            for error in errorlog:
-                if error in log:
-                    ERROR = 'Test resulted %s in syslogs"' % error
+        logs = process.system_output("dmesg -Txl 1,2,3,4")
+        for error in errorlog:
+            if error in logs:
+                ERROR = 'Test resulted %s in syslogs"' % error
```
narasimhan-v: @harihare We should ideally be appending the error logs to `ERROR`.

Also, I think it makes more sense to split the logs using `splitlines()` and check for errors, and keep on appending such lines to `ERROR` and finally display them. Please let me know your take on this.
Something like this:
```
ERROR = []
logs = process.system_output("dmesg -Txl 1,2,3,4").splitlines()
for error in errorlog:
    for log in logs:
        if error in log:
            ERROR.append(log)
self.log.debug("\n".join(ERROR))
```

harihare: I agree with you, I have made the changes accordingly.

PraveenPenguin:
@harihare thanks for patch , @abdhaleegit @narasimhan-v Thanks for Review

PR #682, opened 2017-11-06
Added VA test

PraveenPenguin:
added VA test which test virtual address limit on  linux

### DIFF
--- a/memory/vatest.py
+++ b/memory/vatest.py
@@ -1,0 +1,87 @@
+#!/usr/bin/env python
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2017 IBM
+# Author:Praveen K Pandey<praveen@linux.vnet.ibm.com>
+#
+
+import os
+import shutil
+
+from avocado import Test
+from avocado import main
+from avocado.utils import process,  build, memory
+from avocado.utils.software_manager import SoftwareManager
+
+
+class VATest(Test):
+    """
+    Performs Virtual address space validation
+
+    :avocado: tags=memory, power
+    """
+
+    def setUp(self):
+        '''
+        Build VA Test
+        '''
+
+        # Check for basic utilities
+        smm = SoftwareManager()
+        for packages in ['gcc', 'make']:
+            if not smm.check_installed(packages) and not smm.install(packages):
+                self.cancle('%s is needed for the test to be run' % packages)
+
+        data_dir = os.path.abspath(self.datadir)
+
+        shutil.copyfile(os.path.join(data_dir, 'va_test.c'),
+                        os.path.join(self.srcdir, 'va_test.c'))
+
+        shutil.copyfile(os.path.join(data_dir, 'Makefile'),
+                        os.path.join(self.srcdir, 'Makefile'))
+
+        build.make(self.srcdir)
+
+    def test(self):
+        '''
+        Execute VA test
+        '''
+        os.chdir(self.srcdir)
+        scenario_arg = self.params.get('scenario_arg', default=1)
+
+        if scenario_arg in [2, 3, 4]:
+            if memory.meminfo.Hugepagesize.mb != 16:
+                self.cancel(
+                    "Test need to skip as 16MB huge need to configured")
+            if scenario_arg == 4:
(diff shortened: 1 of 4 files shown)


memory/vatest.py
```diff
@@ -0,0 +1,87 @@
...
+        # Check for basic utilities
+        smm = SoftwareManager()
+        for packages in ['gcc', 'make']:
+            if not smm.check_installed(packages) and not smm.install(packages):
+                self.cancle('%s is needed for the test to be run' % packages)
+
+        data_dir = os.path.abspath(self.datadir)
+
+        shutil.copyfile(os.path.join(data_dir, 'va_test.c'),
+                        os.path.join(self.srcdir, 'va_test.c'))
```
narasimhan-v: https://github.com/avocado-framework/avocado/issues/1924

With `srcdir` being agreed upon for deprecation, please use an alternative

memory/vatest.py
```diff
@@ -0,0 +1,87 @@
...
+                        os.path.join(self.srcdir, 'Makefile'))
+
+        build.make(self.srcdir)
+
+    def test(self):
+        '''
+        Execute VA test
+        '''
+        os.chdir(self.srcdir)
+        scenario_arg = self.params.get('scenario_arg', default=1)
```
narasimhan-v: better to have in `setUp()`

memory/vatest.py
```diff
@@ -0,0 +1,87 @@
...
+        if scenario_arg in [2, 3, 4]:
+            if memory.meminfo.Hugepagesize.mb != 16:
+                self.cancel(
+                    "Test need to skip as 16MB huge need to configured")
+            if scenario_arg == 4:
+                memory.set_num_huge_pages(131072)
+            else:
+                memory.set_num_huge_pages(1024)
+        elif scenario_arg in [5, 6, 7]:
+            if memory.meminfo.Hugepagesize.gb != 16:
```
narasimhan-v: can this check be done before the `scenario_arg` check ?

memory/vatest.py
```diff
@@ -0,0 +1,87 @@
...
+    def setUp(self):
+        '''
+        Build VA Test
+        '''
+
+        # Check for basic utilities
+        smm = SoftwareManager()
+        for packages in ['gcc', 'make']:
+            if not smm.check_installed(packages) and not smm.install(packages):
+                self.cancle('%s is needed for the test to be run' % packages)
```
narasimhan-v: cancel

memory/vatest.py
```diff
@@ -0,0 +1,87 @@
...
+
+        data_dir = os.path.abspath(self.datadir)
```
narasimhan-v: is this extra variable needed ? I guess `self.datadir` already holds the absolute path

memory/vatest.py
narasimhan-v: Also the below check can be in `setUp()`

memory/vatest.py
```diff
@@ -0,0 +1,87 @@
...
+                self.cancel("Test need to skip as 16GB huge need to configured")
+            if scenario_arg == 7:
+                memory.set_num_huge_pages(2)
+            else:
+                memory.set_num_huge_pages(1)
+
+        result = process.run('./va_test -s %s' %
+                             scenario_arg, shell=True, ignore_status=True)
+        for line in result.stdout.splitlines():
+            if 'Problem' in line:
```
narasimhan-v: Should we not check all the `Problem`s instead of failing fast ?

memory/vatest.py
PraveenPenguin: ah sure typo , will rectify

memory/vatest.py
PraveenPenguin: i will remove this

memory/vatest.py
PraveenPenguin: sure i will

memory/vatest.py
PraveenPenguin: will move this on setup

memory/vatest.py
PraveenPenguin: that is requirement as only  scenario 5, 6 , 7 need 16 GB as huge page configured

memory/vatest.py
PraveenPenguin: should be fine as test code file wrote like that way if after first problem execution if test code exit (c file ) as another occurrence is not possible , any way will check this and try to modify test code (c file )

PraveenPenguin:
@narasimhan-v thanks for review , raised new PR #683  with suggested change please have a look

memory/vatest.py
narasimhan-v: This check is done for both 2,3,4 and 5,6,7

PraveenPenguin: no i don't think so for  scenario 2,3,4 we need 16MB huge page should configured

narasimhan-v: What I ask is, you are checking `memory.meminfo.Hugepagesize.mb != 16` for both cases. If that is needed, we can handle the check together

PR #685, opened 2017-11-07
Adding cxl(CAPI) tests

sudeeshjohn:
This is to test CAPI cxl driver and libcxl library . Wrapper is making use of the test cases  at https://github.com/ibm-capi/cxl-tests

Signed-off-by: sudeesh john <sudeesh@linux.vnet.ibm.com>

### DIFF
--- a/generic/cxl.py
+++ b/generic/cxl.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2017 IBM
+# Author: sudeesh john <sudeesh@linux.vnet.ibm.com>
+
+import os
+from avocado import Test, main
+from avocado.utils import build, git, process
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Cxl(Test):
+    """
+    This tests the CAPI functionality in IBM Power machines.
+    This wrapper uses the testcases from
+    https://github.com/ibm-capi/cxl-tests.git
+    """
+
+    def setUp(self):
+        """
+        Preparing the machie for the cxl test
+        """
+        self.script = self.params.get('script', default='memcpy_afu_ctx')
+        self.args = self.params.get('args', default='')
+        lspci_out = process.system_output("lspci")
+        if "accelerators" not in lspci_out:
+            self.cancel("No capi card preset. Unable to initialte the test")
+        smngr = SoftwareManager()
+        for pkgs in ['gcc', 'make', 'automake', 'autoconf']:
+            if not smngr.check_installed(pkgs) and not smngr.install(pkgs):
+                self.cancel('%s is needed for the test to be run' % pkgs)
+        git.get_repo('https://github.com/ibm-capi/cxl-tests.git',
+                     destination_dir=self.teststmpdir)
+        os.chdir(self.teststmpdir)
+        if not os.path.isfile('memcpy_afu_ctx'):
+            build.make(".")
+
+    def test(self):
+        """
+        Runs the cxl tests.
+        """
+        cmd = "./%s  %s" % (self.script, self.args)
+        result = process.run(cmd, ignore_status=True)
+        if "Unable to open cxl device" in result.stderr:
+            self.fail("%s is failed" % cmd)
+        elif "failed" in result.stdout:
+            self.fail("%s is failed" % cmd)
+
+
+if __name__ == "__main__":
+    main()
--- a/generic/cxl.py.data/cxl.yaml
+++ b/generic/cxl.py.data/cxl.yaml
@@ -1,0 +1,35 @@
+#Usage: memcpy_afu_ctx [options]
+#Options:
+#	-c <card_num>	Use this CAPI card (default 0).
+#	-h		Display this help text.
+#	-I <irq_count>	Define this number of interrupts (default 4).
+#	-i <irq_num>	Use this interrupt command source number (default 0).
+#	-k		Use the Stop_on_Invalid_Command and Restart logic.
+#	-l <loops>	Run this number of memcpy loops (default 1).
+#	-p <procs>	Fork this number of processes (default 1).
+#	-s <bufsize>	Copy this number of bytes (default 1024).
+#	-t		Do not memcpy. Test timebase sync instead.
+#	-e <timeout>		End timeout.
+#			Seconds to wait for the AFU to signal completion.
+# SAMPLE RUN
+# A typical run of the program would look like this:
+# ./memcpy_afu_ctx -e 10 -p 508 -l 10
+setup:
+  script: 'memcpy_afu_ctx'
+  general: !mux
+    memcpy_afu_ctx_1:
+        args: "-e 10 -p 508 -l 1"
+    memcpy_afu_ctx_2:
+        args: "-e 10 -p 508 -l 10"
+    memcpy_afu_ctx_3:
+        args: "-e 10 -p 508 -l 1000"
+    memcpy_afu_ctx_4:
+        args: "-e 10 -p 200 -l 100000"
+    memcpy_afu_ctx_5:
+        args: "-e 10 -p 100 -l 1000000"
+    memcpy_afu_ctx_6:
+        args: "-e 5 -p 508 -l 10"
+    cxl_eeh_tests.sh:
+      script: 'cxl_eeh_tests.sh'
+      args: ''
+      


generic/cxl.py
```diff
@@ -0,0 +1,58 @@
...
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2016 IBM
```
narasimhan-v: 2017

generic/cxl.py
```diff
@@ -0,0 +1,58 @@
...
+# Author: Pooja B Surya <pooja@linux.vnet.ibm.com>
```
narasimhan-v: Please change

generic/cxl.py
```diff
@@ -0,0 +1,58 @@
...
+        if "accelerators" not in lspci_out:
+            self.cancel("No capi card preset. Unable to initialte the test")          
+        sm = SoftwareManager()
+        deps = ['gcc', 'make', 'automake', 'autoconf']
+        for package in deps:
+            if not sm.check_installed(package) and not sm.install(package):
+                self.error(package + ' is needed for the test to be run')
+        git.get_repo('https://github.com/ibm-capi/cxl-tests.git',
+                     destination_dir=self.srcdir)
+        os.chdir(self.srcdir)
```
narasimhan-v: Please refrain from using `srcdir`. It will get deprecated soon. `teststmpdir` will be preferred.

generic/cxl.py
```diff
@@ -0,0 +1,58 @@
...
+        build.make(self.srcdir)
+
+    def test(self):
+        os.chdir(self.srcdir)
+        script = self.params.get('script', default='memcpy_afu_ctx')
+        args = self.params.get('args', default='')
```
narasimhan-v: Getting parameters can be done in `setUp()`

generic/cxl.py
```diff
@@ -0,0 +1,58 @@
...
+        if script == 'memcpy_afu_ctx':
+            cmd = './' + script + ' ' + args
+            result = process.run(cmd, ignore_status=True)
+        if script == 'cxl_eeh_tests.sh':
+            cmd = './' + script 
+            result = process.run(cmd, ignore_status=True)
```
narasimhan-v: Suggestion:
```
cmd = "./%s" % script
if script == 'memcpy_afu_ctx':
    cmd += " %s" % args
process.run('./%s' % script, ignore_status=True)
```

generic/cxl.py
```diff
@@ -0,0 +1,58 @@
...
+from avocado.utils import build
+from avocado.utils.software_manager import SoftwareManager
+from avocado.utils import distro
+
+
+class Cxl(Test):
+ 
+    def setUp(self):
+        # Check for basic utilities
+        lspci_out = process.system_output("lspci -v")
```
narasimhan-v: `lspci` output is enough here, I guess. Please clarify.

PraveenPenguin:
@sudeeshjohn please rebase with as having pep8 or pylinit compliance and you can include suggested change as well .

sudeeshjohn:
Thanks for all the comments and suggestions . will work on it .

generic/cxl.py
```diff
@@ -0,0 +1,68 @@
...
+import os
+
+from avocado import Test, main
+from avocado.utils import build, git, process
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Cxl(Test):
+    """
+        This tests the CAPI functionality in IBM Power machines.
```
PraveenPenguin: i think tab is not required for comment string

generic/cxl.py
```diff
@@ -0,0 +1,68 @@
...
+        This wrapper uses the testcases from
+        https://github.com/ibm-capi/cxl-tests.git
+    """
+
+    def setUp(self):
+        """
+            Preparing the machie for the cxl test
```
PraveenPenguin: here as well

generic/cxl.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        self.script = self.params.get('script', default='memcpy_afu_ctx')
+        self.args = self.params.get('args', default='')
```
PraveenPenguin: some default value for `self.args` will be good

generic/cxl.py
```diff
@@ -0,0 +1,68 @@
...
+        lspci_out = process.system_output("lspci")
+        if "accelerators" not in lspci_out:
+            self.cancel("No capi card preset. Unable to initialte the test")
+        smngr = SoftwareManager()
+        deps = ['gcc', 'make', 'automake', 'autoconf']
+        for pkgs in deps:
```
PraveenPenguin: can this be like 
`for pkgs in ['gcc', 'make', 'automake', 'autoconf']: ` just want to avoid extra variable

generic/cxl.py
```diff
@@ -0,0 +1,68 @@
...
+            if not smngr.check_installed(pkgs) and not smngr.install(pkgs):
+                self.error(pkgs + ' is needed for the test to be run')
```
PraveenPenguin: please change this to ` self.cancel` as well using `+` concatenating string not looks good please use this way
 
` self.cancel('%s is needed for the test to be run' %pkgs  )`

generic/cxl.py
```diff
@@ -0,0 +1,68 @@
...
+        git.get_repo('https://github.com/ibm-capi/cxl-tests.git',
```
PraveenPenguin: use `https://github.com/ibm-capi/cxl-tests/archive/master.zip` for download URL that help to use fetch_asset lib as that lib has more functionality

generic/cxl.py
```diff
@@ -0,0 +1,68 @@
...
+                     destination_dir=self.teststmpdir)
+        os.chdir(self.teststmpdir)
```
PraveenPenguin: i think this is not required

generic/cxl.py
narasimhan-v: If the above method is to be used, there is no need of 2 if conditions. Only 1 is needed.

generic/cxl.py
```diff
@@ -0,0 +1,68 @@
...
+        build.make(self.teststmpdir)
+
+    def test(self):
+        """
+            Runs the cxl tests.
+        """
+        os.chdir(self.teststmpdir)
```
narasimhan-v: Isn't this redundant ? We already are in this dir.

generic/cxl.py
```diff
@@ -0,0 +1,68 @@
...
+            if not smngr.check_installed(pkgs) and not smngr.install(pkgs):
+                self.error(pkgs + ' is needed for the test to be run')
+        git.get_repo('https://github.com/ibm-capi/cxl-tests.git',
+                     destination_dir=self.teststmpdir)
+        os.chdir(self.teststmpdir)
+        build.make(self.teststmpdir)
+
+    def test(self):
+        """
+            Runs the cxl tests.
```
PraveenPenguin: here also please about comment

generic/cxl.py.data/cxl.yaml
```diff
@@ -0,0 +1,35 @@
...
+#	-l <loops>	Run this number of memcpy loops (default 1).
+#	-p <procs>	Fork this number of processes (default 1).
+#	-s <bufsize>	Copy this number of bytes (default 1024).
+#	-t		Do not memcpy. Test timebase sync instead.
+#	-e <timeout>		End timeout.
+#			Seconds to wait for the AFU to signal completion.
+# SAMPLE RUN
+# A typical run of the program would look like this:
+# ./memcpy_afu_ctx -e 10 -p 508 -l 10
+
```
PraveenPenguin: please remove this white space

generic/cxl.py.data/cxl.yaml
```diff
@@ -0,0 +1,35 @@
...
+        args: "-e 10 -p 508 -l 1000"
+    memcpy_afu_ctx_4:
+        args: "-e 10 -p 200 -l 100000"
+    memcpy_afu_ctx_5:
+        args: "-e 10 -p 100 -l 1000000"
+    memcpy_afu_ctx_6:
+        args: "-e 5 -p 508 -l 10"
+    cxl_eeh_tests.sh:
+      script: 'cxl_eeh_tests.sh'
+      args: ''
```
PraveenPenguin: there's no line break after the `args: ''` please rectify this

generic/cxl.py
sudeeshjohn: the test has its own default values.. those will be used ..

generic/cxl.py
sudeeshjohn: I think I didn't understand this. Whats the extra benefit that I can use in this test ?

generic/cxl.py
sudeeshjohn: it requires

generic/cxl.py.data/cxl.yaml
sudeeshjohn: sure

generic/cxl.py.data/cxl.yaml
sudeeshjohn: done

generic/cxl.py
sudeeshjohn: yea.. removed

generic/cxl.py
sudeeshjohn: removed the tabs

generic/cxl.py
sudeeshjohn: can be done, but that makes the for loop looks weird.

generic/cxl.py
sudeeshjohn: done

generic/cxl.py
PraveenPenguin: ah sure

generic/cxl.py
PraveenPenguin: i don't think so as if use extra variable it increase work for garbage collector :)

generic/cxl.py
PraveenPenguin: fetch_asset we have a param as expire which used to avoid download again as till when time got expire but in git case it will do every time i think that is not needed

generic/cxl.py
PraveenPenguin: still i don't think so as build.make we are passing self.teststmpdir so chdir is not require

generic/cxl.py
```diff
@@ -0,0 +1,68 @@
...
+        cmd = "./%s " % self.script
+        if self.script == 'memcpy_afu_ctx':
+            cmd += "%s " % self.args
+            result = process.run(cmd, ignore_status=True)
+            if "Unable to open cxl device" in result.stderr:
+                self.fail("%s command failed" % cmd)
+        if self.script == 'cxl_eeh_tests.sh':
+            result = process.run(cmd, ignore_status=True)
+            if "failed" in result.stdout():
+                self.fail("%s is failed" % cmd)
```
narasimhan-v: Lines 55 to 64 can become like this:
```
cmd = "./%s" % script
if script == 'memcpy_afu_ctx':
    cmd += " %s" % args
result = process.run('./%s' % script, ignore_status=True)
if "Unable to open cxl device" in result.stderr:
    self.fail()
elif 'failed' in result.stdout:
    self.fail()
```

sudeeshjohn: done

PraveenPenguin:

PR #686, opened 2017-11-07
Fixed memory:integrity test 

PraveenPenguin:
```
1-move scenario_arg varibale from test to setup subroutine
2- added check about scenario_arg as it should 1,2 or 3

```

### DIFF
--- a/memory/integrity.py
+++ b/memory/integrity.py
@@ -39,6 +39,9 @@ class Integrity(Test):
 
         # Check for basic utilities
         smm = SoftwareManager()
+        self.scenario_arg = self.params.get('scenario_arg', default='1')
+        if self.scenario_arg not in ['1', '2', '3']:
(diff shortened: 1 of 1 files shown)


memory/integrity.py
```diff
@@ -39,6 +39,9 @@ def setUp(self):
 
         # Check for basic utilities
         smm = SoftwareManager()
+        self.scenario_arg = self.params.get('scenario_arg', default='1')
+        if self.scenario_arg not in ['1', '2', '3']:
+            self.cancel("Test need to skip as scenario will be 1-3")
```
narasimhan-v: Message can be: `Test need to skip as scenario needs to be 1-3`

PraveenPenguin:
@narasimhan-v thanks for review , made changes please have a look

PR #699, opened 2017-11-13
Test to verify RAS tools.

Pavithra1602:
Added few negetive cases for vpdupdate.

### DIFF
--- a/generic/ras_extended.py
+++ b/generic/ras_extended.py
@@ -15,7 +15,7 @@
 # Author: Pavithra <pavrampu@linux.vnet.ibm.com>
 
 import os
-from shutil import copyfile
+import shutil
 from avocado import Test
 from avocado import main
 from avocado.utils import process, distro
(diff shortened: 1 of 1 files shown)


generic/ras_extended.py
```diff
@@ -173,8 +173,23 @@ def test1_vpdupdate(self):
                                         "head -1", shell=True).strip()
         if path_db:
             copyfile_path = os.path.join(self.outputdir, 'vpd.db')
-            copyfile(path_db, copyfile_path)
+            shutil.copyfile(path_db, copyfile_path)
             self.run_cmd("vpdupdate --path=%s" % copyfile_path)
+        if os.path.exists('/var/lib/lsvpd/run.vpdupdate'):
+            path = '/var/lib/lsvpd/run.vpdupdate'
+        elif os.path.exists('/var/lib/lsvpd/run.vpdupdate'):
```
nasastry: I don't see any difference between `if` condition file path and `elif` condition file path. Both are `/var/lib/lsvpd/run.vpdupdate`.

generic/ras_extended.py
```diff
@@ -173,8 +173,23 @@ def test1_vpdupdate(self):
...
+            path = '/run/run.vpdupdate'
+        shutil.move(path, "/root/run.vpdupdate")
+        self.log.info("Running vpdupdate after removing run.vpdupdate")
+        self.run_cmd("vpdupdate")
+        shutil.move("/root/run.vpdupdate", path)
```
nasastry: Many instances having files names hardcoded and the same used in different places. It's a good practice to use variable names.

generic/ras_extended.py
```diff
@@ -173,8 +173,23 @@ def test1_vpdupdate(self):
...
+        process.run("rm -rf /var/lib/lsvpd/vpd.db; touch /var/lib/lsvpd/vpd.db",
```
nasastry: `/var/lib/lsvpd/vpd.db` is a file and to remove file there is no need of `-r`.

Pavithra1602:
Raised new PR.

PR #701, opened 2017-11-16
 Add options in xfstests for user creation and deletion on SuSE

harish-24:
xfstests user specific tests fail due to lack of home dir creation on SuSE. This patch adds option to do the same.

### DIFF
--- a/fs/xfstests.py
+++ b/fs/xfstests.py
@@ -155,9 +155,11 @@ class Xfstests(Test):
                         for test in self.gen_exclude_tests:
                             fp.write('generic/%s\n' % test)
 
-        process.run('useradd fsgqa', sudo=True)
         if self.detected_distro.name is not 'SuSE':
             process.run('useradd 123456-fsgqa', sudo=True)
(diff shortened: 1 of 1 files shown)


PraveenPenguin:
@harish-24 i think both patch  is require for  support to run  test on Suse distribution i think it should be in single commit what you say ?

harish-24:
@PraveenPenguin Both commits makes different impact on the tests, so thought of having multiple commits

PraveenPenguin:
@harish-24 sure but still i think it can be part of one commit as it good to revert or identify commit if any thing fail for particular  distribution

narasimhan-v:
@harish-24 what happens to the test with and without the commit https://github.com/avocado-framework-tests/avocado-misc-tests/pull/701/commits/80c0a8827af51d9aadb18bc25c60e22c46b3b772 ?

harish-24:
@narasimhan-v There are a set of tests which requires `80c0a88` as a setup to run properly. The previous commit is a setup for a different set of tests.

PR #703, opened 2017-11-20
code to git clone upstream sysbench

harihare:
Few distro does not provide sysbench as part of ISO, For those
distros git clone upstream sysbench, compile and use it.

Signed-off-by: Hariharan T.S <hari@linux.vnet.ibm.com>

### DIFF
--- a/generic/sysbench.py
+++ b/generic/sysbench.py
@@ -14,15 +14,14 @@
 # Copyright: 2017 IBM
 # Author: Hariharan T.S.  <harihare@in.ibm.com>
 
-
+import os
 from avocado import Test
 from avocado import main
-from avocado.utils import process
+from avocado.utils import process, git
 from avocado.utils.software_manager import SoftwareManager
 
 
 class Sysbench(Test):
-
     """
     sysbench supports following performance tests
     :cpu, threads, oltp, fileio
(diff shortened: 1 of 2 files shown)


generic/sysbench.py
```diff
@@ -48,9 +47,26 @@ def run_cmd(self, cmdline):
             self.fail("The sysbench failed: %s" % details)
 
     def setUp(self):
-        softmanager = SoftwareManager()
-        if not softmanager.check_installed('sysbench') and not softmanager.install('sysbench'):
-            self.cancel("sysbench package is required to continue the test")
+        if not process.system("which sysbench"):
+            softmanager = SoftwareManager()
+            for package in ("autoconf", "libtool", "make"):
```
PraveenPenguin: i think sysbench is required for run test if it not from upstream

generic/sysbench.py
```diff
@@ -48,9 +47,26 @@ def run_cmd(self, cmdline):
...
+                if not softmanager.check_installed(package) and not softmanager.install(package):
+                    self.error(
+                        "Fail to install %s required for this test." % package)
+            if not softmanager.check_installed('sysbench') and not softmanager.install('sysbench'):
```
PraveenPenguin: i think should be before other package installation if it failed than package installed is required to build upstream sysbench

PraveenPenguin: can be part of else block

harihare: done with the required changes.

generic/sysbench.py
```diff
@@ -48,9 +47,30 @@ def run_cmd(self, cmdline):
             self.fail("The sysbench failed: %s" % details)
 
     def setUp(self):
-        softmanager = SoftwareManager()
-        if not softmanager.check_installed('sysbench') and not softmanager.install('sysbench'):
-            self.cancel("sysbench package is required to continue the test")
+        if not process.system("which sysbench"):
+            softmanager = SoftwareManager()
+            for package in ("autoconf", "libtool", "make"):
```
PraveenPenguin: seems me redundant code:

harihare:
Missed to delete the code. done the same

PR #718, opened 2017-11-30
Fix lshw test failure collection

harish-24:
Lshw command fails were not collected properly for report. This patch fixes the failures collection.

### DIFF
--- a/generic/lshw.py
+++ b/generic/lshw.py
@@ -182,20 +182,17 @@ class Lshwrun(Test):
         self.log.info("===============Verifying -disable, -enable, -quiet, -numeric and -notime options ===="
                       "===========")
         self.is_fail = 0
-        if self.run_cmd_out("lshw -disable scsi | grep SCSI"):
-            self.is_fail += 1
+        self.run_cmd("lshw -disable scsi | grep SCSI")
         if "description" not in self.run_cmd_out("lshw -enable scsi | grep SCSI"):
             self.is_fail += 1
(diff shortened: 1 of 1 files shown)


generic/lshw.py
```diff
@@ -182,16 +182,16 @@ def test_lshw_options(self):
         self.log.info("===============Verifying -disable, -enable, -quiet, -numeric and -notime options ===="
                       "===========")
         self.is_fail = 0
-        if self.run_cmd_out("lshw -disable scsi | grep SCSI"):
+        if self.run_cmd("lshw -disable scsi | grep SCSI"):
```
narasimhan-v: `run_cmd()` increments the `is_fail` already, right ?

PraveenPenguin:
@harish-24 can you please more precise  what this patch does and execution part i see both code will have same effect ?

harish-24:
@narasimhan-v @PraveenPenguin `is_fails` is not a problem here. It gets incremented on any of the function. The actual fix is for line 197 `self.fail_cmd`, which is empty in case of `run_cmd_out`. So `if` checks are changed to capture the failed commands.

narasimhan-v:
@harish-24 Also, since fail count is captured, it would make sense to display that as well :-)

harish-24:
@narasimhan-v Sure, makes sense :-)

harish-24:
@narasimhan-v @PraveenPenguin Done with the changes. Please check.

PraveenPenguin:

Issue #716, opened 2017-11-29
Architecture dependent EEH tests fail on other architectures than powerpc


fkromer:
The tests for pci ["enhanced error handling"](https://www.kernel.org/doc/Documentation/powerpc/eeh-pci-error-recovery.txt) are specific to the powerpc architecture:

```
$ avocado list ./io/pci
INSTRUMENTED ./io/pci/PowerNVEEH.py:PowerNVEEH.test_eeh_basic_pe
INSTRUMENTED ./io/pci/PowerVMEEH.py:PowerVMEEH.test_eeh_basic_pe
...
```

However they are not tagged with `power` or some other suitable tag ([`power` seems to be used for another class of tests](https://github.com/avocado-framework-tests/avocado-misc-tests/blob/master/cpu/cpupower.py#L31)):

```
$ avocado list ./io/pci --filter-by-tags=power
$
```

They fail on other architectures (observed on ARMv7):

```
avocado run ./io/pci/PowerNVEEH.py:PowerNVEEH.test_eeh_basic_pe
JOB ID     : f42b620a9dee655e5d2adef9812052e29ae8fce7
JOB LOG    : /home/rc/avocado/job-results/job-2017-11-29T14.37-f42b620/job.log
 (1/1) ./io/pci/PowerNVEEH.py:PowerNVEEH.test_eeh_basic_pe: ERROR (0.00 s)
RESULTS    : PASS 0 | ERROR 1 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 0.43 s
```

Excerpt from the log file `/home/rc/avocado/job-results/job-2017-11-29T14.37-f42b620/job.log`:
```
2017-11-29 14:37:19,157 test             L0965 ERROR| ERROR 1-./io/pci/PowerNVEEH.py:PowerNVEEH.test_eeh_basic_pe -> TestSetupFail: [Errno 13] Permission denied: '/sys/kernel/debug/powerpc/eeh_enable'
```

fkromer:
`CANCEL` of the test during setup like in [io/pci/pci_hotplug.py:PCIHotPlugTest.test](https://github.com/avocado-framework-tests/avocado-misc-tests/blob/master/io/pci/pci_hotplug.py#L44) could be suitable.

PraveenPenguin:
@fkromer sure seems  good solution to fix this issue . Would you like to send a pull request to fix this  ?

[...]

PR #721, opened 2017-11-30
cancel pci eeh test if system is no ppc

fkromer:
closes GH-716

### DIFF
--- a/io/pci/PowerNVEEH.py
+++ b/io/pci/PowerNVEEH.py
@@ -25,6 +25,7 @@ from avocado import Test
 from avocado.utils import process
 from avocado.utils import pci
 from avocado.utils import genio
+from avocado.utils import distro
 
 EEH_HIT = 0
 EEH_MISS = 1
@@ -54,6 +55,8 @@ class PowerNVEEH(Test):
         """
         Gets the console and set-up the machine for test
         """
+        if 'ppc' not in distro.detect().arch:
+            self.cancel("Processor is not ppc64")
         output = genio.read_file("/sys/kernel/debug/powerpc/eeh_enable")\
             .strip()
         if output != '0x1':
--- a/io/pci/PowerVMEEH.py
+++ b/io/pci/PowerVMEEH.py
@@ -26,6 +26,7 @@ from avocado import Test
 from avocado.utils import process
 from avocado.utils import pci
 from avocado.utils import genio
+from avocado.utils import distro
 
 EEH_HIT = 0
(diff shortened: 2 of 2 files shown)


fkromer:
Seems like the PR failed because the [commit message is not long enough](https://travis-ci.org/avocado-framework-tests/avocado-misc-tests/builds/309547250#L558). From a functional point of view it works just fine on a system with ARMv7 processor.

vrbagalkote:
@fkromer correct from coding point it works fine, but there is a format which travis expect. Just to make travis happy, send the patch in below format.

        echo "ERR: Commit message is too short (less than 5 lines)."

        echo "     Here a minimal template:"

        echo "     ------------------------------------"

        echo "     header          <- Limited to 50 characters. No period."

        echo "                     <- Blank line"

        echo "     message         <- Any number of lines, limited to 72 characters per line."

        echo "                     <- Blank line"

        echo "     Signed-off-by:  <- Signature (created by git commit -s)"

        echo "     ------------------------------------"

io/pci/PowerNVEEH.py
```diff
@@ -54,6 +54,9 @@ def setUp(self):
         """
         Gets the console and set-up the machine for test
         """
+        cmd = "uname -p"
+        if 'ppc' not in process.system_output(cmd, ignore_status=True):
```
harish-24: Please make use of the avocado library `from avocado.utils import distro`
```
if 'ppc' not in distro.detect().arch:
    self.cancel("Processor is not ppc64")
```

io/pci/PowerVMEEH.py
```diff
@@ -56,6 +56,9 @@ def setUp(self):
         """
         Gets the console and set-up the machine for test
         """
+        cmd = "uname -p"
+        if 'ppc' not in process.system_output(cmd, ignore_status=True):
```
harish-24: here too

PraveenPenguin:
@fkromer one more request please squash all three commit in single commit

io/pci/PowerNVEEH.py
fkromer: @harish-24 Done

io/pci/PowerVMEEH.py
fkromer: Done

fkromer:
@PraveenPenguin should be ok now

PraveenPenguin:
@vrbagalkote @harish-24 thanks for review

PR #722, opened 2017-11-30
Fix: Test will check the processor chipset

vrbagalkote:
Test will now check the processor chipset, if its not
Power processor, test will exit.


vrbagalkote:
#721 is taken care of this. Sorry for the confusion.

PR #740, opened 2017-12-05
Htxfix

PraveenPenguin:
Fixed issues in htx_test 
1- fixed issue in run_smt subroutine as one process.system cmd not valid
2- change process.run('htxcmdline -geterrlog') to process.system('htxcmdline -geterrlog')
as if this command fails it make script fail and process.system is more suitable here
3- fix pep8 issue
move htx_test.data to htx_test.py.data that is more relevent.


generic/htx_test.py.data/htx_test.yaml
```diff
@@ -0,0 +1,3 @@
+time_limit: 2
```
narasimhan-v: You should remove the `generic/htx_test.data/htx_test.yaml `

generic/htx_test.py
```diff
@@ -56,8 +56,8 @@ def setUp(self):
             packages.extend(['gcc-c++', 'ncurses-devel',
                              'dapl-devel', 'libcxl-devel'])
         elif detected_distro.name == "Ubuntu":
-            packages.extend(['libncurses5', 'g++', 'libdapl-dev', 'ncurses-dev',
-                             'libncurses-dev', 'libcxl-dev'])
+            packages.extend(['libncurses5', 'g++', 'libdapl-dev',
+                             'ncurses-dev', 'libncurses-dev', 'libcxl-dev'])
```
narasimhan-v: pep8 and pylint fixes generally go as a separate commit, as per previous PRs. Consider that

PraveenPenguin: @narasimhan-v sure make sense

generic/htx_test.py.data/htx_test.yaml
PraveenPenguin: oh sorry forgot that will do , thanks for review

PraveenPenguin:
@narasimhan-v adopted the changes , can you please have a look

PR #745, opened 2017-12-06
Bonnie test fix

PraveenPenguin:
Fixed io:disk:bonnie 

1-added support as bonnie++ available from distro if it use distro version otherwise use upstream
2-ixed pep8 issue in test
Signed-off-by: Praveen K Pandey praveen@linux.vnet.ibm.com

### DIFF
--- a/io/disk/bonnie.py
+++ b/io/disk/bonnie.py
@@ -46,21 +46,36 @@ class Bonnie(Test):
         Source:
          http://www.coker.com.au/bonnie++/experimental/bonnie++-1.03e.tgz
         """
-        fstype = self.params.get('fs', default='ext4')
-        smm = SoftwareManager()
-        deps = ['gcc', 'make']
-        if distro.detect().name == 'Ubuntu':
-            deps.extend(['g++'])
-        else:
-            deps.extend(['gcc-c++'])
-        if fstype == 'btrfs':
-            if distro.detect().name == 'Ubuntu':
-                deps.extend(['btrfs-tools'])
 
-        for package in deps:
-            if not smm.check_installed(package) and not smm.install(package):
-                self.cancel("Fail to install/check %s, which is needed for"
-                            "Bonnie test to run" % package)
+        fstype = self.params.get('fs', default='ext4')
+
+        if process.system("which bonnie++", ignore_status=True):
+            smm = SoftwareManager()
+            if not smm.check_installed('bonnie++') and not smm.check_installed('bonnie++'):
+                '''Install the package from upstream'''
+                deps = ['gcc', 'make']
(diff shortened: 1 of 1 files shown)


io/disk/bonnie.py
```diff
@@ -46,21 +46,39 @@ def setUp(self):
         Source:
          http://www.coker.com.au/bonnie++/experimental/bonnie++-1.03e.tgz
         """
+
```
harish-24: can you remove this line?

io/disk/bonnie.py
```diff
@@ -46,21 +46,39 @@ def setUp(self):
...
-                            "Bonnie test to run" % package)
+
+        if process.system("which bonnie++", ignore_status=True):
+            smm = SoftwareManager()
+            if not smm.check_installed('bonnie++') \
+                    and not smm.check_installed('bonnie++'):
+                '''Install the package from upstream'''
+                deps = ['gcc', 'make']
+                if distro.detect().name == 'Ubuntu':
+                    deps.extend(['g++'])
```
harish-24: Can you extend the `btrfs` check of L62 here?

io/disk/bonnie.py
```diff
@@ -99,15 +108,16 @@ def test(self):
...
+        cmd = ('bonnie++ %s' % " ".join(args))
         if process.system(cmd, shell=True, ignore_status=True):
             self.fail("test failed")
 
     def tearDown(self):
         '''
         Cleanup of disk used to perform this test
         '''
+        if os.listdir(self.source):
+            build.make(self.source, extra_args='clean')
```
harish-24: As it will not `uninstall` the binary from system, can you help me why it is needed?

PraveenPenguin: aah you are right i thought it will uninstall binary as well

io/disk/bonnie.py
PraveenPenguin: not get that i think that is suitable if condition can you help me to rectify this

io/disk/bonnie.py
```diff
@@ -46,21 +46,39 @@ def setUp(self):
...
-
-        for package in deps:
-            if not smm.check_installed(package) and not smm.install(package):
-                self.cancel("Fail to install/check %s, which is needed for"
-                            "Bonnie test to run" % package)
+
+        if process.system("which bonnie++", ignore_status=True):
+            smm = SoftwareManager()
+            if not smm.check_installed('bonnie++') \
+                    and not smm.check_installed('bonnie++'):
```
Naresh-ibm: I think it should be smm.install('bonnie++').

PraveenPenguin:
closing this PR as new version of this Test raised #784

PR #746, opened 2017-12-06
Fix: Certain tests were failing due to MFC values, which is handeled

vrbagalkote:
Tests such as ghs_dhs rebuild online_offline were failing, due to perticular
MFC values, and all those scenario are handeled now. Also changes made for migrate
test case, if there is no enough drives to pop for migration, test will exit.

### DIFF
--- a/io/disk/Avago_storage_adapter/avago9361_vd.py
+++ b/io/disk/Avago_storage_adapter/avago9361_vd.py
@@ -83,6 +83,13 @@ class Avago9361(Test):
         self.io_policy = ['direct', 'cached']
         self.stripe = [64, 128, 256, 512, 1024]
         self.state = ['start', 'stop', 'start', 'pause', 'resume']
+        self.list_test = ['online_offline', 'rebuild']
+        if str(self.name) in self.list_test:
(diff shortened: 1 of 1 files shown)


vrbagalkote:
Attached is job.log file
[job.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1534434/job.txt)

io/disk/Avago_storage_adapter/avago9361_vd.py
```diff
@@ -156,6 +163,8 @@ def test_online_offline(self):
                 self.set_online_offline(state)
                 time.sleep(10)
         self.vd_delete()
+        cmd = "%s /c%d set autorebuild=on" % (self.tool, self.controller)
+        self.check_pass(cmd, "Failed to set auto rebuild on")
```
narasimhan-v: Guess this is not needed, as they are already handled in `setUp()` and `tearDown()`

narasimhan-v:
@vrbagalkote There are more changes done. Please explain all in the commit message.

PraveenPenguin:

Issue #739, opened 2017-12-04
Test Lshwrun.test_lshw_verification fails due to command line output not considered yet


fkromer:
Test `generic/lshw.py:Lshwrun.test_lshw_verification` [fails in this check](https://github.com/avocado-framework-tests/avocado-misc-tests/blob/master/generic/lshw.py#L108) due to `ifconfig` output not considered yet.

Output of `ifconfig | head -1 | cut -d':' -f1`:

```
eth0      Link encap
```

Output of `lshw -class network`:

```
WARNING: you should run this program as super-user.
  *-network               
       ...
       logical name: eth0
       ...
   ...
WARNING: output may be incomplete or inaccurate, you should run this program as super-user.
```

harish-24:
@fkromer Can you elaborate the issue here? Apart from running with `sudo`, is there anything else?

fkromer:
@harish-24 Probably the test fails [here](https://github.com/avocado-framework-tests/avocado-misc-tests/blob/master/generic/lshw.py#L108)

```
active_interface = process.system_output("ifconfig | head -1 | cut -d':' -f1", shell=True).strip()
...
if self.active_interface not in process.system_output("lshw -class network"):
    self.fail("lshw failed to show correct active network interface")
```

because the string output `eth0      Link encap` of `ifconfig | head -1 | cut -d':' -f1` (in variable `active_interface`) is not contained in `process.system_output("lshw -class network")`

```
WARNING: you should run this program as super-user.
  *-network               
       ...
       logical name: eth0
       ...
   ...
WARNING: output may be incomplete or inaccurate, you should run this program as super-user.
```

However I am not sure about `.strip()` in `process.system_output("ifconfig | head -1 | cut -d':' -f1", shell=True).strip()` which probably strips `eth0      Link encap` to `eth0`. (If that would be the case I have no idea why the test failed.)

harish-24:
@fkromer Yeah the problem is with the `strip()` part. `"eth0 Link encap".strip()` strips only the extra spaces. As a fix you could as something like `"eth0 Link encap".strip().split()[0]` which gives you `eth0`

fkromer:
@harish-24 Ah right. As `.strip().split()[0]` does work in case no link encapsulation info is included in the output of `ifconfig` as well I PR with this your solution.

PraveenPenguin:
closing this issue as PR already merged

PR #748, opened 2017-12-06
consider link encapsulation output of ifconfig

fkromer:
The call of ifconfig may contain info about the
link encapsulation in the output in addition to
the network interface. If the link encapsulation
info is not included in the output `.split()[0]` works
as well.

Reference: GH-739
Signed-off-by: kromer <florian.kromer@roboception.de>


fkromer:
@PraveenPenguin Thanks for flowing dev workflow.

PR #761, opened 2017-12-16
Revamp of PowerNVEEH using memory address and mask

narasimhan-v:
* Revamped PowerNV EEH to use memory address and mask.
* Fixed the code to pass only necessary data between functions.
* Pylint fixes

[EEH_avocado_log.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1571569/EEH_avocado_log.txt)

[Direct_PHB_Power8_EEH_job_log.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1578650/EEH_job_log.txt)

### DIFF
--- a/io/pci/PowerNVEEH.py
+++ b/io/pci/PowerNVEEH.py
@@ -35,13 +35,13 @@ class EEHRecoveryFailed(Exception):
     """
     Exception class, if EEH fails to recover
     """
-    def __init__(self, thing, dev, log=None):
-        self.thing = thing
+    def __init__(self, msg, dev, log=None):
+        self.msg = msg
         self.dev = dev
         self.log = log
 
     def __str__(self):
-        return "%s %s recovery failed: %s" % (self.thing, self.dev, self.log)
+        return "%s %s recovery failed: %s" % (self.msg, self.dev, self.log)
 
 
 class PowerNVEEH(Test):
@@ -65,13 +65,22 @@ class PowerNVEEH(Test):
         cmd = "echo %d > /sys/kernel/debug/powerpc/eeh_max_freezes"\
             % self.max_freeze
         process.system(cmd, ignore_status=True, shell=True)
-        self.function = str(self.params.get('function'))
+        self.function = str(self.params.get('function'), default='4')
         self.err = str(self.params.get('err'))
         self.pci_device = str(self.params.get('pci_device', default=' '))
         self.phb = self.pci_device.split(":", 1)[0]
         self.addr = genio.read_file("/sys/bus/pci/devices/%s/"
                                     "eeh_pe_config_addr" % self.pci_device)
         self.addr = str(self.addr).rstrip()
+        self.err = 0
+        for line in process.system_output('lspci -vs %s' % self.pci_device,
+                                          ignore_status=True,
+                                          shell=True).splitlines():
+            if 'Memory' in line and '64-bit, prefetchable' in line:
+                self.err = 1
+                break
+        self.mem_addr = pci.get_memory_address(self.pci_device)
+        self.mask = pci.get_mask(self.pci_device)
         self.log.info("Test-----> %s", self.addr)
         self.log.info("===============Testing EEH Frozen PE==================")
 
@@ -85,20 +94,17 @@ class PowerNVEEH(Test):
         num_of_hit = 0
         while num_of_hit <= self.max_freeze:
             for func in self.function:
-                self.log.info("Running error inject on pe %s function %s"
-                              % (self.pci_device, func))
+                self.log.info("Running error inject on pe %s function %s",
+                              self.pci_device, func)
                 if num_of_miss < 5:
-                    return_code = self.basic_eeh(self.addr, self.err, func,
-                                                 self.phb, self.pci_device)
+                    return_code = self.basic_eeh(func)
                     if return_code == EEH_MISS:
                         num_of_miss += 1
-                        self.log.info("number of miss is %d"
-                                      % num_of_miss)
+                        self.log.info("number of miss is %d", num_of_miss)
                         continue
                     else:
                         num_of_hit += 1
-                        self.log.info("number of hit is %d"
-                                      % num_of_hit)
+                        self.log.info("number of hit is %d", num_of_hit)
                         if num_of_hit <= self.max_freeze:
                             if not self.check_eeh_pe_recovery():
                                 self.fail("PE %s recovery failed after"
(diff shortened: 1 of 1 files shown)


io/pci/PowerNVEEH.py
```diff
@@ -116,35 +121,35 @@ def test_eeh_basic_pe(self):
...
-    def error_inject(cls, addr, err, func, phb, pci_device):
+    def error_inject(cls, addr, err, func, mem_addr, mask, phb):
 
         """
         Form a command to inject the error
         """
-        cmd = "echo %s:%s:%s:0:0 > /sys/kernel/debug/powerpc/PCI%s/err_injct \
-               && lspci -ns %s; echo $?" % (addr, err, func, phb, pci_device)
+        cmd = "echo %s:%s:%s:%s:%s > /sys/kernel/debug/powerpc/PCI%s/err_injct \
+               && lspci; echo $?" % (addr, err, func, mem_addr, mask, phb)
```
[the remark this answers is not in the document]
narasimhan-v: @vrbagalkote yes, supposed to.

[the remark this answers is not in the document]
vrbagalkote: can you please attach job.log file plz

io/pci/PowerNVEEH.py
```diff
@@ -65,13 +65,21 @@ def setUp(self):
...
         self.phb = self.pci_device.split(":", 1)[0]
         self.addr = genio.read_file("/sys/bus/pci/devices/%s/"
                                     "eeh_pe_config_addr" % self.pci_device)
         self.addr = str(self.addr).rstrip()
+        self.err = 0
+        for line in process.system_output('lspci -vs %s' % self.pci_device,
+                                          ignore_status=True,
+                                          shell=True).splitlines():
+            if 'Memory' in line and '64-bit, prefetchable' in line:
+                self.err = 1
```
vrbagalkote: there are certain instances where, its 64bit and non-prefect...

 # lspci -v | grep -i "64-bit"
    Memory at 6000000000000 (64-bit, prefetchable) [size=32M]
    Memory at 6000002000000 (64-bit, prefetchable) [size=32M]
    Memory at 600c080100000 (64-bit, non-prefetchable) [size=64K]
    Memory at 6004000000000 (64-bit, prefetchable) [size=8M]
    Memory at 6004002000000 (64-bit, prefetchable) [size=32K]
    Memory at 6004000800000 (64-bit, prefetchable) [size=8M]
    Memory at 6004002010000 (64-bit, prefetchable) [size=32K]
    Memory at 6004001000000 (64-bit, prefetchable) [size=8M]
    Memory at 6004002020000 (64-bit, prefetchable) [size=32K]
    Memory at 6004001800000 (64-bit, prefetchable) [size=8M]
    Memory at 6004002030000 (64-bit, prefetchable) [size=32K]
    Memory at 600c100100000 (64-bit, non-prefetchable) [size=64K]
    Memory at 600c180000000 (64-bit, non-prefetchable) [size=1M]
    Memory at 600c180180000 (64-bit, non-prefetchable) [size=1K]
    Memory at 600c280000000 (64-bit, non-prefetchable) [size=64K]
    Memory at 600c280010000 (64-bit, non-prefetchable) [size=8K]
    Memory at 6200000000000 (64-bit, prefetchable) [size=32M]
    Memory at 6200002000000 (64-bit, prefetchable) [size=32M]
    Memory at 620c080100000 (64-bit, non-prefetchable) [size=64K]
    Memory at 620c100100000 (64-bit, non-prefetchable) [size=64K]
    Memory at 620c180000000 (64-bit, non-prefetchable) [size=16K]
    Memory at 6220000000000 (64-bit, prefetchable) [size=1M]

narasimhan-v:
```

File: io/pci/PowerNVEEH.py (hunk 1 above)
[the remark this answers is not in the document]
Comment (narasimhan-v) [reply io/pci/PowerNVEEH.py]: @vrbagalkote Done. Please check.

File: io/pci/PowerNVEEH.py (hunk 2 above)
Comment (vrbagalkote) [reply io/pci/PowerNVEEH.py]: In 32 bit also we have both non-prefetchable and prefetchable, does both comes under mode 0. Are you referring any document, can u please share.

File: io/pci/PowerNVEEH.py [hunk 3]
```diff
@@ -65,13 +65,21 @@ def setUp(self):
...
         process.system(cmd, ignore_status=True, shell=True)
-        self.function = str(self.params.get('function'))
-        self.err = str(self.params.get('err'))
+        self.function = str(self.params.get('function'), default='4')
         self.pci_device = str(self.params.get('pci_device', default=' '))
         self.phb = self.pci_device.split(":", 1)[0]
         self.addr = genio.read_file("/sys/bus/pci/devices/%s/"
                                     "eeh_pe_config_addr" % self.pci_device)
         self.addr = str(self.addr).rstrip()
+        self.err = 0
```
vrbagalkote: Its better not to assume this. We should be doing a check, may add else condition below, after checking for 64bit

io/pci/PowerNVEEH.py
narasimhan-v: @vrbagalkote will share in private.

io/pci/PowerNVEEH.py
narasimhan-v: This is not assumption. This is initialising to default value. Only in one case, it is 1, which is handled in `if`

vrbagalkote: I meant, what if lspci itself was reporting wrong, in that case we miss it. Also, there is good chance there could be 128 bit or something like that.

narasimhan-v: @vrbagalkote What alternatives do we have ?

vrbagalkote:
```
   if 'Memory' in line and '64-bit, prefetchable' in line:
                self.err = 1
   else 'Memory' in line and '32-bit, ...' in line:
                self.err = 0
```
is better

narasimhan-v: https://en.wikipedia.org/wiki/PCI_configuration_space

From here, only `32 and 64 bit`, and only `prefetchable and non-prefetchable` are possible.

So, out of 4 cases, only 1 case needs `err` value to be 1. So, the initial approach is better here :-)

vrbagalkote: Two things:

1: From Gavin's note, error code can be 0/1 for any adapter, its only based on function we are going to use while injecting error. Its not dependent on the adapter configuration size.

2: From the notes you shared, also says for direct PHB adapters, address and mask should be 0, 0. You are not checking the adapter whether its direct PHB or not. So I am sceptical whether this change is required at the first place.
```
Gavin's Note:

In the early days of EEH development, the debugfs entry and command are used to
to error injection (Frozen PE case). However, it's not reliable way. At the meanwhile,
I plan to drop the debugfs entry pretty soon. Instead, the below debugfs entry is the
correct one for error injection. 

    /sys/kernel/debug/powerpc/PCIxxxx/err_injct

The debugfs entry accepts commands "PE_number:<0,1>:<function>:0:0"

PE_number: The number of PE the specified PCI device is attached to. It can be retrieved
           from /sys/bus/pci/devices/xxxx:xx:xx.x/eeh_pe_config_addr
<0,1>:     0 - 32-bits error injection. When doing error injection via PCI config, IO or
           32-bits MMIO access.
           1 - 64-bits error injection. When doing error injection via 64-bits MMIO access
           or DMA.
<function>: Please refer to Linux source code arch/powerpc/include/asm/opal-api.h::OpalErrinjectFunc
```
So, finally error code 0/1 is not dependent on adapter configuration. We should be covering both errors on any adapter, thats the reason why its in yaml configuration file. So its up to user.

narasimhan-v: 1. Error being 0 / 1, is based on lspci output. Can you try once injecting error 0 for 64-bit prefetchable BARs and let me know ?

vrbagalkote: The statement I pasted is from commit log message. And guliherme's notes is for his reference, and he put a clear disclaimer. :). So I would like to believe commit message than a personal reference notes

narasimhan-v: Ok, are we clear about direct PHB slots ?

narasimhan-v: About error being 0/1, can you please point me exactly where it says it is user configurable and not based on PCI config space ?

narasimhan-v: @vrbagalkote I tried the script with same patches on Power8 server, on a direct PHB slot:

vrbagalkote: Thanks @narasimhan-v . You have cleared one ambiguity out of the way. So, we can use address mask and memory for any kind of adapter. But having said that we can also 0 and 0 also for any adapter. So I am happy either way. 

One last thing I want to is, Can you also check if error code 0 and 1 works fine for any adapter?
If it works, then there is no need to calculate it in the script, it can be given via yaml file(user configurable).

narasimhan-v: > So I am happy either way.

Cool

vrbagalkote: Exactly for this reason, let user decide which function and which error type. There is a chance that user gives MMIO function and scripts evaluates to wrong error code. Let user decide both, till we figure out what exactly.

narasimhan-v: @vrbagalkote For MMIO function, we should calculate from lspci. So, it should be computed.

narasimhan-v: >  There is a chance that user gives MMIO function and scripts evaluates to wrong error code.

narasimhan-v:
@vrbagalkote I have removed the changes related to `err` type for now. We can deal that later once we have a clarity on it. Please check now.

PraveenPenguin:
@vrbagalkote thanks for review

PR #765, opened 2017-12-18
Support to run distribution specific ltrace test

harish-24:
Added support to run tests from distrution ltrace source

### DIFF
--- a/toolchain/ltrace.py
+++ b/toolchain/ltrace.py
@@ -69,15 +69,23 @@ class Ltrace(Test):
             if not smm.check_installed(package) and not smm.install(package):
                 self.cancel("Fail to install %s required for this test." %
                             package)
+        run_type = self.params.get("type", default="upstream")
+        if run_type == "upstream":
+            source = self.params.get('url', default="git://git.debian.org/git/"
+                                     "collab-maint/ltrace.git")
+            git.get_repo(source, destination_dir=os.path.join(
+                self.srcdir, 'ltrace'))
 
-        # Source: git clone git://git.debian.org/git/collab-maint/ltrace.git
-        git.get_repo('git://git.debian.org/git/collab-maint/ltrace.git',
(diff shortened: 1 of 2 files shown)


PraveenPenguin:
@harish-24 please resolved pep8 issue and rebase

toolchain/ltrace.py
```diff
@@ -69,15 +69,22 @@ def setUp(self):
...
                      destination_dir=os.path.join(self.srcdir, 'ltrace'))
 
-        self.src_lt = os.path.join(self.srcdir, "ltrace")
-        os.chdir(self.src_lt)
-        process.run('patch -p1 < %s' %
-                    os.path.join(self.datadir, 'ltrace.patch'), shell=True)
+            self.src_lt = os.path.join(self.srcdir, "ltrace")
+            os.chdir(self.src_lt)
+            process.run('patch -p1 < %s' %
+                        os.path.join(self.datadir, 'ltrace.patch'), shell=True)
```
sathnaga: @harish-24 Does it always true, having ltrace.patch?,what if not it would be better to handle with a flag so that direct git HEAD can be tested. and how about moving the git link to config(yaml) so that custom gits can be tested, applicable for all such occurrences in other tests too..

PraveenPenguin: @sathnaga ltrace.patch is temp fix for compilation issue of test suite , i already submitted patch against ltrace  source , when it got accepted we can remove from test code

harish-24: @sathnaga Sure, I will move the url to config.

harish-24:
@sathnaga @PraveenPenguin Changed as suggested. Please have a look. Thanks.

PraveenPenguin:
@sathnaga sure will remove this patch when it accepted via ltrace community

PraveenPenguin:
@sathnaga thanks for review

PR #770, opened 2017-12-23
Compressing multiple yaml into 1

narasimhan-v:
In reference to commit 3cc16493d034db3a385ab06941267337ff987ce8,
wherein since profilers and sysinfo per test was not supported,
we allowed this commit to have multiple yaml files.

Since that was fixed in avocado, we are compressing the yaml files
into single yaml.


narasimhan-v:
@Venutiwa @harihare 

https://github.com/avocado-framework/avocado/pull/2272/

https://github.com/open-power-host-os/tests/pull/85

PraveenPenguin:
@narasimhan-v thanks for the patch , @harihare thanks for ACK , merging as already one ACK for this PR

PR #773, opened 2018-01-01
Fix: ignore the kill command status

abdhaleegit:
Some times the process gets killed while testing affinity as expected,
so kill command fails when process not found so ignore the sataus

### DIFF
--- a/cpu/cpustress.py
+++ b/cpu/cpustress.py
@@ -116,7 +116,7 @@ class cpustresstest(Test):
     @staticmethod
     def __kill_process(pids):
         for pid in pids:
-            process.run("kill -9 %s" % pid)
+            process.run("kill -9 %s" % pid, ignore_status=True)
 
     def test(self):
         """


cpu/cpustress.py
```diff
@@ -116,7 +116,7 @@ def __cpu_toggle(core):
     @staticmethod
     def __kill_process(pids):
         for pid in pids:
-            process.run("kill -9 %s" % pid)
+            process.run("kill -9 %s" % pid, ignore_status=True)
```
narasimhan-v: @abdhaleegit both the approaches does not catch the failure and do anything with it. Is that intentional / fine ?

abdhaleegit: It is fine, as we are setting cpu affinity in random way, it is expected that process gets terminated...so at the time of cleanup 'kill' command fails if pid not found.. making the test fail.

so we are not catching any failure here.

narasimhan-v:
Waiting for end of day, will merge if no other comments from anyone.

narasimhan-v:
@abdhaleegit can you force push once, so that travis can run once again ?

PR #776, opened 2018-01-03
Handling smt snooze delay for sles 11

narasimhan-v:
In SLES 11, and could be with other older distros, the smt snooze
delay of cpu in sysfs is /sys/devices/system/cpu/.
Whereas in newer distros, it is /sys/bus/cpu/devices/.
Handled this in this commit.

### DIFF
--- a/cpu/ppc64_cpu_test.py
+++ b/cpu/ppc64_cpu_test.py
@@ -147,7 +147,11 @@ class PPC64Test(Test):
         Tests the smt snooze delay in ppc64_cpu command.
         """
         command1 = "ppc64_cpu --smt-snooze-delay | awk '{print $NF}'"
-        command2 = "cat /sys/bus/cpu/devices/cpu*/smt_snooze_delay | uniq"
+        snz_delay = "cpu*/smt_snooze_delay"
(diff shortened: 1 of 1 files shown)


cpu/ppc64_cpu_test.py
```diff
@@ -147,7 +147,11 @@ def smt_snoozedelay(self):
         Tests the smt snooze delay in ppc64_cpu command.
         """
         command1 = "ppc64_cpu --smt-snooze-delay | awk '{print $NF}'"
-        command2 = "cat /sys/bus/cpu/devices/cpu*/smt_snooze_delay | uniq"
+        tmp = "cpu*/smt_snooze_delay"
```
harish-24: can `tmp` be changed to something like `snz_delay_file`? Otherwise, looks good.

narasimhan-v: @harish-24 agreed, will change and send a patch.

narasimhan-v: @harish-24 Done. please check now.

PraveenPenguin:
@narasimhan-v thanks for patch , @harish-24 @narasimhan-v thanks for review

PR #779, opened 2018-01-08
Add proper stat for thp on redhat

harish-24:
thp_split_page is not available on redhat, support has been added in this patch with thp_split

### DIFF
--- a/memory/transparent_hugepages.py
+++ b/memory/transparent_hugepages.py
@@ -48,6 +48,12 @@ class Thp(Test):
         self.mem_path = os.path.join(data_dir.get_tmp_dir(), 'thp_space')
         free_mem = int(memory.freememtotal() / 1024)
         self.dd_timeout = 900
+        self.thp_split = None
+        try:
+            memory.read_from_vmstat("thp_split_page")
+            self.thp_split = "thp_split_page"
+        except IndexError:
+            self.thp_split = "thp_split"
 
         # Set block size as hugepage size * 2
         self.block_size = (memory.get_huge_page_size() / 1024) * 2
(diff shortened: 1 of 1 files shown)


memory/transparent_hugepages.py
```diff
@@ -48,6 +49,11 @@ def setUp(self):
         self.mem_path = os.path.join(data_dir.get_tmp_dir(), 'thp_space')
         free_mem = int(memory.freememtotal() / 1024)
         self.dd_timeout = 900
+        self.thp_split = None
+        if distro.detect().name == "rhel":
+            self.thp_split = "thp_split"
+        else:
+            self.thp_split = "thp_split_page"
```
nasastry: The above can also be implemented in a simple way.
```
self.thp_split = "thp_split_page"
if distro.detect().name == "rhel":
    self.thp_split = "thp_split"
```

harish-24: Sure

harish-24:
@nasastry Done. Thanks.

memory/transparent_hugepages.py
```diff
@@ -48,6 +49,9 @@ def setUp(self):
         self.mem_path = os.path.join(data_dir.get_tmp_dir(), 'thp_space')
         free_mem = int(memory.freememtotal() / 1024)
         self.dd_timeout = 900
+        self.thp_split = "thp_split_page"
+        if distro.detect().name == "rhel":
```
PraveenPenguin: pegas release still having `thp_split_page`   we need to consider that as well

narasimhan-v:
@harish-24 cosmetic, but a worthy change: please change redhat to rhel in commit message

harish-24:
@narasimhan-v Sure, changed accordingly.

memory/transparent_hugepages.py
narasimhan-v: Instead of handling based on distros, handling the IndexError with either values would be my suggestion.

harish-24: @narasimhan-v As some rhel releases may have `thp_split_page`, it makes sense to do it this way, will make the changes needed.

harish-24:
@narasimhan-v @PraveenPenguin Changed based on the comments, please verify. Thanks.

PR #786, opened 2018-01-22
Fix numatop compilation

harish-24:
numatop compile failed due to a package install failure. Patch adds the package required

### DIFF
--- a/cpu/numatop.py
+++ b/cpu/numatop.py
@@ -47,7 +47,8 @@ class Numatop(Test):
         if not detected_distro == "ubuntu":
             self.cancel("Upsupported OS %s" % detected_distro)
         smm = SoftwareManager()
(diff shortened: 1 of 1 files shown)


cpu/numatop.py
```diff
@@ -47,7 +47,8 @@ def setUp(self):
         if not detected_distro == "ubuntu":
             self.cancel("Upsupported OS %s" % detected_distro)
         smm = SoftwareManager()
-        for package in ['gcc', 'numatop', 'make', 'libnuma-dev']:
+        for package in ['gcc', 'numatop', 'make', 'libnuma-dev',
+                        'libncurses-dev']:
```
narasimhan-v: @harish-24 package name same for all supported distros ?

PraveenPenguin:
@harish-24 thanks for patch , merging as already one ACK , @narasimhan-v thanks for ACK

PR #790, opened 2018-01-31
Add numa scenario in stress-ng

harish-24:
Stress-ng can be run with numa, patch adds the same


PraveenPenguin:
@abdhaleegit your review comment will help here as you are the owner for test

PR #800, opened 2018-02-09
Support to run individual component of selftest

harish-24:
Added yaml to run each component of selftest as a variant

### DIFF
--- a/kernel/kselftest.py
+++ b/kernel/kselftest.py
@@ -52,7 +52,7 @@ class kselftest(Test):
         if 'Ubuntu' in detected_distro.name:
             deps.extend(['libpopt0', 'libc6', 'libc6-dev',
                          'libpopt-dev', 'libcap-ng0', 'libcap-ng-dev',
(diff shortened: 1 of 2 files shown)


kernel/kselftest.py
```diff
@@ -81,7 +84,8 @@ def test(self):
         Execute the kernel selftest
         """
         error = False
-        result = build.make(self.sourcedir, extra_args='run_tests')
+        result = build.make(
+            self.sourcedir, extra_args='%s run_tests' % self.comp)
```
narasimhan-v: 2 questions:
1. Is the null value for `self.comp` taken care ?
2. no space needed before -C ? or is that also taken care ?

harish-24: 1. Yes, Line 47 handles it, and replaces with empty string.
2. Yes, as part of `build_make` it is taken care.

PraveenPenguin:
@harish-24 thanks for patch , merging as already on ACK for this PR , @narasimhan-v thanks for ACK

PR #801, opened 2018-02-14
Fix numa as stressor instead of class

harish-24:
'numa' was used wrongly as class instead of stress. Handled a minor fix if either of the stressors are None


harish-24:
Closing this as for merge confilct. opening new commit #802

PR #804, opened 2018-02-22
Implement the parameter "iterations".

tambewilliam:
This commit implement the parameter "iterations"
allowing the test to be run multiple time, in a similar
manner that "iterations" is implemented in kernel/kernbench.py

At AMD, we use the "iterations" feature to smooth out benchmark values by running them multiple times and using the average value.


PraveenPenguin:
@tambewilliam thanks for patch , can you please add this in

tambewilliam:
@PraveenPenguin "iterations" parameter added in test case yaml file:
`avocado-misc-tests/perf/hackbench.py.data/hackbench.yaml`

PR #807, opened 2018-03-02
Hugepage sanity test to make use supported hugepages

harish-24:
This test uses mmap call to allocate and memset hugepages of supported sizes. By default uses default hugepage size allocating one page of it.

### DIFF
--- a/memory/hugepage_sanity.py
+++ b/memory/hugepage_sanity.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2018 IBM
+# Author: Harish <harish@linux.vnet.ibm.com>
+#
+
+
+import os
+import shutil
+from avocado import Test
+from avocado import main
+from avocado.utils import process, build
+from avocado.utils.software_manager import SoftwareManager
+
+
+class HugepageSanity(Test):
+    """
+    Test allocates given number of hugepages of given size and mmap's
+    using MAP_HUGETLB with corresponding hugepage sizes
+    """
+    def copyutil(self, file_name):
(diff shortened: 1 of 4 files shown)


memory/hugepage_sanity.py
```diff
@@ -0,0 +1,58 @@
...
+    Test allocates given no-hugepages of given size and mmap's
+    using MAP_HUGETLB with corresponding hugepage sizes
+    """
+    def copyutil(self, file_name):
+        shutil.copyfile(os.path.join(self.datadir, file_name),
+                        os.path.join(self.teststmpdir, file_name))
+
+    def setUp(self):
+        smm = SoftwareManager()
+        self.hpagesize = int(self.params.get('hpagesize', default=''))
```
PraveenPenguin: can it has some default value ?

memory/hugepage_sanity.py
```diff
@@ -0,0 +1,58 @@
...
+        self.no_huge = int(self.params.get('no_pages', default=''))
```
PraveenPenguin: this also

PraveenPenguin: please rename this variable as `num_huge` as `no_huge` arise confusion as no huge page

memory/hugepage_sanity.py.data/hugepage_sanity.c
```diff
@@ -0,0 +1,152 @@
+#include <stdio.h>
```
PraveenPenguin: please add license

memory/hugepage_sanity.py
```diff
@@ -0,0 +1,58 @@
...
+import shutil
+from avocado import Test
+from avocado import main
+from avocado.utils import process, build
+from avocado.utils.software_manager import SoftwareManager
+
+
+class HugepageSanity(Test):
+    """
+    Test allocates given no-hugepages of given size and mmap's
```
PraveenPenguin: her as well `number of hugepages`

harish-24: Sure

memory/hugepage_sanity.py
harish-24: A default value may be different from system's default hugepage. Hence, if it is empty the c script will use the default hugepage of the machine. It will be repetitive to do the same here.

memory/hugepage_sanity.py
harish-24: 1) The same case here as well, it will be repetitive as c script will use single page if not provided.
2) sure

memory/hugepage_sanity.py.data/hugepage_sanity.c
harish-24: Sure

PR #809, opened 2018-03-08
Fix xfs

PraveenPenguin:
1-Fix compilation issue in recent ubuntu distro 
2- fixed multiple pep8 issue

### DIFF
--- a/fs/xfstests.py
+++ b/fs/xfstests.py
@@ -63,6 +63,8 @@ class Xfstests(Test):
                  'uuid-runtime', 'libaio-dev', 'fio', 'dbench', 'btrfs-tools'])
             if '14' in self.detected_distro.version:
                 packages.extend(['libtool'])
+            elif '18' in self.detected_distro.version:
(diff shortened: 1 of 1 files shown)


fs/xfstests.py
```diff
@@ -65,11 +65,14 @@ def setUp(self):
                 packages.extend(['libtool'])
             else:
                 packages.extend(['libtool-bin'])
+            if '18' in self.detected_distro.version:
```
harish-24: Can we add this as an `elif` prior to this so that on every run `version` is checked only once?

PraveenPenguin: @harish-24 yes make sense will do as one info it new change in xfs test may for other distro case we also need to take care in future till it clear all other distro

PraveenPenguin:
@harish-24 adopted the changes please have look

PraveenPenguin:
@narasimhan-v can you please have a look on this PR

PR #810, opened 2018-03-08
generic/ras.py: skip test instead of canceling it

clebergnu:
It makes more sense to skip a test altogether if possible.


PraveenPenguin:
@harish-24 thanks for trying this patch

PR #817, opened 2018-03-14
Switch from Test.srcdir to Test.workdir

clebergnu:
Due to Test.srcdir deprecation on Avocado.

Reference: https://github.com/avocado-framework/avocado/pull/2509


cpu/ebizzy.py
```diff
@@ -55,9 +55,9 @@ def setUp(self):
                                    '/project/ebizzy/ebizzy/0.3'
                                    '/ebizzy-0.3.tar.gz')
         data_dir = os.path.abspath(self.datadir)
-        archive.extract(tarball, self.srcdir)
+        archive.extract(tarball, self.wordir)
```
harish-24: Probably a typo `wordir` for `workdir`. Please change elsewhere.

fs/filebench.py
```diff
@@ -52,9 +52,9 @@ def setUp(self):
                                    'filebench/releases/download/1.5-alpha3/'
                                    '%s.tar.gz' % name_version)
 
-        archive.extract(tarball, self.srcdir)
+        archive.extract(tarball, self.wordir)
```
harish-24: In this test, `self.workdir` is used already for another purpose. Please take care of having unique directories.

generic/kdump.py
```diff
@@ -51,7 +51,7 @@ def setUp(self):
         self.prompt = self.params.get('prompt', default='')
 
     def test(self):
-        log_file = os.path.join(self.srcdir, "file")
+        log_file = os.path.join(self.wordir, "file")
```
harish-24: Here too, `self.workdir` is already used for another purpose

io/disk/lvsetup.py
```diff
@@ -81,7 +81,7 @@ def setUp(self):
         self.lv_name = lv_name
         if lv_utils.lv_check(vg_name, lv_snapshot_name):
             self.cancel('Snapshot %s already exists' % lv_snapshot_name)
-        self.mount_loc = self.srcdir
+        self.mount_loc = self.wordir
```
harish-24: Here too, `workdir` is already used in test.

clebergnu:
It was a mass change, so the same mistake is in all changes. I'll fix them.
Thanks!

fs/filebench.py
clebergnu: True.  The test actually PASSes like this, because the installation will just create the directories and files on top of the source directory, but you're right, it makes sense to use different directories.

clebergnu:
Closing in favor of #819

PR #821, opened 2018-03-15
Fixed generic:sysbench test

PraveenPenguin:
as in powerpc64 bit system compilation of upstream branch is failing
so this patch address that issue as try to clone ppc64 specfic LuaJIT upstream

based on issue : https://github.com/akopytov/sysbench/issues/129

### DIFF
--- a/generic/sysbench.py
+++ b/generic/sysbench.py
@@ -15,6 +15,7 @@
 # Author: Hariharan T.S.  <harihare@in.ibm.com>
 
 import os
+import shutil
 from avocado import Test
 from avocado import main
 from avocado.utils import process, git
@@ -64,9 +65,17 @@ class Sysbench(Test):
                 self.urllink = self.params.get(
                     'url-link', default="https://github.com/akopytov/"
                                         "sysbench.git")
+                self.fixlink = self.params.get('fixlink', default=None)
+                self.fix_dir = self.params.get('fixdir', default=None)
                 self.bch = self.params.get('branch', default='master')
                 git.get_repo(self.urllink, branch=self.bch,
                              destination_dir=self.teststmpdir)
+
+                if self.fixlink and self.fix_dir:
+                    fixpath = '%s%s' % (self.teststmpdir, self.fix_dir)
+                    shutil.rmtree(fixpath)
+                    git.get_repo(self.fixlink, branch="ppc64-port",
+                                 destination_dir=fixpath)
                 os.chdir(self.teststmpdir)
                 self.run_cmd("./autogen.sh")
                 self.run_cmd("./configure --without-mysql")
--- a/generic/sysbench.py.data/sysbench.yaml
+++ b/generic/sysbench.py.data/sysbench.yaml
@@ -1,6 +1,8 @@
 max-time: null
 max-request: null
 url-link: 'https://github.com/akopytov/sysbench.git'
+fixlink: 'https://github.com/PPC64/LuaJIT.git'
+fixdir: '/third_party/luajit/luajit/'
 #Comment out branch if you want to checkout master.
 branch: '0.4'
 tests:


generic/sysbench.py
```diff
@@ -67,6 +68,11 @@ def setUp(self):
                 self.bch = self.params.get('branch', default='master')
                 git.get_repo(self.urllink, branch=self.bch,
                              destination_dir=self.teststmpdir)
+                if self.urllink == 'https://github.com/akopytov/sysbench.git':
```
harihare: Instead of  urllink check can you check for ppc as the fix is only for ppc. 
if "ppc"  in distro.detect().arch:
then download the luajit for ppc

PraveenPenguin:
@harihare thanks for review adopted the change please have a look , @abdhaleegit can you also please verify this

generic/sysbench.py
PraveenPenguin: @narasimhan-v  not sure it temporary as akopytov is working on . And this need for ppc system in as x86 no issue as `vm_ppc64.dasc` broken in ppc arch , so i adopted hari's suggestion

PraveenPenguin: @harihare i think we need to have both condition as if any other url passed from ymal file it will broke

generic/sysbench.py
```diff
@@ -67,6 +68,12 @@ def setUp(self):
                 self.bch = self.params.get('branch', default='master')
                 git.get_repo(self.urllink, branch=self.bch,
                              destination_dir=self.teststmpdir)
+                if self.urllink == 'https://github.com/akopytov/sysbench.git' \
+                        and "ppc" in distro.detect().arch:
```
harihare: self.urllink == 'https://github.com/akopytov/sysbench.git'  this check is not required.

PraveenPenguin: @harihare i think it required as we pass different url from yaml file than it will fail , i already replied this on earlier :)

PraveenPenguin:
@narasimhan-v can you please have look

generic/sysbench.py
```diff
@@ -64,9 +65,16 @@ def setUp(self):
                 self.urllink = self.params.get(
                     'url-link', default="https://github.com/akopytov/"
                                         "sysbench.git")
+                self.fixlink = self.params.get('fixlink', default=None)
                 self.bch = self.params.get('branch', default='master')
                 git.get_repo(self.urllink, branch=self.bch,
                              destination_dir=self.teststmpdir)
+
+                if self.fixlink:
+                    luajiturl = '%s/third_party/luajit/luajit/' % self.teststmpdir
```
narasimhan-v: how about this:
```
self.fix_dir = self.params.get('fixdir', default=None)
...
if self.fixlink and self.fix_dir:
    fixdir = os.path.join(self.teststmpdir, self.fix_dir)
    shutil.rmtree(fixdir)
    git.get_repo(self.fixlink, branch="ppc64-port", destination_dir=lfixdir)
```

PraveenPenguin: @narasimhan-v `self.fix_dir` i think we need not to make it as variable as it is always constant and later operation depend on only this path so i think it should  be fine

narasimhan-v: @PraveenPenguin When fix link comes from yaml, then fix dir also should be so..

PraveenPenguin:
@narasimhan-v adopted the change can you please have a look

generic/sysbench.py
```diff
@@ -64,9 +65,17 @@ def setUp(self):
...
                     'url-link', default="https://github.com/akopytov/"
                                         "sysbench.git")
+                self.fixlink = self.params.get('fixlink', default=None)
+                self.fix_dir = self.params.get('fixdir', default=None)
                 self.bch = self.params.get('branch', default='master')
                 git.get_repo(self.urllink, branch=self.bch,
                              destination_dir=self.teststmpdir)
+
+                if self.fixlink and self.fix_dir:
+                    luajiturl = '%s%s' % (self.teststmpdir, self.fix_dir)
```
narasimhan-v: Very minor, but I would name this variable something generic.
Let me know if you can quickly do it, else I am fine with merging this one.

PraveenPenguin:
@narasimhan-v done

narasimhan-v:
@PraveenPenguin Thanks

PR #823, opened 2018-03-16
Script to add and delete Network Virtualisation interfaces

hathyaga:
This script will add and delete Network virtualized interfaces
using the SRIOV capable IO adapter given as an input by the user
and performs this operation multiple times based on the count
given by the user as well

### DIFF
--- a/io/net/Network_virtualization_dlpar.py
+++ b/io/net/Network_virtualization_dlpar.py
@@ -1,0 +1,233 @@
+#!/usr/bin/python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2018 IBM
+# Author: Harsha Thyagaraja <harshkid@linux.vnet.ibm.com>
+
+import os
+import shutil
+try:
+    import pxssh
+except ImportError:
+    from pexpect import pxssh
+from avocado import Test
+from avocado import main
+from avocado.utils import process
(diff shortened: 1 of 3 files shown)


hathyaga:
[job.log.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1818125/job.log.txt)

Job log

io/net/Network_virtualization_dlpar.py
```diff
@@ -0,0 +1,233 @@
...
+        return "Command '%s' exited with %d.\nOutput:\n%s" \
+               % (self.command, self.exitcode, self.output)
+
+
+class NetworkVirtualization(Test):
+    '''
+    Adding and deleting Network Virtualized ports from the vios
+    '''
+    @skipUnless("ppc" in distro.detect().arch,
+                "supported only on Power platform")
```
harish-24: Please add skipIf only once here. There is no need for adding it before `test()` function or any other function as we do not want to execute the `setup()` if test needs to be skipped.
```
@skipIf(IS_POWER_NV or IS_KVM_GUEST, 
        "This test is not supported on KVM guest or PowerNV platform")
```

harish-24: Note that Skip message in the above comment is moved to next line as the previous implementation seemed to have caused unnecessary spaces.
```
 TestSkipError: This test is not supported on KVM         guest or PowerNV platform
```

io/net/Network_virtualization_dlpar.py
```diff
@@ -0,0 +1,233 @@
...
+        except CmdError as details:
+            self.log.debug(str(details))
+            self.fail("Starting service %s failed", svc)
+
+        output = process.system_output("lssrc -a", ignore_status=True,
+                                       shell=True, sudo=True)
+        if "inoperative" in output:
+            self.fail("Failed to start the rsct and rsct_rm services")
+
+        for val in range(self.count):
```
harish-24: If `val` is not used, we may replace as follows
```
for _ in range(self.count):
```

hathyaga:
New PR: https://github.com/avocado-framework-tests/avocado-misc-tests/pull/826

PR #831, opened 2018-03-22
Test for memory check using md5sum

harish-24:
Test creates a chunk of RAM size and checks its md5sum multiple times so that memory integrity is intact

### DIFF
--- a/memory/sum_check.py
+++ b/memory/sum_check.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2018 IBM
+# Author: Harish <harish@linux.vnet.ibm.com>
+#
+
+import os
+from avocado import Test
+from avocado import main
+from avocado.utils import process, memory, disk, genio
+
+
+class SumCheck(Test):
+    """
+    Test allocates file chuck of RAM size and checks for md5sum of the file
+    repetitively so that memory integrity persists
+    """
+
+    def setUp(self):
+        self.iter = int(self.params.get('iterations', default='5'))
+        self.memsize = int(self.params.get(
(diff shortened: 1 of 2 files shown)


memory/sum_check.py
```diff
@@ -0,0 +1,61 @@
...
+            mdsum.append(process.system_output('md5sum %s' % self.ddfile))
+            self.log.info("MD5 : %s", mdsum[i - 1])
+
+        if len(set(mdsum)) > 1:
+            self.fail('Md5sum for created file differs')
+
+    def tearDown(self):
+        process.system("echo 3 > /proc/sys/vm/drop_caches",
+                       shell=True, ignore_status=True)
+        os.remove(self.ddfile)
```
narasimhan-v: Doesn't the self.workdir get deleted automatically ?

memory/sum_check.py
```diff
@@ -0,0 +1,61 @@
...
+            self.fail("Chunk creation failed due to %s" % details)
+        for i in range(self.iter):
+            mdsum.append(process.system_output('md5sum %s' % self.ddfile))
+            self.log.info("MD5 : %s", mdsum[i - 1])
+
+        if len(set(mdsum)) > 1:
+            self.fail('Md5sum for created file differs')
+
+    def tearDown(self):
+        process.system("echo 3 > /proc/sys/vm/drop_caches",
```
narasimhan-v: `genio.write_file()`

memory/sum_check.py
```diff
@@ -0,0 +1,61 @@
...
+
+        self.log.info("Creating chunk with dd")
+        try:
+            process.system('dd if=/dev/urandom of=%s bs=%s count=1024' %
+                           (self.ddfile, self.memsize))
+        except process.CmdError as details:
+            self.fail("Chunk creation failed due to %s" % details)
+        for i in range(self.iter):
+            mdsum.append(process.system_output('md5sum %s' % self.ddfile))
+            self.log.info("MD5 : %s", mdsum[i - 1])
```
narasimhan-v: Why `i - 1` ?

memory/sum_check.py
harish-24: Oh, missed that. Will remove.

memory/sum_check.py
harish-24: Got confused with range(). Right, `i` would be fine.

memory/sum_check.py
harish-24: sure, will adapt.

harish-24:
@narasimhan-v Thanks for the review. Changes are made.

PraveenPenguin:
@harish-24 thanks for patch , @narasimhan-v thanks for ACK

PR #828, opened 2018-03-20
Script to unbind and bind the Network virtualized device

hathyaga:
This script will unbind and bind the driver of the
Network virtualized device based on the count given by
the user

Signed-off-by: Harsha Thyagaraja <harshkid@linux.vnet.ibm.com>

PR #836, opened 2018-03-22
Client initiated failover for a Network virtualized device

hathyaga:
This script performs a Client initiated failover for a
Network virtualized device. It also has the option to
run this execution any number of times based on user
input

### DIFF
--- a/io/net/virt-net/Network_virtualization_client_initialized_failover.py
+++ b/io/net/virt-net/Network_virtualization_client_initialized_failover.py
@@ -1,0 +1,77 @@
+#!/usr/bin/python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2018 IBM
+# Author: Harsha Thyagaraja <harshkid@linux.vnet.ibm.com>
+
+import time
+import netifaces
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+from avocado.utils.software_manager import SoftwareManager
+from avocado.utils.process import CmdError
+
+
+class NetworkVirtualizationFailoverTest(Test):
+
+    """
+    This test performs failover for a Network virtualized device
+    """
+
+    def setUp(self):
+        """
+        Install necessary packages and identify the network
+        virtualized device.
+        """
+        smm = SoftwareManager()
+        for pkg in ["net-tools"]:
+            if not smm.check_installed(pkg) and not smm.install(pkg):
+                self.cancel("%s package is need to test" % pkg)
+        interfaces = netifaces.interfaces()
+        self.interface = self.params.get('interface')
+        if self.interface not in interfaces:
+            self.cancel("%s interface is not available" % self.interface)
+        self.device = process.system_output("ls -l /sys/class/net/ | \
+                                             grep %s | cut -d '/' -f \
+                                             5" % self.interface,
+                                            shell=True).strip()
+        self.count = int(self.params.get('count', default="1"))
+        self.peer_ip = self.params.get('peer_ip', default=None)
+
+    def test(self):
+        '''
+        Performs failover for Network virtualized device
+        '''
+        try:
+            for val in range(self.count):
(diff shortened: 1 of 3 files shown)


hathyaga:
Job log 
[job.log.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1837153/job.log.txt)

io/net/virt-net/Network_virtualization_client_initialized_failover.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class NetworkVirtualizationFailoverTest(Test):
+
+    """
+    This test performs failover for a Network virtualized device
+    """
+
+    def setUp(self):
+        """
```
narasimhan-v: setUp() is almost same as #828
Can you make them into a single python script with 2 tests ?

io/net/virt-net/Network_virtualization_client_initialized_failover.py.data/README
```diff
@@ -0,0 +1,11 @@
+This script will perform Client initiated failover for
+a Network virtualized device for as many times as specified by the user
+
+Before starting the test, make sure that a Backing device is added to the
+Network virtualized device under test
+
+Inputs Needed (in multiplexer file):
+------------------------------------
+interface - The Network virtualized device
+count - The number of times the unbind and bind test has to be executed
```
narasimhan-v: copied and did not change ?

io/net/virt-net/Network_virtualization_client_initialized_failover.py
```diff
@@ -0,0 +1,77 @@
...
+        '''
+        Performs failover for Network virtualized device
+        '''
+        try:
+            for val in range(self.count):
+                self.log.info("Performing Client initiated\
+                              failover - Attempt %s" % int(val+1))
+                process.run('echo 1 > /sys/devices/vio/%s/failover' %
+                            self.device, shell=True, sudo=True)
+                time.sleep(10)
```
narasimhan-v: Is it always 10s ?

io/net/virt-net/Network_virtualization_client_initialized_failover.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def test(self):
+        '''
+        Performs failover for Network virtualized device
+        '''
+        try:
+            for val in range(self.count):
+                self.log.info("Performing Client initiated\
+                              failover - Attempt %s" % int(val+1))
+                process.run('echo 1 > /sys/devices/vio/%s/failover' %
```
narasimhan-v: use `genio.write_file()`

io/net/virt-net/Network_virtualization_client_initialized_failover.py
hathyaga: I am reluctant on including both of them in a single script. 
I appreciate the thought of adding both tests as a single script but I strongly recommend not to judge it that way.  
Failover functionality might change as it is still in Tech Preview. For now, this is how the functionality is coded I suggest we take both those PR requests separately 

i would keep the tests separate with different scripts based on the functionality. :)

io/net/virt-net/Network_virtualization_client_initialized_failover.py.data/README
hathyaga: Oops, my bad. Will take care of that.

io/net/virt-net/Network_virtualization_client_initialized_failover.py
hathyaga: Well, it.s less than 10 but definitely above 5. Chose 10 seconds just so it looks visually pleasing.

narasimhan-v: Is there any way to find if failover is complete ?

io/net/virt-net/Network_virtualization_client_initialized_failover.py
PraveenPenguin: i also think so it will be good if both test can be part of single script and maintenance  will be easy what i feel @hathyaga i think if Failover functionality change any how we need to change test script accordingly so if it as just cover in one test definition and adoption on new changes will be easy , if you put in single script on in two test definition it not going to matter as avocado run and result publication  prospective any you are owner of test , if still feel it should be in two script should be fine with me :) , those are just suggestion

hathyaga: Sure @PraveenPenguin and @narasimhan-v I'll include this test in the already existing network_virtualization_dlpar.py script. 

Thanks for the comments.

narasimhan-v:
Closing this one, assuming these are taken care via other PRs.
Re-open if not.

PR #849, opened 2018-04-05
Test to verify diag_encl.

Pavithra1602:
This test verifies '-d' option of diag_encl. It checks whether single xml file is generated when "diag_encl" command is run multiple times and validates xml content.


Pavithra1602:
raised new PR.

PR #874, opened 2018-05-02
Clear dmesg before starting test

harish-24:
Test reports failure as dmesg prior to the start of tests has some keywords which is not cleared before starting the test

### DIFF
--- a/memory/memhotplug.py
+++ b/memory/memhotplug.py
@@ -37,6 +37,10 @@ errorlog = ['WARNING: CPU:', 'Oops',
             'double fault:', 'BUG: Bad page state in']
 
 
+def clear_dmesg():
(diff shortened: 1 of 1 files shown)


memory/memhotplug.py
```diff
@@ -37,6 +37,10 @@
             'double fault:', 'BUG: Bad page state in']
 
 
+def clear_dmesg():
+    process.run("dmesg -c ", sudo=True)
+
```
PraveenPenguin: can you remove this white space

harish-24: Added it due to pep8 issue.

PraveenPenguin: ah sure

PR #878, opened 2018-05-07
Added flail test

PraveenPenguin:
added flail test which does system call fuzz


fs/flail.py
```diff
@@ -0,0 +1,67 @@
...
+    def setUp(self):
+        '''
+        Setup Flail
+        '''
+        smm = SoftwareManager()
+        for package in ['gcc', 'make']:
+            if not smm.check_installed(package) and not smm.install(package):
+                self.cancel(package + ' is needed for the test to be run')
+
+        tarball = os.path.join(self.datadir, "flail-0.2.0.tar.gz")
```
narasimhan-v: Is flail not maintained in upstream ?

fs/flail.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def test(self):
+        '''
+        Runs flail with the appropriate parameters.
+
+        :param fstype: Filesystem type there user want to run flail
+        '''
+        self.clear_dmesg()
+        os.chdir(self.build_dir)
+        fs_type = self.params.get('fstype', default='xfs')
```
narasimhan-v: Better to read yaml in setUp()

fs/flail.py
```diff
@@ -0,0 +1,67 @@
...
+        process.system('./flail %s 1' % fs_type)
+        dmesg = process.system_output('dmesg')
+        match = re.search(r'Call Trace:', dmesg, re.M | re.I)
+        if match:
+            self.log.info("some call traces seen please check")
```
narasimhan-v: I think we can FAIL, or at least WARN, in such cases. Your thoughts ?

fs/flail.py
PraveenPenguin: @narasimhan-v no we got this test from internal

fs/flail.py
PraveenPenguin: sure will do

fs/flail.py
PraveenPenguin: @narasimhan-v sure i miss that will change to `self.fail` thanks for catching this

fs/flail.py
narasimhan-v: @PraveenPenguin my suggestion would be to have it in upstream (maybe create a git if owner is fine). If that does not work out, then we can have tarball here.

fs/flail.py
```diff
@@ -0,0 +1,67 @@
...
+    def test(self):
+        '''
+        Runs flail with the appropriate parameters.
+
+        :param fstype: Filesystem type there user want to run flail
+        '''
+        self.clear_dmesg()
+        os.chdir(self.build_dir)
+        fs_type = self.params.get('fstype', default='xfs')
+        process.system('./flail %s 1' % fs_type)
```
harish-24: What does this second input `1` signify to? If it is significant, better to move to yaml.

harish-24: And if the command fails, test would error. Better to catch it or use `ignore_status` if we can skip it.

fs/flail.py
narasimhan-v: @PraveenPenguin We can contact the maintainer, and if no response, then we can push this commit.
How long can we wait for the maintainer's response ? any suggestions ?

PraveenPenguin: @narasimhan-v i think we can have it here only as i did not get any response from maintainer , will send V2 as soon as possible

PraveenPenguin:
@narasimhan-v done with changes can you please have a look

fs/flail.py
narasimhan-v: @PraveenPenguin any conclusion here ?

narasimhan-v:
@PraveenPenguin Thanks for the patch, @harish-24 Thanks for the review.

PR #879, opened 2018-05-09
fix: add unicode param when file is read as string

abdhaleegit:
The /tmp/tmpfile is read as string and not converted to unicode on python2.7
so test is failing with this error:

UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 18: ordinal

fixed to decode from str to unicode(UTF-8)

### DIFF
--- a/io/net/scp_ssh.py
+++ b/io/net/scp_ssh.py
@@ -115,11 +115,11 @@ class ScpTest(Test):
         '''
         self.log.info('removing data')
         cmd = "rm /tmp/tempfile"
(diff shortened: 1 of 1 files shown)


abdhaleegit:
[beforelogs.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1986537/beforelogs.txt)

[afterlogs.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/1986538/afterlogs.txt)

PR #889, opened 2018-05-16
set the StrictHostKeyChecking flag for smooth ssh

abdhaleegit:
ssh command is hung when the ssh keys are not in known_host file
so set the flag to 'no' that will add the keys to know hosts automatically
so which allows test to complete

### DIFF
--- a/io/net/scp_ssh.py
+++ b/io/net/scp_ssh.py
@@ -84,7 +84,7 @@ class ScpTest(Test):
         '''
         check scp and ssh
         '''
-        cmd = "ssh %s@%s \"echo hi\"" % (self.user, self.peer)
+        cmd = "ssh -o StrictHostKeyChecking=no %s@%s \"echo hi\"" % (self.user, self.peer)
         if process.system(cmd, shell=True, ignore_status=True) != 0:
             self.fail("unable to ssh into peer machine")
         process.run("dd if=/dev/zero of=/tmp/tempfile bs=1024000000 count=1",
@@ -118,7 +118,7 @@ class ScpTest(Test):
         if process.system(cmd, shell=True, ignore_status=True, encoding="utf-8") != 0:
             self.log.info("unable to remove data from client")
         msg = "rm /tmp/tempfile"
-        cmd = "ssh %s@%s \"%s\"" % (self.user, self.peer, msg)
+        cmd = "ssh -o StrictHostKeyChecking=no %s@%s \"%s\"" % (self.user, self.peer, msg)
         if process.system(cmd, shell=True, ignore_status=True, encoding="utf-8") != 0:
             self.log.info("unable to remove data from peer")
 


io/net/scp_ssh.py
```diff
@@ -84,7 +84,7 @@ def test_scpandssh(self):
         '''
         check scp and ssh
         '''
-        cmd = "ssh %s@%s \"echo hi\"" % (self.user, self.peer)
+        cmd = "ssh -o StrictHostKeyChecking=no %s@%s \"echo hi\"" % (self.user, self.peer)
```
narasimhan-v: @abdhaleegit Readme says ssh key to peer needs to be generated. Is that not enough ?
https://github.com/abdhaleegit/avocado-misc-tests/blob/241858f987421fb923349977686363a858a372c6/io/net/scp_ssh.py.data/README.txt

abdhaleegit: @narasimhan-v  Ah ok it is not enough, thinking of including passwordless ssh setup code in library as it is used in many tests,  instead of manual setting, will that be usefull ?

PraveenPenguin: @abdhaleegit not sure we can target this to lib as it a user choice to do that so README i think should be fine

narasimhan-v: @abdhaleegit If that can be handled in each of the tests, then I am fine with that as well.

narasimhan-v: @abdhaleegit But cleaner way is to use a python lib or write an avocado lib :)

abdhaleegit: In my system even after setting passwordless ssh, still I see the test stop here in some case:

Running 'ssh root@140.1.1.122 "rm /tmp/tempfile"'
The authenticity of host '140.1.1.122 (140.1.1.122)' can't be established.
ECDSA key fingerprint is SHA256:sOcszqO0aqSOhaqll0ez29+70StJTRPtTlUX19blPE0.
Are you sure you want to continue connecting (yes/no)?

so when I enter 'yes' the test continued with out password, this was the reason for this patch. 

Having a avocado lib is not a immediate requirement. so can we push this patch for now ?

narasimhan-v: @abdhaleegit Can you please check if passwordless ssh was set using this IP address/hostname ?
Because, we usually send the keys via one IP and try with another IP, which asks for confirmation first time.

If this is reproduced with same IP, then we can look into it.

abdhaleegit: the public keys were copied not with the same IP, but by default IP. 

On the utility part I am planning to add  passwd_less_ssh() to utils/network.py by making use of pexpect library  

This will solve scpssh and bonding test issue, we can close this

abdhaleegit:
this will be required, so that we do not have to manually add ssh keys in to known_host file every time

PraveenPenguin:
closing this PR due to in activeness , @abdhaleegit if you still want to pursue  this  PR please feel free to re open

PR #891, opened 2018-05-16
Patch to fix -notime option failure.

Pavithra1602:
lshw output should not contain modified timestamp when -notime option is used.


PraveenPenguin:
@Pavithra1602 thanks for patch

PR #896, opened 2018-05-18
Fix: service_check test

harish-24:
1) Fix: service_check syntax 
Test failed with the following error as it was trying to remove a list instead of a string

L0047 ERROR|     services_list.remove(['networking'])
L0047 ERROR| ValueError: list.remove(x): x not in list

2) Fix: newly changed SuSE service 
Patch fixes newly changed firewall service in SuSE versions higher
than 15.

3) Fix: Packages for SuSE services
Added packages for SuSE service failures as pre-requisite

### DIFF
--- a/generic/service_check.py.data/services.cfg
+++ b/generic/service_check.py.data/services.cfg
@@ -3,7 +3,7 @@ services = sshd,ufw,networking,libvirtd
 [rhel]
 services = sshd,firewalld,network,libvirtd
 [SuSE]
-services = sshd,SuSEfirewall2,network,libvirtd
+services = sshd,network,libvirtd
 [fedora]
 services = sshd
 [centos]
--- a/generic/service_check.py
+++ b/generic/service_check.py
@@ -35,6 +35,11 @@ class service_check(Test):
         config_file = self.datadir + '/services.cfg'
         parser.read(config_file)
         services_list = parser.get(detected_distro.name, 'services').split(',')
+        if detected_distro.name == 'SuSE':
+            if detected_distro.version >= 15:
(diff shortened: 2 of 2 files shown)


generic/service_check.py.data/services.cfg
```diff
@@ -3,7 +3,7 @@ services = sshd,ufw,networking,libvirtd
 [rhel]
 services = sshd,firewalld,network,libvirtd
 [SuSE]
-services = sshd,SuSEfirewall2,network,libvirtd
+services = sshd,network,libvirtd
```
PraveenPenguin: can we just depend on system service using chkconfig not depend on cfg file

harish-24: @PraveenPenguin chkconfig shows only SysV services and does not include distro specific services. Will `systemctl list-unit-files` help here? If so, we may have to make this cfg file obsolete.

PraveenPenguin: sure that will help but we need to have support for older distro as well so i think both need to be used

harish-24: @PraveenPenguin Sure, as discussed we can make this approach an enhancement for the test in future.

PraveenPenguin: sure , will merge

PraveenPenguin:
@narasimhan-v please your thought  on this PR

narasimhan-v:
@harish-24 As per the discussion, the enhancement as an avocado utilities can be tried.
@PraveenPenguin for now, we can merge this PR.

PR #906, opened 2018-05-28
Vnic count default

narasimhan-v:


io/net/virt-net/network_virtualization.py
```diff
@@ -111,8 +111,9 @@ def setUp(self):
         self.backing_adapter_id = self.params.get("backing_adapter_id",
                                                   '*', default=None)
         self.bandwidth = self.params.get("bandwidth", '*', default=None)
-        self.count = int(self.params.get('count', default="1"))
+        self.count = int(self.params.get('vnic_test_count', default="1"))
         self.device_ip = self.params.get('device_ip', '*', default=None)
+        self.mac_id = self.params.get('mac_id', default="02:03:03:03:03:01")
```
PraveenPenguin: can we have check about if mac_id not present in system test should cancel ?

narasimhan-v: @PraveenPenguin There are many parameters like `device_ip` which need to be checked.. I will address that in a following commit.

narasimhan-v: Thanks

PR #908, opened 2018-05-30
Add check for numa tests

harish-24:
Patch adds pre-check for few tests which requires NUMA memory nodes with newer avocado API

### DIFF
--- a/cpu/numactl.py
+++ b/cpu/numactl.py
@@ -20,7 +20,7 @@ import os
 
 from avocado import Test
 from avocado import main
-from avocado.utils import archive, build, process, distro
+from avocado.utils import archive, build, process, distro, memory
 from avocado.utils.software_manager import SoftwareManager
 
 
@@ -38,7 +38,6 @@ class Numactl(Test):
         Source:
         https://github.com/numactl/numactl
         '''
-
         # Check for basic utilities
         smm = SoftwareManager()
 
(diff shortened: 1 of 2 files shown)


cpu/numactl.py
```diff
@@ -38,6 +38,9 @@ def setUp(self):
         Source:
         https://github.com/numactl/numactl
         '''
+        self.add_warn = False
+        if len(memory.numa_nodes_with_memory()) < 2:
```
narasimhan-v: Can this check be done just before the `self.log.warn()` ?

harish-24: We can, but is it a suggestion to have a local variable rather than a `self` attribute? Otherwise it is more aligned to be at `setUp` phase.

narasimhan-v: Yes, local check is suggested, and variable's use can also be removed.

memory/numa_test.py
```diff
@@ -43,8 +43,10 @@ def setUp(self):
             'nr_pages', default=memsize / memory.get_page_size())
         self.map_type = self.params.get('map_type', default='private')
 
-        if len(memory.numa_nodes()) < 2:
-            self.cancel('Test requires two numa nodes to run')
+        nodes = memory.numa_nodes_with_memory()
+        if len(nodes) < 2:
```
narasimhan-v: `len(memory.numa_nodes_with_memory())`

harish-24: It was used as if the condition is met, same is displayed in message at https://github.com/avocado-framework-tests/avocado-misc-tests/pull/908/files#diff-082c2aa8ea45734f25c0147833e7a73bR49.

cpu/numactl.py
harish-24: sure, will make the change.

memory/numa_test.py
narasimhan-v: Ok, got it. Missed the usage in the next line

PR #911, opened 2018-06-01
Fix the raw code for P9 generic events

shriyak:
The generic events on P8 and P9 differ in raw code and hence the event name remains , only raw code changes.
This fix will update the raw code for P9 events like LLC-stores,cache-references,L1-dcache-loads,L1-dcache-load-misses etc.
Example for P8 'LLC-stores' the raw code is : 0x17080 and for P9 it is :0x16880
The raw code value for each event is obtained under sysfs directory : /sys/bus/event_source/devices/cpu/events
for P8 and P9.

### DIFF
--- a/perf/raw_code.cfg
+++ b/perf/raw_code.cfg
@@ -1,0 +1,52 @@
+[POWER8]
+cpu-cycles = 0x1e
+stalled-cycles-frontend = 0x100f8
+stalled-cycles-backend = 0x4000a
+instructions = 0x02
+branch-instructions = 0x10068
+branch-misses = 0x400f6
+cache-references = 0x100ee
+cache-misses = 0x3e054
+L1-dcache-load-misses = 0x3e054
+L1-dcache-loads = 0x100ee
+L1-dcache-prefetches = 0xd8b8
+L1-dcache-store-misses = 0x300f0
+L1-icache-load-misses = 0x200fd
+L1-icache-loads = 0x4080
+L1-icache-prefetches = 0x408e
+LLC-load-misses = 0x300fe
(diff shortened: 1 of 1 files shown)


perf/perf_genericevents.py
```diff
@@ -46,7 +48,34 @@ def generic_events(x):
...
+        'LLC-load-misses': "0x300fe",
+        'LLC-loads': "0x4c042",
+        'LLC-prefetches': "0x4e052",
+        'LLC-store-misses': "0x26880",
+        'LLC-stores': "0x16880",
+        'branch-load-misses': "0x400f6",
+        'branch-loads': "0x4d05e",
+        'dTLB-load-misses': "0x300fc",
+        'iTLB-load-misses': "0x400fc"
+    }}
```
PraveenPenguin: can these dict generated  from a cfg file as i think editing test code in not good

PraveenPenguin:
@nasastry can you please have a look in this PR as i not able to link the changes

kamalesh-babulal:
looks like the fix is incomplete, I am assuming the `raw_code.cfg` is suppose to be with `perf_genericevents.py`. I can work on it. @harish-24 This is perf specific and it should be under `perf` directory.

PraveenPenguin:
@kamalesh-babulal please close older PR as already new PR in place , that will really helpful to reviewer

kamalesh-babulal:
@PraveenPenguin Thanks, I was treating it like an issue where we could have closed it once the other PR is merged.

PraveenPenguin:
@kamalesh-babulal sure that also make sense , but it kind of confusing any way should be fine with me

PR #918, opened 2018-06-12
Added check for NVM Subsystem Reset

narasimhan-v:
NVM Subsystem Reset is not supported in all nvme controllers.
So, added a check and skipped the test if not supported.


narasimhan-v:
```
# nvme show-regs /dev/nvme0 -H  | grep "NVM Subsystem Reset Supported"
	NVM Subsystem Reset Supported  (NSSRS): Yes
```

PR #920, opened 2018-06-12
perf:lmbench Add parameter for memory value

nasastry:
Add parameter to pass memory value to the lmbench

Signed-off-by: Nageswara R Sastry <rnsastry@linux.vnet.ibm.com>

### DIFF
--- a/io/disk/multipath_test.py
+++ b/io/disk/multipath_test.py
@@ -135,33 +135,33 @@ class MultipathTest(Test):
             self.log.info(" Failing and reinstating the individual paths")
             for path in path_dic["paths"]:
                 if multipath.fail_path(path) is False:
-                    msg += "test failed while failing %s" % path
+                    msg += "test failed while failing %s\n" % path
                 elif multipath.reinstate_path(path) is False:
-                    msg += "test failed while reinstating %s" % path
+                    msg += "test failed while reinstating %s\n" % path
             self.mpath_svc.restart()
 
             # Failing n-1 paths for short time and reinstating back
             self.log.info("Failing and reinstating the n-1 paths")
             for path in path_dic['paths'][:-1]:
                 if multipath.fail_path(path) is False:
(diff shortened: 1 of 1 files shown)


perf/lmbench.py
```diff
@@ -101,6 +102,8 @@ def setUp(self):
                        (fsdir, config_file, tmp_config_file))
         process.system("sed 's!^FILE=.*$!FILE=%s!' '%s'  '%s'" %
                        (temp_file, tmp_config_file, config_file))
+        process.system("sed 's!^MB=.*$!MB=%s!' '%s'  '%s'" %
```
harish-24: As this repeats for some params, can we have a small `replace` function to do the same?

harish-24: Better if that can be targeted towards avocado utility too, IMO, would be very useful. @narasimhan-v Any suggestions on this.

narasimhan-v: 1. This is better if done natively in python. Something like below:
```
with open(config_file, "r") as in_file:
    with open(tmp_config_file, "w") as out_file:
        for line in in_file:
            out_file.write(line.replace('old', 'new'))
```

2. Whenever some code is repeating, better to have a loop / function. If that serves multiple tests, then utility. I think local function would do here.

harish-24: @nasastry The first change is from `config_file` to `tmp_config_file`. The other two seems to be opposite of it. Is this right?

nasastry: @harish-24 
`sed` tool changes they are capturing in to a temporary file and then replace original file with the temporary one. This is common for `sed` related operation.
@narasimhan-v and @harish-24 
Let me try a way to replace `sed` commands with `python` code. Will let you know asap.

nasastry:
@narasimhan-v @harish-24 Not understood why it suddenly showing closed and merged. I just sent the updated code. Seems there is a problem. I will create one more PR.

PR #921, opened 2018-06-12
perf:lmbench Add parameter for memory value

nasastry:
Add parameter to pass memory value to the lmbench

Signed-off-by: Nageswara R Sastry <rnsastry@linux.vnet.ibm.com>


perf/lmbench.py
```diff
@@ -95,12 +96,11 @@ def setUp(self):
 
         # patch the resulted config to use the proper temporary directory and
         # file locations
-        tmp_config_file = config_file + '.tmp'
-        process.system('touch ' + tmp_config_file)
-        process.system("sed 's!^FSDIR=.*$!FSDIR=%s!' '%s'  '%s' " %
-                       (fsdir, config_file, tmp_config_file))
-        process.system("sed 's!^FILE=.*$!FILE=%s!' '%s'  '%s'" %
-                       (temp_file, tmp_config_file, config_file))
+        process.system("sed -i -e 's!^FSDIR=.*$!FSDIR=%s!' "
```
narasimhan-v: @nasastry I would still prefer using python file operations to do it.

nasastry: @narasimhan-v 
To replace three variables `sed` is the easy way. In the above command I am not using any temporary files and it's just a single command.
In python need to open two files, In a loop compare each and every line with the required variables to change, replace them, if there are multiple matches need to handle them, handle the file related exceptions.

harish-24: @nasastry Yes, but still the processing time for shell commands running through python is higher than python native operations. We may need to consider that.

nasastry:
@narasimhan-v @harish-24 
I have gone through both `sed` implementation and `python way of doing`. I didn't see any huge difference. The actual difference is `0.01 seconds`. 
Any way sent PR to python implementation.

In my view we can better concentrate on real issues rather than some thing like this.
If python gives better performance over shell commands, we shouldn't be using shell commands at all. But it's not the case. There are many shell commands used in these test scripts.

Thanks for the review.

narasimhan-v:
@nasastry Thanks for the patch.

The way I see it is: If something can be done in native language, it is always preferred that way. Here, shell should be used only when things can not be done in python.

The idea is to improve the coding standards everyday.
Yes, there were shortcomings in the past, and would be happy to rectify it going forward.
Right now, whenever such a line is touched in a commit, we just ask for the change. We need to spend more time cleaning code, time permitting :-)

nasastry:
@narasimhan-v 
If there are no issues with the current patch and no further comments Please merge. Thanks!!

perf/lmbench.py
```diff
@@ -95,12 +96,20 @@ def setUp(self):
...
+            for line in lines:
+                if line.startswith("FSDIR="):
+                    cfg_file.write("FSDIR=%s\n" % fsdir)
+                elif line.startswith("FILE="):
+                    cfg_file.write("FILE=%s\n" % temp_file)
+                elif line.startswith("MB="):
+                    cfg_file.write("MB=%s\n" % memory_size_mb)
+                else:
+                    cfg_file.write(line)
+        process.system("cat %s" % config_file)
```
narasimhan-v: @nasastry I will happily merge if this is also handled python with file

nasastry:
@narasimhan-v Removed `cat` and provided python native code. Please help me in merging this PR. Thank you.

nasastry:
PASS 1-lmbench.py:Lmbench.test;fsdir-a5f8
PASS 2-lmbench.py:Lmbench.test;temp_file-be21

PR #939, opened 2018-06-25
Fix: Skip memory-less nodes

harish-24:
With memory-less nodes, the test tries to move to it. This patch fixes it by skipping these nodes.


harish-24:
Diff of the tar file.
```
@@ -165,7 +165,7 @@ void get_numa_nodes_to_use(unsigned long
 	/* Get 2 Nodes which contains 10% of total system memory*/
 	for(node_iterator=0; node_iterator <= max_node; node_iterator++){
 		node_size = numa_node_size(node_iterator,&free_node_sizes);
-		if (node_size != -1){
+		if (node_size > 0){
 			if ((free_node_sizes > memory_to_use) && (got_nodes <= 1)){
 				nodes_to_use[got_nodes++] = node_iterator;
 			}
```

PR #941, opened 2018-06-27
Move kdump test to kdump folder.

Pavithra1602:
Patch to create kdump directory and move kdump.py from generic to kdump.


PraveenPenguin:
@Pavithra1602 i think RAS directory will be good option as we can push all RAS test cases in that directory

PR #944, opened 2018-06-27
Fix: Integrity test cancel on single node systems

harish-24:
Patch cancels tests on machines with single nodes or nodes which has less memory.


harish-24:
Tar diff
```
@@ -177,7 +177,7 @@ void get_numa_nodes_to_use(unsigned long
 		printf("Nodes used in test %d %d \n", nodes_to_use[0], nodes_to_use[1]);
 	}else {
 		printf("10 percent of total memory is not found in 2 nodes\n");
-		exit(0);
+		exit(255);
 	}
 }
 
@@ -291,7 +291,7 @@ void write_read_pattern_numa_migration()
         printf("\nScenario : Numa Migration \n\n");
 	if(numa_available == -1){
 		printf("Numa library is not present");
-		exit(-1);
+		exit(255);
 	}
 	max_node = numa_max_node();
 
@@ -374,7 +374,7 @@ int main(int argc, char *argv[])
 	printf("Total memory size %lu bytes \n", total_mem);
 	if (argc < 2){
 		printf("Usage <execname> -s <scenario_no> \n");
-		exit(-1);
+		exit(255);
 	}
 	option = getopt(argc, argv,"s:");
 	if (option != -1){
```

memory/integrity.py
```diff
@@ -63,9 +63,13 @@ def test(self):
...
         os.chdir(self.build_dir)
-        if process.system('./mem_integrity_test -s %s' % self.scenario_arg,
-                          shell=True, ignore_status=True) != 0:
-            self.fail("Test failed")
+        status = process.system('./mem_integrity_test -s %s' %
+                                self.scenario_arg, shell=True, ignore_status=True)
+        if status != 0:
+            if status == 255:
+                self.cancel("System does not have numa/memory to run the test")
+            else:
```
narasimhan-v: `else` is not needed here

harish-24:
Closing this with #947

PR #945, opened 2018-06-27
Fixed libhugetlbfs test

PraveenPenguin:
Fixed libhugetlbfs test 
1-change self.error to self.cancle
2-Removed  root privilege check and used avocado lib param for this
3- added ignore_status=True as a param in process.system to aviod
to error out script when command failed

### DIFF
--- a/memory/libhugetlbfs.py
+++ b/memory/libhugetlbfs.py
@@ -99,7 +99,7 @@ class libhugetlbfs(Test):
                 self.hugetlbfs_dir = os.path.join(self.tmpdir, 'hugetlbfs')
                 os.makedirs(self.hugetlbfs_dir)
             process.system('mount -t hugetlbfs none %s' %
-                           self.hugetlbfs_dir, sudo=True)
+                           self.hugetlbfs_dir, sudo=True, ignore_status=True)
 
         git.get_repo('https://github.com/libhugetlbfs/libhugetlbfs.git',
                      destination_dir=self.workdir)
@@ -191,7 +191,8 @@ class libhugetlbfs(Test):
 
     def tearDown(self):
         if self.hugetlbfs_dir:
-            process.system('umount %s' % self.hugetlbfs_dir)
+            process.system('umount %s' %
+                           self.hugetlbfs_dir, ignore_status=True)
 
 
 if __name__ == "__main__":
--- a/memory/libhugetlbfs.py
+++ b/memory/libhugetlbfs.py
@@ -68,8 +68,8 @@ class libhugetlbfs(Test):
             op = glob.glob("/usr/lib*/libpthread.a")
 
         if not op:
-            self.error("libpthread.a is required!!!"
-                       "\nTry installing glibc-static")
+            self.cancel("libpthread.a is required!!!"
+                        "\nTry installing glibc-static")
 
         # Get arguments:
         self.hugetlbfs_dir = self.params.get('hugetlbfs_dir', default=None)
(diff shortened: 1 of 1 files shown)


memory/libhugetlbfs.py
```diff
@@ -99,7 +99,7 @@ def setUp(self):
                 self.hugetlbfs_dir = os.path.join(self.tmpdir, 'hugetlbfs')
                 os.makedirs(self.hugetlbfs_dir)
             process.system('mount -t hugetlbfs none %s' %
-                           self.hugetlbfs_dir, sudo=True)
+                           self.hugetlbfs_dir, sudo=True, ignore_status=True)
```
harish-24: Test should not continue after this. We may need to cancel if this fails.

memory/libhugetlbfs.py
```diff
@@ -191,7 +191,8 @@ def test(self):
 
     def tearDown(self):
         if self.hugetlbfs_dir:
-            process.system('umount %s' % self.hugetlbfs_dir, sudo=True)
+            process.system('umount %s' % self.hugetlbfs_dir,
+                           sudo=True, ignore_status=True)
```
harish-24: May be a warning would be sufficient here.

memory/libhugetlbfs.py
```diff
@@ -83,11 +83,11 @@ def setUp(self):
                                                       verbose=False,
                                                       shell=True)
             if 'HugePages_' not in Hugepages_support:
-                self.error("No Hugepages Configured")
+                self.cancle("No Hugepages Configured")
```
harish-24: typo: `cancel`. And please correct in commit message too.

memory/libhugetlbfs.py
```diff
@@ -83,11 +83,11 @@ def setUp(self):
...
             memory.set_num_huge_pages(pages_requested)
             pages_available = memory.get_num_huge_pages()
         else:
-            self.error("Kernel does not support hugepages")
+            self.cancle("Kernel does not support hugepages")
```
harish-24: Here too.

memory/libhugetlbfs.py
PraveenPenguin: @harish-24 ah yes will take care

memory/libhugetlbfs.py
PraveenPenguin: no i think we need to come out from test if test is not supported so `cancle` will be good option

memory/libhugetlbfs.py
PraveenPenguin: are you saying we need to have a logger after that if command succeed or failed

harish-24: yes, would be better to let the user know if the command failed.

memory/libhugetlbfs.py
harish-24: i meant the typo :)

memory/libhugetlbfs.py
PraveenPenguin: ah sure my mistake :(

narasimhan-v: @PraveenPenguin typo needs to be changed in commit message as well.

PraveenPenguin:
@narasimhan-v @harish-24 pushed the changes please have a look

memory/libhugetlbfs.py
```diff
@@ -81,7 +78,7 @@ def setUp(self):
         if os.path.exists('/proc/sys/vm/nr_hugepages'):
             Hugepages_support = process.system_output('cat /proc/meminfo',
                                                       verbose=False,
-                                                      shell=True)
+                                                      shell=True, sudo=True)
```
narasimhan-v: The description for sudo says:
```
    :param sudo: Whether the command requires admin privileges to run,
                 so that sudo will be prepended to the command.
                 The assumption here is that the user running the command
                 has a sudo configuration such that a password won't be
                 prompted. If that's not the case, the command will
                 straight out fail.
```

So, should we capture any errors / exceptions here ?

narasimhan-v: Also, this can be a `genio.read_file()` instead.

PraveenPenguin: sure `genio.read_file()` good to use

PraveenPenguin:
@narasimhan-v adopted the change please have a look

PR #947, opened 2018-06-28
Fix: Integrity test cancel on single node systems

harish-24:
Patch cancels tests on machines with single nodes or nodes which has less memory.


narasimhan-v:
@harish-24 can you include the tar diff also into the commit message ?

harish-24:
@narasimhan-v Done.

PR #950, opened 2018-07-03
Handling firewall command for sles 15

narasimhan-v:
SLES 15 has a different firewall disable command than SLES 12.
So, handling the differences with this commit.


Naresh-ibm:
looks good for me. Thanks @narasimhan-v  for this fix.

PR #958, opened 2018-07-06
Fix: ebizzy test to get file from datadir

harish-24:
Patch fixes getting file from datadir, which already had been obtained using get_data method.


nasastry:
@harish-24 Please post the job log.

PraveenPenguin:
@harish-24 thanks for the patch

PR #973, opened 2018-08-09
Hwinfo test 

PraveenPenguin:
Fixed hwinfo test
1- fixed pep8 issue
2- change self.error to self.cancel as if hwinfo package is not installed test should not error out

### DIFF
--- a/ras/hwinfo.py
+++ b/ras/hwinfo.py
@@ -26,6 +26,9 @@ class Hwinfo(Test):
 
     is_fail = 0
 
+    def clear_dmesg(self):
+        process.run("dmesg -C ", sudo=True)
+
     def run_cmd(self, cmd):
         self.log.info("executing ============== %s =================" % cmd)
         if process.system(cmd, ignore_status=True, sudo=True):
@@ -44,6 +47,8 @@ class Hwinfo(Test):
             self.cancel("Fail to install hwinfo required for this test.")
 
     def test(self):
+
+        self.clear_dmesg()
         self.log.info(
             "===============Executing hwinfo tool test===============")
         list = self.params.get('list', default=['--all', '--cpu', '--disk'])
@@ -72,8 +77,7 @@ class Hwinfo(Test):
         self.run_cmd("hwinfo --debug 0 --disk --log=-")
         self.run_cmd("hwinfo --short --block")
         self.run_cmd("hwinfo --disk --save-config=all")
-        if "failed" in process.system_output("hwinfo --disk --save-config"
-                                             "=all | ""grep failed | tail -1",
+        if "failed" in process.system_output("hwinfo --disk --save-config=all",
                                              shell=True):
             self.is_fail += 1
             self.log.info("--save-config option failed")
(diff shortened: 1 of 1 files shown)


ras/hwinfo.py
```diff
@@ -71,8 +72,9 @@ def test(self):
         self.run_cmd("hwinfo --debug 0 --disk --log=-")
         self.run_cmd("hwinfo --short --block")
         self.run_cmd("hwinfo --disk --save-config=all")
-        if "failed" in process.system_output("hwinfo --disk --save-config=all | "
-                                             "grep failed | tail -1", shell=True):
+        if "failed" in process.system_output("hwinfo --disk --save-config"
+                                             "=all | ""grep failed | tail -1",
+                                             shell=True):
```
narasimhan-v: @PraveenPenguin can you change it into native python parsing please ?

PraveenPenguin: sure make sense will do

ras/hwinfo.py
```diff
@@ -71,8 +72,9 @@ def test(self):
         self.run_cmd("hwinfo --debug 0 --disk --log=-")
         self.run_cmd("hwinfo --short --block")
         self.run_cmd("hwinfo --disk --save-config=all")
-        if "failed" in process.system_output("hwinfo --disk --save-config=all | "
-                                             "grep failed | tail -1", shell=True):
+        if "failed" in process.system_output("hwinfo --disk --save-config"
+                                             "=all | ""grep failed | tail -1",
```
kamalesh-babulal: can the above line written as `"=all | grep failed | tail -1",` instead ?

PraveenPenguin: sure , any way i am planning to use native python to parse output

PraveenPenguin:
@narasimhan-v @kamalesh-babulal done with changes can you please review

ras/hwinfo.py
kamalesh-babulal: previously the testcase checks for last message logged into dmesg for hwinfo errors, but with the current change there might be false positives because the check for `failed` is for the whole dmesg those might not be from `hwinfo`. You probably want to clear the `dmesg` before running the test.

PraveenPenguin:
@narasimhan-v @kamalesh-babulal adopted dmesg clear before running test

PraveenPenguin:
@narasimhan-v can you please help to  review this as it think dmesg clear is different functionality  so can go as separate  patch , just want your view on this

ras/hwinfo.py
```diff
@@ -26,6 +26,9 @@ class Hwinfo(Test):
 
     is_fail = 0
 
+    def clear_dmesg(self):
+        process.run("dmesg -C ", sudo=True)
```
narasimhan-v: @PraveenPenguin If there is no option in hwinfo to clear previous erros, then this is an acceptable way.

narasimhan-v:
I will wait for @kamalesh-babulal  to approve

kamalesh-babulal:
Sorry to pick on, how about following commit message ```Adopt python native parsing of the command output. Also introduce dmesg_clear() to empty the dmesg before executing the test, parsing cleared kernel dmesg will eliminate the false negative from previous run test/other programs```

PraveenPenguin:
@kamalesh-babulal thanks for help , i adopt commit message as well

narasimhan-v:
@PraveenPenguin @kamalesh-babulal Thanks

PR #974, opened 2018-08-09
Fix compilation fail on SuSE distros

harish-24:
Compilation fails on SuSE distros due to lack of few packages. This patch fixes it.

### DIFF
--- a/kernel/kernbench.py
+++ b/kernel/kernbench.py
@@ -94,7 +94,7 @@ class Kernbench(Test):
         elif 'SuSE' in detected_distro.name:
             deps.extend(['libpopt0', 'glibc', 'glibc-devel',
                          'popt-devel', 'libcap2', 'libcap-devel',
-                         'libcap-ng-devel'])
+                         'libcap-ng-devel', 'flex', 'bison', 'openssl-devel'])
         elif detected_distro.name in ['centos', 'fedora', 'rhel']:
             deps.extend(['popt', 'glibc', 'glibc-devel', 'libcap-ng',
                          'libcap', 'libcap-devel', 'elfutils-libelf',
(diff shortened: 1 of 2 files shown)


nasastry:
@harish-24 Can you please provide the avocado run output? some thing like the following. Thanks!!
1/1 kernbench.py PASS

kernel/kselftest.py
```diff
@@ -54,7 +54,7 @@ def setUp(self):
                          'libpopt-dev', 'libcap-ng0', 'libcap-ng-dev',
                          'libnuma-dev', 'libfuse-dev', 'elfutils', 'libelf1'])
         elif 'SuSE' in detected_distro.name:
-            deps.extend(['popt', 'glibc', 'glibc-devel', 'popt-devel',
+            deps.extend(['popt', 'glibc', 'glibc-devel', 'popt-devel', 'sudo',
```
PraveenPenguin: why 'sudo' rpm is required i don't think we are running this test on suse on non root user can you please help me here

kernel/kernbench.py
```diff
@@ -94,7 +94,7 @@ def setUp(self):
         elif 'SuSE' in detected_distro.name:
             deps.extend(['libpopt0', 'glibc', 'glibc-devel',
                          'popt-devel', 'libcap2', 'libcap-devel',
-                         'libcap-ng-devel'])
+                         'libcap-ng-devel', 'flex', 'bison', 'openssl-devel'])
```
PraveenPenguin: :+1:

kernel/kselftest.py
harish-24: Yes, that is right. There are tests which by default runs with `sudo` command and hence failing just because command is not found.

harish-24:
(1/1) tests/avocado-misc-tests/kernel/kernbench.py:Kernbench.test: PASS (905.74 s)

kernel/kselftest.py
PraveenPenguin: sure

PR #975, opened 2018-08-13
Added blktests

PraveenPenguin:
Added blktests
Blktests blktests is a test framework for the Linux kernel block layer
and storage stack


kamalesh-babulal:
It will be nice to have more information in the commit message about the test case.

PraveenPenguin:
@kamalesh-babulal updated commit message , thanks for pointing

PR #976, opened 2018-08-13
kernelbench.py: Move bison, flex to common dependencies

kamalesh-babulal:
bison, flex packages names are common across distributions. Move them
to common dependencies list.


PraveenPenguin:
@kamalesh-babulal thanks for patch , @harish-24 @nasastry thanks for ACK

PR #984, opened 2018-08-14
Execute numactl tests sequentially

harish-24:
By default build.make() uses multiple cpus to execute. numactl tests fails sometimes due to parallel execution as it causes background change to numa_hit etc. This patch restricts it by
explicitly specifying -j flag for tests.

### DIFF
--- a/cpu/numactl.py
+++ b/cpu/numactl.py
@@ -70,7 +70,8 @@ class Numactl(Test):
 
     def test(self):
 
-        if build.make(self.sourcedir, extra_args='-k test', ignore_status=True):
+        if build.make(self.sourcedir, extra_args='-k -j 1'
+                      ' test', ignore_status=True):
             if len(memory.numa_nodes_with_memory()) < 2:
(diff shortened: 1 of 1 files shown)


cpu/numactl.py
```diff
@@ -70,7 +70,8 @@ def setUp(self):
 
     def test(self):
 
-        if build.make(self.sourcedir, extra_args='-k test', ignore_status=True):
+        if build.make(self.sourcedir, extra_args='-k -j 1'
```
kamalesh-babulal: How about having the `extra_args` on the same line and `ignore_status` on the next line ?. Just a suggestion, if you do not want to change that's also fine.

harish-24: `pylint` and `pep8` is compliant with the existing way. I would like to keep it this way. Thanks

kamalesh-babulal: It not about the compliance, I was mentioning about the readability. I am ok with it.

PR #993, opened 2018-08-20
Fix: transparent_hugepages test check for THP usage

harish-24:
Sometimes kernel may not be able to use THP depending on the system configuration and the requesting memory block size. At that point existing THP may never change, thus failing the test case.
This patch provides a warning to the user that THP has not been used.

### DIFF
--- a/memory/transparent_hugepages.py
+++ b/memory/transparent_hugepages.py
@@ -101,11 +101,14 @@ class Thp(Test):
                                        ("thp_collapse_alloc"))
 
         # Check whether THP is Used or not
-        if thp_alloted_after <= thp_alloted_before:
+        if thp_alloted_after < thp_alloted_before:
             e_msg = "Thp usage count has not increased\n"
             e_msg += "Before Stress:%d\nAfter stress:%d" % (thp_alloted_before,
                                                             thp_alloted_after)
             self.fail(e_msg)
+        elif thp_alloted_after == thp_alloted_before:
+            self.log.warn("Kernel did not allocate THP. Make sure "
+                          "you run on an apt configuration")
         else:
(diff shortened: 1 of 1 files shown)


harish-24:
@kamalesh-babulal Right. I have modified slightly to bail out in such cases. Thanks for the suggestion.

harish-24:
@kamalesh-babulal Can you have a look?

memory/transparent_hugepages.py
```diff
@@ -101,11 +101,14 @@ def test(self):
                                        ("thp_collapse_alloc"))
 
         # Check whether THP is Used or not
-        if thp_alloted_after <= thp_alloted_before:
+        if thp_alloted_after < thp_alloted_before:
             e_msg = "Thp usage count has not increased\n"
             e_msg += "Before Stress:%d\nAfter stress:%d" % (thp_alloted_before,
                                                             thp_alloted_after)
             self.fail(e_msg)
```
kamalesh-babulal: Sorry for the late review. How is that `thp_alloted_after < thp_alloted_before` is an error condition whereas `thp_alloted_after == thp_alloted_before` is not ? As per the test condition, if we do not increase allocation count there might be one of the two reasons:

1.  System misconfiguration and we need to bailout
2.  Issue in the THP code

How are you checking for the second case, where you know that the system can allocate memory but failed to do so ?

PraveenPenguin:
@harish-24 @kamalesh-babulal what is the next step here

harish-24:
@PraveenPenguin @kamalesh-babulal As I have concluded with the discussions offline that there is no way to verify whether THP would be provided by the kernel, the existing code holds good. Closing this as it is better to fail even if the kernel cannot allocate THP.

PR #994, opened 2018-08-20
fixed lshw test

bskjois:
Skip this test for SLES12 and below versions.

Signed-off-by: Kowshik Jois B S <kowshik.linux@gmail.com>

### DIFF
--- a/ras/lshw.py
+++ b/ras/lshw.py
@@ -53,7 +53,7 @@ class Lshwrun(Test):
 
         dist = distro.detect()
         if dist.name == "SuSE" and dist.version < 15:
-            self.cancel("lshw not supported on SUES-%s. Please run on SLES15 or higher versions only " % dist.version)
+            self.cancel("lshw not supported on SUES-%s. Please run on SLES15 or any higher version only " % dist.version)
 
         for package in ("lshw", "net-tools", "iproute", "pciutils"):
             if not sm.check_installed(package) and not sm.install(package):
--- a/ras/lshw.py
+++ b/ras/lshw.py
@@ -18,6 +18,7 @@ from avocado import main
 from avocado.utils import process
 from avocado.utils import genio
 from avocado.utils.software_manager import SoftwareManager
+from avocado.utils import distro
 
 
 class Lshwrun(Test):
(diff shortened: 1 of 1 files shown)


narasimhan-v:
@kamalesh-babulal Since the condition is checking for sles 12 and below, doesn't the current message hold good ?

kamalesh-babulal:
@narasimhan-v its just a minor nit, there was a disparity between the commit message and error message printed, it should ideally match. Commit message speak about `SLES12` whereas error message prints `SLES15`.

narasimhan-v:
@kamalesh-babulal Ah ok
* commit message - sles 12 and below
* condition check - sles 12 and below
* error message - sles 15 and above.
@bskjois makes sense to change the error message.

ras/lshw.py
```diff
@@ -49,6 +50,10 @@ def run_cmd_out(self, cmd):
 
     def setUp(self):
         sm = SoftwareManager()
+        dist = distro.detect()
+        if dist.name == "SuSE" and dist.version <= 12:
+            self.cancel("lshw is supported on SLES15 or later versions only.")
```
PraveenPenguin: can be like 
 ```       if dist.name == "SuSE" and dist.version <= 12:
               self.cancel("lshw  not  supported on SUES-%s " % dist.version ) ```

PraveenPenguin: @narasimhan-v @bskjois @kamalesh-babulal above changes will hold good

kamalesh-babulal: @PraveenPenguin with `self.cancel("lshw  not  supported on SUES-%s " % dist.version )` the user will not informed about the supported version.

kamalesh-babulal: I vote for the `lshw is supported on SLES15 or later versions only`.

PraveenPenguin: @bskjois still on error message i not convinced as I suggested  it should print distro version where it is not able to run (running distro) and skip the test  , as test not  meant for notify user where it is supported

bskjois: @PraveenPenguin I thought even that message will let user know that the test is not supported on any lower versions of SLES15. If not we shall have that message explicitly as "This test is not supported on any version lower to SLES15" OR "lshw not supported on SUES-%s. Please run on SLES15 or higher versions " % dist.version"

This will let the user know that the test is not supported on the current version and also where it is supported as well. is this looks fine?

PraveenPenguin:
@bskjois please modfiy commit message as commit message is too short 

```
 Commit message is too short (less than 5 lines).
     Here a minimal template:
     ------------------------------------
     header          <- Limited to 50 characters. No period.
                     <- Blank line
     message         <- Any number of lines, limited to 72 characters per line.
                     <- Blank line
     Signed-off-by:  <- Signature (created by git commit -s)
     ------------------------------------
.```

it make travis happy :)

bskjois:
Closing as a new PR has been raised. (https://github.com/avocado-framework-tests/avocado-misc-tests/pull/1024)

Issue #752, opened 2017-12-08
Configuration of lshw.py tests


fkromer:
The test `lshw.py:test_lshw_options` is specific to `scsi`. For systems without `scsi` it would be helpful to beeing able to disable this test.

abdhaleegit:
How do we check if system is without scsi ?
could you please run 'lshw -class scsi -class disk -class storage -short' and paste the out put here ?

skip the test if no scsi devices found in above ouput.

in powerpc:
# lshw -class scsi -class disk -class storage -short
H/W path    Device                                                           Class      Description
===================================================================================================
/3/0.1.0    /dev/sda                                                         disk       16GB VDASD
/3/0.2.0    /dev/cdrom                                                       disk       VOPTA

fkromer:
@abdhaleegit The output on a system without `scsi` is as follows:
```
$ sudo lshw -class scsi -class disk -class storage -short
H/W path  Device  Class       Description
=========================================
```

PR #995, opened 2018-08-21
ras/lshw: Fix the -disable option test

kamalesh-babulal:
Using 'scsi' -class for the -disable/-enable check, may not be right
-class, reason being there might be an empty 'scsi' -class but the
keyword 'scsi' may show in the -class 'storage' hence replace it with
more reliable -class 'network' for -enable/-disable test.

This patch also removes the dependencies on 'grep' command return value
for testcase passing criteria.

Fixes: #752


kamalesh-babulal:
v2 Changes:
- Used grep command to trim the output, this minimize the clutter on the `job.log`

PR #999, opened 2018-08-27
ras/sosreport: Mute unparsed output from commands

kamalesh-babulal:
Mute output from commands, those outputs are not parsed, like '-h help'
switch(s). We are only worried about the exit status of such commands
and as the side effect, the job.log is uncluttered with the
non-interesting information.


kamalesh-babulal:
V2 Changes:
- Use `verbose` switch instead of `allow_output_check`.

PR #1000, opened 2018-08-30
ras/perf: Add perf tool testcase

kamalesh-babulal:
This testcase verifies the following perf subcommands:
- help
- version
- list
- record
- report
- kallsyms
- annotate
- evlist
- stat
- bench

Additionally executes the builtin test benchmark

### DIFF
--- a/ras/perf.py
+++ b/ras/perf.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2018 IBM.
+# Author: Kamalesh Babulal <kamalesh@linux.vnet.ibm.com>
+
+import os
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+from avocado.utils import distro
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Perf(Test):
+
+    """
+    Performance analysis tools for Linux
+    :avocado: tags=privileged
+    """
+
+    fail_cmd = list()
+
+    def run_cmd(self, cmd, verbose=True):
+        self.log.info("executing ============== %s =================", cmd)
+        if process.system(cmd, verbose=verbose, ignore_status=True, sudo=True, shell=True):
+            self.is_fail += 1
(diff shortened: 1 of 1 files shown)


kamalesh-babulal:
@nasastry Thanks for the review.

kamalesh-babulal:
v2 Changes:
- Added the `SuSE` to the `RHEL` distro
- Explicitly add outfile to `perf record`
- Added `perf script` command
- Improved the subcommand execution logic

ras/perf.py
```diff
@@ -0,0 +1,129 @@
...
+
+import os
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+from avocado.utils import distro
+from avocado.utils.software_manager import SoftwareManager
+
+
+class perf(Test):
```
harish-24: Class name `Perf` would be appropriate here.

ras/perf.py
```diff
@@ -0,0 +1,129 @@
...
+
+    """
+    Performance analysis tools for Linux
+    :avocado: tags=privileged
+    """
+
+    fail_cmd = list()
+
+    def run_cmd(self, cmd, verbose=True):
+        self.log.info("executing ============== %s =================" % cmd)
```
harish-24: Minor suggestion. `self.log.info` is usually used with `,`.
```        self.log.info("executing ============== %s =================", cmd)```
Just a nit pick, can be left as is.

File: ras/perf.py [hunk 3]
```diff
@@ -0,0 +1,129 @@
...
+        if process.system(cmd, verbose=verbose, ignore_status=True, sudo=True, shell=True):
+            self.is_fail += 1
+            self.fail_cmd.append(cmd)
+
+    def run_cmd_out(self, cmd):
```
harish-24: As it is a generic function(not specific to class), better to add `@staticmethod` before the method. Refer https://github.com/avocado-framework-tests/avocado-misc-tests/blob/master/cpu/cpustress.py#L80 for the same.

ras/perf.py
```diff
@@ -0,0 +1,129 @@
...
+        elif dist.name in ['centos', 'fedora', 'rhel', 'SuSE']:
+            pkgs = ['perf']
+        else:
+            self.cancel("perf is not supported on %s" % dist.name)
+
+        for pkg in pkgs:
+            if not smg.check_installed(pkg) and not smg.install(pkg):
+                self.cancel("Package %s is missing/could not be installed" % pkg)
+
+    def testPerfShort(self):
```
harish-24: minor suggestion: better to have `test_perf_short`. Camel case is usually used for classes. Same for other `test*` methods.

kamalesh-babulal:
@harish-24 Thanks for the review. 

**V3 Changes:**
- Added `@staticmethod` for `run_cmd_out()`
- Changed the function naming convention to c function style.
- `self.log.info` replaced `%` with `,`

PraveenPenguin:
@nasastry @harish-24 thanks for review

PR #1003, opened 2018-08-31
[RFC] Convert non class member as static methods

kamalesh-babulal:
This series converts the non-class specific functions into staticmethod, mostly converts `run_cmd_out()` into staticmethod, expect for ` io/net/vlan_test.py` where another function `ping_check_host()` can also be converted into staticmethod. This series is inspired by review comments from pull request #1000

### DIFF
--- a/io/disk/mvcli_test.py
+++ b/io/disk/mvcli_test.py
@@ -76,7 +76,8 @@ class MvcliTest(Test):
         except CmdError as details:
             self.fail("Command %s failed %s" % (cmd, details))
 
-    def run_cmd_output(self, cmd):
+    @staticmethod
+    def run_cmd_output(cmd):
         """
         Execute the command and return output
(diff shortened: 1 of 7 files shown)


narasimhan-v:
@kamalesh-babulal does this need 7 different commits ?

PraveenPenguin:
@narasimhan-v i think we need to wrap this in single commit

kamalesh-babulal:
@narasimhan-v @PraveenPenguin thought behind was a single commit per testcase, if you suggest me to fold it into a single commit, I will re-do the pull request.

narasimhan-v:
@kamalesh-babulal the files are different, yes. But same impact (ie, none), so single commit would do here.

narasimhan-v:
@kamalesh-babulal thanks

PR #1004, opened 2018-09-01
Shutdown right HTX exerciser

bssrikanth:
Currently excerciser getting shut down is hardcoded, fixing it.

Signed-off-by: Srikanth Aithal <sraithal@linux.vnet.ibm.com>


PraveenPenguin:
@kamalesh-babulal thanks for review

PR #1006, opened 2018-09-05
ras/kprobe: Add kprobe testcase

kamalesh-babulal:
This patch adds test the kprobe functionality of the kernel. The
kprobe functionality is tested by:
1. Creating a kernel module from the
   <kernel-src>/samples/kprobe/kprobe_example.c
2. Load the module a.k.a planting the kprobe in do_fork syscall
3. Execute 'date' command to invoke do_fork syscall
4. Unload the module a.k.a unregister the kprobe

### DIFF
--- a/ras/kprobe.py
+++ b/ras/kprobe.py
@@ -1,0 +1,137 @@
+#!/usr/bin/env python
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2018 IBM.
+# Author: Kamalesh Babulal <kamalesh@linux.vnet.ibm.com>
+
+import os
+import shutil
+import tempfile
+from avocado import Test
+from avocado import main
+from avocado.utils import build
+from avocado.utils import distro
+from avocado.utils import process
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Kprobe(Test):
+
+    """
+    Test kernel kprobe
+    :avocado: tags=privileged
+    """
+
+    fail_cmd = list()
+
+    def run_cmd(self, cmd):
+        self.log.info("executing ============== %s =================", cmd)
+        if process.system(cmd, ignore_status=True, sudo=True, shell=True):
+            self.is_fail += 1
+            self.fail_cmd.append(cmd)
+        return
+
+    @staticmethod
+    def run_cmd_out(cmd):
+        return process.system_output(cmd, shell=True, ignore_status=True,
+                                     sudo=True)
+
+    def clear_dmesg(self):
+        process.run("dmesg -C ", sudo=True)
+
+    def setUp(self):
+        """
+        Setting up the env for the kernel building
+        """
+        smg = SoftwareManager()
+        detected_distro = distro.detect()
+        deps = ['gcc', 'make', 'automake', 'autoconf', 'time', 'bison', 'flex']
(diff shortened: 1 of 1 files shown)


ras/kprobe.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def setUp(self):
+        """
+        Setting up the env for the kernel building
+        """
+        smg = SoftwareManager()
+        detected_distro = distro.detect()
+        deps = ['gcc', 'make', 'automake', 'autoconf', 'time', 'bison', 'flex']
+        if 'Ubuntu' in detected_distro.name:
+            linux_headers = 'linux-headers' + os.uname()[2]
```
PraveenPenguin: please use %s to concatenate string 
as ```linux_headers = 'linux-headers%s' %s os.uname()[2]```

ras/kprobe.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        Building of the kprobe kernel module
+        """
+        self.log.info("============== Building kprobe Module =================")
+        self.sourcedir = tempfile.mkdtemp()
+        os.chdir(self.sourcedir)
+
+        self.location = ('https://raw.githubusercontent.com/torvalds/linux/'
+                         'master/samples/kprobes/kprobe_example.c')
+        kprobe_file = self.fetch_asset(self.location, expire='1d')
```
PraveenPenguin: is this files changes every day ?

ras/kprobe.py
```diff
@@ -0,0 +1,135 @@
...
+        shutil.copy(kprobe_file, self.sourcedir + '/kprobe_example.c')
```
PraveenPenguin: please use ```os.path.join```

ras/kprobe.py
```diff
@@ -0,0 +1,135 @@
...
+        Write module make file on the fly
+        """
+        makefile = open("Makefile", "w")
+        makefile.write('obj-m := kprobe_example.o\nKDIR := /lib/modules/$(shell uname -r)/build'
+                       '\nPWD := $(shell pwd)\ndefault:\n\t'
+                       '$(MAKE) -C $(KDIR) SUBDIRS=$(PWD) modules\n')
+        makefile.close()
+
+        self.is_fail = 0
+        self.run_cmd("make")
```
PraveenPenguin: ```build.make(self.sourcedir)``` can be good option

ras/kprobe.py
```diff
@@ -0,0 +1,135 @@
...
+        if self.is_fail >= 1:
+            self.fail("insmod kprobe_example.ko failed")
+
+        if "Planted kprobe" not in self.run_cmd_out("dmesg |grep -i planted"):
+            self.fail("kprobe couldn't be planted, check dmesg for more information")
+
+        """
+        Execute date to trigger do_fork syscall
+        """
+        self.run_cmd("date")
```
PraveenPenguin: ok

ras/kprobe.py
kamalesh-babulal: `os.uname().[2]` returns string, so it should be safe to concatenate. Sure will change it

ras/kprobe.py
kamalesh-babulal: I choosed that value based on the `kernbench.py` but yeah on other hand it might not change everyday, So do you think `7days` as an ideal number ?

ras/kprobe.py
kamalesh-babulal: Sure will change that.

ras/kprobe.py
kamalesh-babulal: Sure, will replace it

ras/kprobe.py
PraveenPenguin: generally we use 7 days for all the place can be good option if any other thought is welcome , as it just make sure to avoid download if there is no change in file

kamalesh-babulal:
**V2 Changes:**
- Used `%s` for string concatenation using string format `%`
- Used `build.make` instead of `rum_cmd` to build module
- Change the expiry date from 1day to 7days for expire of the kprobe test
- Used `os.join.path` instead of simple string concatenation.

PR #1011, opened 2018-09-07
 ras/kretprobe: Add kretprobe testcase

kamalesh-babulal:
This patch adds test the kretprobe functionality of the kernel. The
kretprobe functionality is tested by:
1. Creating a kernel module from the
   <kernel-src>/samples/kprobe/kretprobe_example.c
2. Load the module a.k.a planting the kretprobe in do_fork syscall
3. Execute 'date' command to invoke do_fork syscall
4. Unload the module a.k.a unregister the kretprobe

### DIFF
--- a/ras/kretprobe.py
+++ b/ras/kretprobe.py
@@ -1,0 +1,137 @@
+#!/usr/bin/env python
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2018 IBM.
+# Author: Kamalesh Babulal <kamalesh@linux.vnet.ibm.com>
+
+import os
+import shutil
+import tempfile
+from avocado import Test
+from avocado import main
+from avocado.utils import build
+from avocado.utils import distro
+from avocado.utils import process
+from avocado.utils.software_manager import SoftwareManager
+
+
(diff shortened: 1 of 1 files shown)


ras/kretprobe.py
```diff
@@ -0,0 +1,137 @@
...
+from avocado import Test
+from avocado import main
+from avocado.utils import build
+from avocado.utils import distro
+from avocado.utils import process
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Kretprobe(Test):
+
```
sathnaga: @kamalesh-babulal can be combined with Kprobe test and handle different type of probes using yaml params? or subtests here
given that lot of steps are common?

kamalesh-babulal: Thanks for the review and I was expecting this question. My personal preference is a short and individual testcase, instead of handling too many things in the same testcase. The major reason behind having them as individual testcase, is to later add different testcase scenarios supported by each of them, this testcases serves as the template. I might be missing some information here, please correct me if I am wrong:
- Both kprobe and kretprobe share:
    * common setup code
    * common build code (not really, we need to check if kprobe/kretprobe download file based on it)
2. The search would be completely different for patterns from dmesg

The way I see it while unification is, read from yaml file on if we are probing kprobe or kretprobe:
- Use common setup code (installing package)
- Alter the file based on probe
- Use common build code
- Use different string matching based on probe (two different functions)
- Add a yam file

The unification plan doesn't considers the options like kprobe sub tests/kretprobe sub tests, which might added later.

PraveenPenguin: @kamalesh-babulal thanks for explanation as this is just a suggestion if you want this should be fine as you are owner of test case :)

kamalesh-babulal: @PraveenPenguin I will glad,if you can pick it up or else will add this to kprobe.

PraveenPenguin: @kamalesh-babulal  i think good to go will review and if all looks ok will merge

PR #1012, opened 2018-09-12
perf/perf_fuzzer: Add Linux perf_events syscall fuzzer

kamalesh-babulal:
This testcase add a Linux perf_events syscall fuzzer. More information
on the fuzzer can be found at
https://github.com/deater/perf_event_tests/blob/master/fuzzer/README

This test might crash the system.


perf/perf_fuzzer.py
```diff
@@ -0,0 +1,97 @@
...
+            kernel_ver = platform.uname()[2]
+            deps.extend(['linux-tools-common', 'linux-tools-%s'
+                         % kernel_ver])
+        # FIXME: "redhat" as the distro name for RHEL is deprecated
+        # on Avocado versions >= 50.0.  This is a temporary compatibility
+        # enabler for older runners, but should be removed soon
+        elif detected_distro.name in ['rhel', 'SuSE', 'fedora', 'redhat']:
+            deps.extend(['perf'])
+        else:
+            self.cancel("Install the package for perf supported by %s"
```
PraveenPenguin: can this message  will be having more sense 
            `self.cancel("Test not supported on  %s"% detected_distro.name)`

perf/perf_fuzzer.py
```diff
@@ -0,0 +1,97 @@
...
+        Building the perf event test suite
+        """
+        tarball = self.fetch_asset('perf-event.zip', locations=[
+                                   'https://github.com/deater/'
+                                   'perf_event_tests/archive/'
+                                   'master.zip'], expire='7d')
+        archive.extract(tarball, self.workdir)
+        self.sourcedir = os.path.join(self.workdir, 'perf_event_tests-master')
+        if build.make(self.sourcedir, extra_args="-s -S") > 0:
+            self.fail("Building perf event test suite failed")
```
PraveenPenguin: can this whole section moved to setUp sub routine   , as this section looks like test setup or at least from line 69 - line 74

perf/perf_fuzzer.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def execute_perf_fuzzer(self):
+        os.chdir(self.sourcedir)
+        genio.write_one_line("/proc/sys/kernel/perf_event_paranoid", "-1")
+        if "-1" not in genio.read_one_line("/proc/sys/kernel/perf_event_paranoid"):
+            self.error("Unable to set perf_event_paranoid to -1 ")
```
PraveenPenguin: i think `self.fail ` will be more suitable option here as it also part of test

perf/perf_fuzzer.py
```diff
@@ -0,0 +1,97 @@
...
+        self.perf_fuzzer = os.path.join(self.sourcedir, "fuzzer/perf_fuzzer")
+        if not os.path.exists(self.perf_fuzzer):
+            self.error("fuzzer not found at %s" % self.perf_fuzzer)
```
PraveenPenguin: here i think self.cancel will be used as it is test pre requisite

perf/perf_fuzzer.py
kamalesh-babulal: Here is the `perf` package is missing, the right error is to complain about the missing package instead of failing with test not supported.

perf/perf_fuzzer.py
PraveenPenguin: ah so it is system issue so it think it comes as `pre requisite` so better option is self.cancel

perf/perf_fuzzer.py
kamalesh-babulal: It would be clean to have system pre-requisite in setup and move all test related things to build function. Because as the function does nothing beyond setting up the testcase. It should be something like extended setup function. If you insist, we should move the testcase downloading to setup, I can re-spin the patch.

perf/perf_fuzzer.py
PraveenPenguin: sure make sense so i think message can be rephrase as 
```Perf package installation not supported by %s or can be issue with software repository please check  "% detected_distro.name ```

perf/perf_fuzzer.py
PraveenPenguin: @kamalesh-babulal no  hard feeling here , just i want to make uniform all the test cases in repository and it also aligned with purpose of setup

perf/perf_fuzzer.py
kamalesh-babulal: I beg to differ, here we are writing into the `sys` interface and the kernel returns error. It should be considered error, rather than cancel. It really confusing on where to use `self.{fail,cancel, error}` do we have guidelines on it ?

perf/perf_fuzzer.py
kamalesh-babulal: Sure, I will address it

perf/perf_fuzzer.py
PraveenPenguin: @kamalesh-babulal hey no hard feeling just a suggestion , i am suggesting here self.cancel as it seems me system pre requisite ,  you are owner of test if you feel if it is like self.error should be fine   as it just kind of suggestion as it just kind of representation of issues ,

PR #1013, opened 2018-09-14
ras/kprobes: optprobe test

kamalesh-babulal:
By default the optprobes is enabled for the kernel functions and this
test disables the optprobes using the '/proc/' knob available to
disable optrobes, while planting a kprobe. By disabling optprobe, the
kprobe mechanism fallbacks to un-optimizes probes. It reuses the kprobes
kernel module already build to test un-optimized version of kprobe.


kamalesh-babulal:
Thanks for the review. I wonder how the `inspekt checkall --disable-lint W,R,C,E1002,E1101,E1103,E1120,F0401,I0011,E0203 --no-license-check` missed it. Is there anything I can add to inspekt to report such issues ? or `inspekt` doesn't catch such issues ?

**v2 changes:**
1.  Converted `self.log.fail` -> `self.fail`, suggested by Praveen.
2.  Used `self` prefix for the local variables.

ras/kprobe.py
```diff
@@ -128,9 +135,33 @@ def execute_test(self):
         if "kprobe" not in self.run_cmd_out("dmesg |grep -i unregistered"):
             self.fail("kprobe unregistering failed, check dmesg for more information")
 
+    def optprobes_disable_test(self):
+        self.optprobes_file = "/proc/sys/debug/kprobes-optimization"
```
PraveenPenguin: i don't think so this variable required class scope , otherwise looks good

kamalesh-babulal:
**v3 changes:**
- Change the class scope of variables.

PR #1017, opened 2018-09-18
Fixed ras:lshw test

PraveenPenguin:
1- changed package missing logging  error to cancel as
if prerequisite not match test should cancel out
2-Change package name of iproute to iproute2 onwards ubuntu 18

log:

2018-09-18 02:18:49,746 process          L0392 DEBUG| [stdout] Reading package lists...
2018-09-18 02:18:49,883 process          L0392 DEBUG| [stdout] Building dependency tree...
2018-09-18 02:18:49,883 process          L0392 DEBUG| [stdout] Reading state information...
2018-09-18 02:18:49,898 process          L0392 DEBUG| [stdout] Package iproute is not available, but is referred to by another package.
2018-09-18 02:18:49,898 process          L0392 DEBUG| [stdout] This may mean that the package is missing, has been obsoleted, or
2018-09-18 02:18:49,898 process          L0392 DEBUG| [stdout] is only available from another source
2018-09-18 02:18:49,898 process          L0392 DEBUG| [stdout] However the following packages replace it:
2018-09-18 02:18:49,898 process          L0392 DEBUG| [stdout]   iproute2
2018-09-18 02:18:49,898 process          L0392 DEBUG| [stdout]
2018-09-18 02:18:49,952 process          L0392 DEBUG| [stderr] E: Package 'iproute' has no installation candidate
2018-09-18 02:18:49,952 process          L0648 INFO | Command '/usr/bin/apt-get --yes --allow-unauthenticated -o Dpkg::Options::="--force-confdef" -o D


kamalesh-babulal:
looks good, you can fold the first commit into the second because it just a minor change.

PraveenPenguin:
@kamalesh-babulal no i don't think so it need to be fold in single commit

kamalesh-babulal:
Second commit is adding the check for the `iproute` and first commit bails with cancel instead of error when package dependency is not meet. With error or cancel the testcases exit with different error code but the functionality remains same.

kamalesh-babulal:
@PraveenPenguin It appears to be same after second read also ;). Let me try to explain little more on my intention, I might be every well wrong on the coding style. AFAIK, the patch is altering the same function and changing the error type is trivial and could be every well added into a single code. The choice is up to you.

PraveenPenguin:
@kamalesh-babulal kind of confused here :) , let's wait for others input

narasimhan-v:
The code path affected by both commits is the same (ie, the package install). And the second commit is not really adding any functionality, so does not ask for a separate commit.

PraveenPenguin:
@narasimhan-v so can wrap in single commit ?

narasimhan-v:
@PraveenPenguin yes, can be a single commit :-)

PR #1018, opened 2018-09-18
Convert error to cancel for package installation failure

kamalesh-babulal:
To maintain consistency across all the testcases, convert the error to
cancel if the package dependency is not meet.

### DIFF
--- a/generic/connectathon.py
+++ b/generic/connectathon.py
@@ -62,8 +62,8 @@ class Connectathon(Test):
 
         for package in packages:
             if not smm.check_installed(package) and not smm.install(package):
-                self.error("Fail to install %s required for this test." %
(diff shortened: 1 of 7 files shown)


PraveenPenguin:
@kamalesh-babulal looks good , commit message  can be rephrased as it not kind of  consistency across the repository as test should `skip` or `cancel` if test pre requisite  not satisfied

kamalesh-babulal:
How about "Testcase is expected to skip or cancel execution, if the per-requisite are not meet. This patch converts the testcases across the repository to abide to this rule, instead of failing or error'ing out."

PraveenPenguin:
@kamalesh-babulal have you got time to re phrased the commit message

PR #1019, opened 2018-09-19
perf/generic_events: Add support for POWER9 events

kamalesh-babulal:
This patch series is motivated by #911, Where the raw code events file is adopted from it with a minor addition of event code for `mem_access` on `POWER8`. `perf_genericevents` file, will use the raw code events file in the second commit, instead of using the hard coded value. Each commit message explains in details about the changes made by the commits.

### DIFF
--- a/perf/perf_genericevents.py.data/raw_code.cfg
+++ b/perf/perf_genericevents.py.data/raw_code.cfg
@@ -1,0 +1,53 @@
+[POWER8]
+cpu-cycles = 0x1e
+stalled-cycles-frontend = 0x100f8
+stalled-cycles-backend = 0x4000a
+instructions = 0x02
+branch-instructions = 0x10068
+branch-misses = 0x400f6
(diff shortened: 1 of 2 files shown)


perf/perf_genericevents.py.data/raw_code.cfg
```diff
@@ -0,0 +1,53 @@
...
+LLC-load-misses = 0x300fe
+LLC-loads = 0x4c042
+LLC-prefetches = 0x4e052
+LLC-store-misses = 0x26880
+LLC-stores = 0x16880
+branch-load-misses = 0x400f6
+branch-loads = 0x4d05e
+dTLB-load-misses = 0x300fc
+iTLB-load-misses = 0x400fc
+
```
PraveenPenguin: ah just one issue can this blank line removed , any way it ran fine and looks good , I will take care while merging

PR #1036, opened 2019-02-06
Changing the order of execution.

Naresh-ibm:
Changing the order of execution due to input dependency of next functions.


PraveenPenguin:
@Naresh-ibm thanks for patch , @narasimhan-v thanks for ACK

PR #1052, opened 2019-03-22
Fixed fs:xfstest

PraveenPenguin:
indent package support drop from RHEL8 so removing this from
required list of package

### DIFF
--- a/fs/xfstests.py
+++ b/fs/xfstests.py
@@ -77,6 +77,9 @@ class Xfstests(Test):
                              'xfsdump', 'psmisc', 'sed', 'libacl-devel',
                              'libattr-devel', 'libaio-devel', 'libuuid-devel',
                              'openssl-devel', 'xfsprogs-devel'])
+
+            if self.detected_distro.name == 'rhel' and self.detected_distro.version.startswith('8'):
+                packages.remove('indent')
             if self.detected_distro.name == 'SuSE':
                 packages.extend(['libbtrfs-devel', 'libcap-progs'])
             else:


PraveenPenguin:
depend on https://github.com/avocado-framework/avocado/pull/3078

fs/xfstests.py
```diff
@@ -73,17 +73,21 @@ def setUp(self):
         # enabler for older runners, but should be removed soon
         elif self.detected_distro.name in ['centos', 'fedora', 'rhel',
                                            'redhat', 'SuSE']:
-            packages.extend(['acl', 'bc', 'dump', 'indent', 'libtool', 'lvm2',
+            packages.extend(['acl', 'bc', 'dump',  'libtool', 'lvm2',
                              'xfsdump', 'psmisc', 'sed', 'libacl-devel',
                              'libattr-devel', 'libaio-devel', 'libuuid-devel',
                              'openssl-devel', 'xfsprogs-devel'])
+
+            if not (self.detected_distro.name == 'rhel' and '8' in self.detected_distro.version):
```
harish-24: Instead of adding the same package for each distribution, better to add it once and remove only in case of `rhel 8`.

PraveenPenguin: @harish-24 ah yes looks good idea

PraveenPenguin:
@harish-24 thanks for review Pushed the changes

fs/xfstests.py
```diff
@@ -77,6 +77,9 @@ def setUp(self):
                              'xfsdump', 'psmisc', 'sed', 'libacl-devel',
                              'libattr-devel', 'libaio-devel', 'libuuid-devel',
                              'openssl-devel', 'xfsprogs-devel'])
+
+            if self.detected_distro.name == 'rhel' and '8' in self.detected_distro.version:
```
harish-24: Missed suggesting this first, safer to use `version.startswith('8')` so that we can cover all cases.

PraveenPenguin: @harish-24 sure make sense will adopt

PraveenPenguin:
@harish-24 pushed the changes

PraveenPenguin:
@harish-24 thanks for review

PR #1089, opened 2019-04-24
Added support for SLES.

Pavithra1602:
As numatop package is now available on SLES aswell, adding support to run the test on SLES.


PraveenPenguin:
@Pavithra1602 can you please split single commit in two  commit as i see code changes i not only for suse   example running `autogen.sh`  is generic code change

Pavithra1602:
Raised seperate PR.

PR #1101, opened 2019-05-07
hugepage_sanity: handle error if hugepage attributes are not available

balamuruhans:
If hugepage attributes are not available in /proc/meminfo, handle
prior starting to test and error out gracefully.

fix compiler warnings implicit declaration for `access()` and `time()`
by including unistd.h, declare `alloc_hugepage()` definition before using
it in `check_alloc_free_huge_page()` and few style changes.

If hugepages attributes are not available in /proc/meminfo,

Before fix:
Using default hugepagesize and 1 hugepage
PAGESIZE: 137438926347 MB	Number of pages: 1
Given hugepage size is not supported in kernel

After fix:
Using default hugepagesize and 1 hugepage
Error: Pagesize is invalid or not available

Signed-off-by: Balamuruhan S <bala24@linux.vnet.ibm.com>

### DIFF
--- a/memory/hugepage_sanity.py.data/hugepage_sanity.c
+++ b/memory/hugepage_sanity.py.data/hugepage_sanity.c
@@ -14,6 +14,7 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include <unistd.h>
 #include <sys/mman.h>
 #include <errno.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


PraveenPenguin:
@balamuruhans Thanks for Patch , @harish-24 thanks for ACK , merging as we already have a ack

PR #1106, opened 2019-05-13
memory/libhugetlbfs: get supported hugepage sizes

balamuruhans:
The test would fail if run inside VM as hugepages are
hardcoded for PowerNV, query the supported hugepage sizes
to proceed for test.

Signed-off-by: Balamuruhan S <bala24@linux.vnet.ibm.com>

### DIFF
--- a/memory/libhugetlbfs.py
+++ b/memory/libhugetlbfs.py
@@ -26,8 +26,10 @@ from avocado.utils import process
 from avocado.utils import build
 from avocado.utils import kernel
 from avocado.utils import git
+from avocado.utils import distro
+from avocado.utils import genio
+from avocado.utils import memory
 from avocado.utils.software_manager import SoftwareManager
-from avocado.utils import distro, genio
 
 
 class libhugetlbfs(Test):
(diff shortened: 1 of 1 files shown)


balamuruhans:
depends on: https://github.com/avocado-framework/avocado/pull/3119

memory/libhugetlbfs.py
```diff
@@ -67,9 +69,8 @@ def setUp(self):
             self.cancel("libpthread.a is required!!!"
                         "\nTry installing glibc-static")
 
-        self.page_sizes = ['16']
-        if 'POWER9' in cpuinfo and 'PowerNV' in cpuinfo:
-            self.page_sizes = ['2', '1024']
+        page_sizes = memory.get_supported_huge_pages_size()
+        self.page_sizes = [str(each // 1024) for each in page_sizes]
```
narasimhan-v: Can't we do `// 1024` also in the utils ?

balamuruhans: @narasimhan-v I too thought about that, but the existing APIs in avocado return values in kB so I adhered to it.

narasimhan-v: oh, then it makes sense.

PraveenPenguin:
merging as utils is part of avocado lib

PR #1109, opened 2019-05-17
Changing bonding interface parameter in yaml

narasimhan-v:
Bonding test needed input interfaces as host_interfaces, which is
now changed into bond_interfaces.
This is needed, since when we run multiple tests with bonding,
and if they need to run on the bond interface created, then the
parameter name on bonding test and other tests need to be
different.


narasimhan-v:
Since this is just a yaml parameter change, I assume there is no need for any job.log

PR #1115, opened 2019-05-23
memory: pylint and code improvements

harish-24:
Patch makes sure memory tests are compliant with pylint and has some code improvements.

### DIFF
--- a/memory/dma_memtest.py
+++ b/memory/dma_memtest.py
@@ -51,6 +51,7 @@ class DmaMemtest(Test):
         self.nfail = 0
         self.tmpdir = data_dir.get_tmp_dir()
         self.tmpdir = self.params.get('dir_to_extract', default=self.tmpdir)
+        self.base_dir = os.path.join(self.tmpdir, 'linux.orig')
         tarball_base = self.params.get('tarball_base',
                                        default='linux-2.6.18.tar.bz2')
         kernel_repo = self.params.get('kernel_repo',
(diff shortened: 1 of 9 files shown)


memory/memhotplug.py
```diff
@@ -65,19 +62,19 @@ def get_hotpluggable_blocks(path, ratio):
         if memory.is_hot_pluggable(block):
             mem_blocks.append(block)
 
-    def f(num):
+    def percent(num):
```
narasimhan-v: docstring please

memory/dma_memtest.py
```diff
@@ -49,6 +49,7 @@ def setUp(self):
         copies of the linux kernel that will be uncompressed.
         """
         self.nfail = 0
+        self.base_dir = None
```
narasimhan-v: why is this initialised to None ?
It can be `self.base_dir = os.path.join(self.tmpdir, 'linux.orig')` here, right ?

harish-24: Right. Will move it after initializing `self.tmpdir`.

harish-24:
@narasimhan-v Pushed with changes, Thanks.

memory/memhotplug.py
narasimhan-v: @harish-24

harish-24: @narasimhan-v Changes pushed. Forgot to put it here.

narasimhan-v: @harish-24 this function is doing the reverse of percent.

harish-24: @narasimhan-v pushed with appropriate definition.

narasimhan-v: @harish-24 Thanks, I am merging this. Maybe you can send a patch that optimises the function later.

PR #1120, opened 2019-05-28
Fix: Install patch wherever applicable

harish-24:
Patch adds/removes 'patch' packages based on requirement of test


harish-24:
Re-based after conflict

PR #1130, opened 2019-06-03
Fix fuzz

PraveenPenguin:
1-Fix fuzzer test cases as if command execution failed it error out 
2-Fix style issue

### DIFF
--- a/fuzz/fsfuzzer.py
+++ b/fuzz/fsfuzzer.py
@@ -90,7 +90,8 @@ class Fsfuzzer(Test):
         Runs the fsfuzz test suite. By default uses all supported fstypes,
         but you can specify only one by `fstype` param.
         '''
-        process.system("%s %s" % (self._fsfuzz, self._args), sudo=True, ignore_status=True)
+        process.system("%s %s" % (self._fsfuzz, self._args), sudo=True,
+                       ignore_status=True)
 
 
 if __name__ == "__main__":
--- a/fuzz/fsfuzzer.py
+++ b/fuzz/fsfuzzer.py
@@ -68,7 +68,8 @@ class Fsfuzzer(Test):
 
         if d_name == "ubuntu":
             # Patch for ubuntu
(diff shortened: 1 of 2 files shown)


fuzz/fsfuzzer.py
```diff
@@ -90,7 +90,7 @@ def test(self):
         Runs the fsfuzz test suite. By default uses all supported fstypes,
         but you can specify only one by `fstype` param.
         '''
-        process.system("%s %s" % (self._fsfuzz, self._args), sudo=True)
+        process.system("%s %s" % (self._fsfuzz, self._args), sudo=True, ignore_status=True)
```
narasimhan-v: pep8

PraveenPenguin:
@narasimhan-v thanks for review pushed the changes

fuzz/fsfuzzer.py
```diff
@@ -90,7 +90,8 @@ def test(self):
         Runs the fsfuzz test suite. By default uses all supported fstypes,
         but you can specify only one by `fstype` param.
         '''
-        process.system("%s %s" % (self._fsfuzz, self._args), sudo=True)
+        process.system("%s %s" % (self._fsfuzz, self._args), sudo=True,
+                       ignore_status=True)
```
narasimhan-v: @PraveenPenguin what is the plan here to pass / fail the test ?

PraveenPenguin: @narasimhan-v still i need to figure out  which condition i need to have for fail and it will take some time so thinking i can have this on coming patch as checking exit status of this not really help

narasimhan-v: @PraveenPenguin I guess then we can use ignore_status=True when we come up with that. If we make the test PASS irrespective of the test effect, we will have false PASSES.

Your thoughts ?

PraveenPenguin: @narasimhan-v as this is fuzzer so i think should be ok , if you think i will check exit status

PraveenPenguin:
@narasimhan-v pushed the changes please have a look

PR #1141, opened 2019-06-07
Nsstress test fixes

PraveenPenguin:
1- Added yaml input support for test input 
2- Fix pep8 issue in test


generic/nstress.py
```diff
@@ -47,15 +47,18 @@ def setUp(self):
...
         self.duration = self.params.get('duration', default=300)
+        self.memchunk = self.params.get('memchunk', default="250")
+        self.memchunk_nmem64 = self.params.get('memchunk_nmem64', default="2047")
+        self.procs = self.params.get('procs', default="250")
 
     def test(self):
         os.chdir(self.workdir)
-        self.run_cmd("./nmem -m 250 -s %s" % self.duration)
-        self.run_cmd("./nmem64 -m 2047 -s %s" % self.duration)
+        self.run_cmd("./nmem -m %s -s %s" % (self.memchunk, self.duration))
```
narasimhan-v: @PraveenPenguin Is there a need to run these tests in sequence ?
If yes:
    the test is failing fast after nmem64
If no:
    can you split variants into mux

PraveenPenguin: @narasimhan-v sequence not needed , one help which variants you are looking as both has different cmd

narasimhan-v: @PraveenPenguin the command and param can be given via yaml

PraveenPenguin: @narasimhan-v ok may be i think should fine as test should able to run without yaml input as well

PraveenPenguin:
@narasimhan-v can this be in , other suggested change  i will send in another PR as git rebase conflict me :(

PR #1146, opened 2019-06-11
Fix: hard lockup detection after stress

harish-24:
After stress runs, on newer distro kernels the Hard lockup pattern has changed.
eg. hard LOCKUP @ _raw_spin_lock_irq+0x20/0xe0

This patch allows to support new and old kernel lockups

### DIFF
--- a/generic/stress-ng.py
+++ b/generic/stress-ng.py
@@ -154,7 +154,7 @@ class Stressng(Test):
         collect_dmesg(self)
         ERROR = []
         pattern = ['WARNING: CPU:', 'Oops', 'Segfault', 'soft lockup',
-                   'Unable to handle', 'Hard LOCKUP']
+                   'Unable to handle', 'ard LOCKUP']
         logs = process.system_output('dmesg').splitlines()
         for fail_pattern in pattern:
             for log in logs:


generic/stress-ng.py
```diff
@@ -154,7 +154,7 @@ def test(self):
         collect_dmesg(self)
         ERROR = []
         pattern = ['WARNING: CPU:', 'Oops', 'Segfault', 'soft lockup',
-                   'Unable to handle', 'Hard LOCKUP']
+                   'Unable to handle', 'ard LOCKUP']
```
narasimhan-v: @harish-24 wouldn't comparing with ignore case be better option ?

harish-24: @narasimhan-v Yes, will adapt that.

harish-24: @narasimhan-v On a second thought, all other strings are case sensitive and messages like `WARNING: CPU`, `Segfault` is better to check as is than to change case and check. So on a wider thought I think existing fix would be better.

narasimhan-v: ok

PraveenPenguin:
@harish-24 after merging this i get to know that one more test has that , will send PR for that

PR #1148, opened 2019-06-11
kselftest.py: removing unwanted package for SuSE

shirishaganta:
Removing popt package name for SuSE, because it is no longer
shipped with the latest versions

Signed-off-by: Shirisha Ganta <shiganta@in.ibm.com>


harish-24:
Commit message can be
```Removing popt package name for SuSE, because it is source package in newer releases and has not been shipped as a binary package after SLES11SP4```

PraveenPenguin:
@harish-24 thanks for ACK , merging

PR #1150, opened 2019-06-11
livepatch.py: Corrected openssl-devel package name

shirishaganta:
In SuSE openssl-devel package is shipped as libopenssl-devel.

Signed-off-by: Shirisha Ganta <shiganta@in.ibm.com>


PraveenPenguin:
@harish-24 thanks for ACK , merging

PR #1157, opened 2019-06-13
Add test for multiple namespace in a single region

harish-24:
Add test for multiple namespace in a single region using the existing regions size

### DIFF
--- a/memory/ndctl.py
+++ b/memory/ndctl.py
@@ -28,6 +28,7 @@ from avocado.utils import process
 from avocado.utils import archive
 from avocado.utils import distro
 from avocado.utils import build
+from avocado.utils import genio
 from avocado.utils.software_manager import SoftwareManager
 
 
@@ -38,21 +39,28 @@ class NdctlTest(Test):
 
     """
 
-    def get_json(self, option=''):
+    def get_json(self, short_opt='', long_opt=''):
         """
         Get the json of each provided options
 
-        return: By default returns entire detail of namespaces
+        return: By default returns entire list of json objects
+        return: Empty list is no output is empty
         """
+        if short_opt:
+            option = short_opt
+        elif long_opt:
+            option = long_opt
+        else:
+            option = ''
         try:
             json_op = json.loads(process.system_output(
                 '%s list %s' % (self.binary, option), shell=True))
         except ValueError:
             json_op = []
-        if option:
+        if short_opt:
             vals = []
             for nid in json_op:
-                vals.append(self.get_json_val(nid, self.opt_dict[option]))
+                vals.append(self.get_json_val(nid, self.opt_dict[short_opt]))
             return vals
         return json_op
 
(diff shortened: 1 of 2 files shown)


memory/ndctl.py
```diff
@@ -38,21 +38,29 @@ class NdctlTest(Test):
...
     """
 
-    def get_json(self, option=''):
+    def get_json(self, short_opt='', long_opt=''):
         """
         Get the json of each provided options
 
         return: By default returns entire detail of namespaces
         """
+        if short_opt and long_opt:
```
narasimhan-v: This check is unnecessary, since the function param is all static.
Also, [] is returned in 2 different cases: option mistake, and value error.
Also, it is better to add this behaviour to docstring

memory/ndctl.py
```diff
@@ -266,8 +274,25 @@ def test_namespace_modes(self):
...
+        region = self.get_json(short_opt='-R')[0]
+        self.log.info("Using %s for muliple namespace regions", region)
+        self.disable_namespace(region=region)
+        self.destroy_namespace(region=region)
+        size = self.get_json_val(self.get_json(
+            long_opt='-r %s' % region)[0], 'size')
+        self.log.info("Creating %s namespaces", cnt)
+        for nid in range(0, cnt):
+            self.create_namespace(region=region, mode='fsdax', size=size / cnt)
+            self.log.info("Namespace %s created", nid + 1)
```
narasimhan-v: Suggestion: Log can be moved to create_namespace()

memory/ndctl.py
harish-24: Right, anyway only the test uses it so will remove the check or document it.

memory/ndctl.py
harish-24: This is specific to particular test, just to specify the count of namespace.

harish-24:
@narasimhan-v Please review again, pushed with few other changes as well.

memory/ndctl.py
narasimhan-v: ok

memory/ndctl.py
```diff
@@ -266,8 +274,39 @@ def test_namespace_modes(self):
...
+            long_opt='-r %s' % region)[0], 'size')
+        self.log.info("Creating %s namespaces", cnt)
+        if 'Hash' in genio.read_file('/proc/cpuinfo').rstrip('\t\r\n\0'):
+            def_align = 16 * 1024 * 1024
+        else:
+            def_align = 2 * 1024 * 1024
+        if ((size / cnt) % def_align) != 0:
+            self.log.warn("Namespace would fail as it is not %sM "
+                          "aligned! Chaning the number of "
+                          "namespaces", def_align / 1024 * 1024)
```
narasimhan-v: `def_align / ( 1024 * 1024 )`

harish-24: Done, Thanks.

PR #1179, opened 2019-07-03
Vnic enhancements [v2]

abdhaleegit:
7534bf0b Get LPAR and Server from HMC
8cc6167 enable vNIC priority option for vNIC tests

### DIFF
--- a/io/net/virt-net/network_virtualization.py
+++ b/io/net/virt-net/network_virtualization.py
@@ -41,6 +41,7 @@ IS_KVM_GUEST = 'qemu' in open('/proc/cpuinfo', 'r').read()
 
 
 class CommandFailed(Exception):
+
     '''
     Defines the exception called when a
     command fails
@@ -58,6 +59,7 @@ class CommandFailed(Exception):
 
 
 class NetworkVirtualization(Test):
+
     '''
     Adding and deleting Network Virtualized devices from the vios
     Performs adding and deleting of Backing devices
@@ -113,6 +115,18 @@ class NetworkVirtualization(Test):
             self.cancel('Backing Device counts and vios name counts differ')
         self.backingdev_count = len(self.backing_adapter)
         self.bandwidth = self.params.get("bandwidth", '*', default=None)
+        self.vnic_priority = self.params.get(
+            "priority", '*', default=None)
+        if not self.vnic_priority:
+            self.vnic_priority = [50] * len(self.backing_adapter)
+        else:
+            self.vnic_priority = self.vnic_priority.split(',')
+        if len(self.vnic_priority) != len(self.backing_adapter):
+            self.cancel('Backing Device counts and priority counts differ')
+        self.auto_failover = self.params.get(
+            "auto_failover", '*', default=None)
+        if self.auto_failover not in ['0', '1']:
+            self.auto_failover = '1'
         self.count = int(self.params.get('vnic_test_count', default="1"))
         self.num_of_dlpar = int(self.params.get("num_of_dlpar", default='1'))
         self.device_ip = self.params.get('device_ip', '*', default=None)
@@ -328,7 +342,7 @@ class NetworkVirtualization(Test):
             for _ in range(self.count):
                 for val in range(int(self.backing_dev_count())):
                     self.log.info("Performing Client initiated\
-                                  failover - Attempt %s", int(val+1))
+                                  failover - Attempt %s", int(val + 1))
                     genio.write_file("/sys/devices/vio/%s/failover"
                                      % device_id, "1")
                     time.sleep(10)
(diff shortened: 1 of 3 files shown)


abdhaleegit:
[testlogs-vnic.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/3353843/testlogs-vnic.txt)

io/net/virt-net/network_virtualization.py
```diff
@@ -79,16 +79,26 @@ def setUp(self):
...
             if line in self.lpar:
                 self.server = line
+                break
         if not self.server:
             self.cancel("Managed System not got")
+        cmd = 'lssyscfg -r lpar -F name -m %s' % self.server
+        output = self.run_command(cmd)
+        for line in output:
+            if "%s-" % self.lpar in line:
+                self.lpar = line
```
narasimhan-v: break here as well

io/net/virt-net/network_virtualization.py
```diff
@@ -79,16 +79,26 @@ def setUp(self):
...
+        if not self.is_lpar_in_server(self.server, self.lpar):
```
narasimhan-v: We are finding the LPAR from server, in line 95.
So, this is unnecessary.

io/net/virt-net/network_virtualization.py
```diff
@@ -114,6 +114,10 @@ def setUp(self):
             self.cancel('Backing Device counts and vios name counts differ')
         self.backingdev_count = len(self.backing_adapter)
         self.bandwidth = self.params.get("bandwidth", '*', default=None)
+        self.vnic_priority = self.params.get("priority", '*', default=50).split(',')
```
narasimhan-v: Do not assign a default here.
If it is left empty, create a default list full of 50s.
If not empty, test the length, and use the values.
Makes sense ?

io/net/virt-net/network_virtualization.py
```diff
@@ -114,6 +114,10 @@ def setUp(self):
...
+        if len(self.vnic_priority) != len(self.backing_adapter):
+            self.cancel('Backing Device counts and priority counts differ')
+        self.auto_failover = int(self.params.get("auto_failover", '*', default=1))
```
narasimhan-v: It is better to evaluate this value after this, since user can give wrong values like 2 (example).
So, you can make it 0 if 0, and 1 if non-zero.

io/net/virt-net/network_virtualization.py
abdhaleegit: perfect, 
if empty : will create a list of 50s with x elements, where x=number of sriov_adapters
if notempty: I am already checking the length here 
        if len(self.vnic_priority) != len(self.backing_adapter):
            self.cancel('Backing Device counts and priority counts differ')

io/net/virt-net/network_virtualization.py
abdhaleegit: I will have a check for that, so for 0 it is 0
for 1 and non zero or empty it is 1

io/net/virt-net/network_virtualization.py
abdhaleegit: it is needed only to check if lpar is not migrated already ?

io/net/virt-net/network_virtualization.py
narasimhan-v: perfect

io/net/virt-net/network_virtualization.py
narasimhan-v: yes, mainly in LPM context. not necessary here.

io/net/virt-net/network_virtualization.py
narasimhan-v: yes, perfect. I had missed empty scenario.

abdhaleegit: done

abdhaleegit:
@narasimhan-v  Thanks for the review, I have all comments implemented

io/net/virt-net/network_virtualization.py
```diff
@@ -145,6 +155,18 @@ def get_mcp_component(component):
                 return line.split()[-1].strip('{}\"')
         return ''
 
+    def is_lpar_in_server(self, server, lpar):
```
narasimhan-v: This needs to be removed

io/net/virt-net/network_virtualization.py
```diff
@@ -450,17 +451,19 @@ def backing_dev_add_remove(self, operation, i):
...
+                                self.vnic_priority[i])
         if operation == 'add':
             cmd = 'chhwres -r virtualio --rsubtype vnic -o s -m %s -s %s \
-                   --id %s -a backing_devices+=%s' % (self.server,
-                                                      self.slot_num,
-                                                      self.lpar_id,
-                                                      add_backing_device)
+                   --id %s -a \"auto_priority_failover=%s,backing_devices+=%s\"' % (self.server,
+                                                                                    self.slot_num,
+                                                                                    self.lpar_id,
```
narasimhan-v: pep8 ?

abdhaleegit: yes autopep8 was run

io/net/virt-net/network_virtualization.py
abdhaleegit: its already removed with latest update

narasimhan-v: but it has to be removed via 1st commit, right ?

abdhaleegit:
[vNIC-priority-logs.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/3361091/vNIC-priority-logs.txt)

io/net/virt-net/network_virtualization.py
abdhaleegit: how do we do that ? squash ?

PR #1184, opened 2019-07-05
Enhancment of scsi_add_remove test script

Naresh-ibm:
existing test script cannot be run on vfc disk whic does not has pci_address.
and also, script doesn't has verification part after remove or added the scsi
disk. So modified the script to run using wwid os disk and also added the
verification part to it.


Naresh-ibm:
closing this in favour of #1188

PR #1187, opened 2019-07-08
Change print statements to log.info

harish-24:
Patch replaces print statements to log.info so that we can avoid both python2 and python3 compatibility

### DIFF
--- a/io/net/tcpdump.py
+++ b/io/net/tcpdump.py
@@ -64,9 +64,9 @@ class TcpdumpTest(Test):
         for line in process.run(cmd, shell=True,
                                 ignore_status=True).stderr.splitlines():
             if "packets dropped by interface" in line:
+                self.log.info(line)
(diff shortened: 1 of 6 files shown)


io/net/tcpdump.py
```diff
@@ -66,7 +66,7 @@ def test(self):
             if "packets dropped by interface" in line:
                 if int(line[0]) >= (int(self.drop) * int(self.count) / 100):
                     self.fail("%s, more than %s percent" % (line, self.drop))
-                print line
+                self.log.info(line)
```
narasimhan-v: @harish-24 though this should not be part of this commit, can you change it like:
```
            if "packets dropped by interface" in line:
                self.log.info(line)
                if int(line[0]) >= (int(self.drop) * int(self.count) / 100):
                    self.fail("%s, more than %s percent" % (line, self.drop))
```

harish-24: @narasimhan-v Done.

PR #1193, opened 2019-07-12
Add test to create namespaces with different modes with single region

harish-24:
Add test to create namespaces with different modes with single region

### DIFF
--- a/memory/ndctl.py
+++ b/memory/ndctl.py
@@ -183,6 +183,7 @@ class NdctlTest(Test):
         self.dist = distro.detect()
         self.package = self.params.get('package', default='upstream')
         self.cnt = self.params.get('namespace_cnt', default=4)
+        self.preserve_setup = self.params.get('preserve_change', default=False)
 
         if 'SuSE' not in self.dist.name:
             self.cancel('Unsupported OS %s' % self.dist.name)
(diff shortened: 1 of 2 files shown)


harish-24:
```# avocado run memory/ndctl.py:NdctlTest.test_multiple_ns_modes_region
...
 (1/1) memory/ndctl.py:NdctlTest.test_multiple_ns_modes_region: PASS (14.88 s)
JOB TIME   : 14.95 s```

File: memory/ndctl.py [hunk 1]
```diff
@@ -274,7 +274,9 @@ def test_namespace_modes(self):
         """
         failed_modes = []
         self.enable_region()
-        region = self.get_json(short_opt='-R')[0]
+        region = self.params.get('region', default=None)
```
narasimhan-v: Can't this be moved to setUp() ?

harish-24: yes, this particular line can be moved. but `self.get_json` in the line following cannot be used without enabling regions. So thought of keeping it together.

narasimhan-v: Ok. I see most of the tests are using this snippet. So the question.

PraveenPenguin: This  way we can assign variable in local scope if it not using any where else and avoid to use heap memory

PR #1194, opened 2019-07-15
Add option for size and mode to be used as input

harish-24:
Add option for size and mode to be used as input

### DIFF
--- a/memory/ndctl.py.data/ndctl.yaml
+++ b/memory/ndctl.py.data/ndctl.yaml
@@ -7,3 +7,10 @@ config:
             package : 'upstream'
         distro:
             package : 'distro'
+    mode_types: !mux
+        fsdax:
+            modes: 'fsdax'
+            size: '1G'
+        sector:
+            modes: 'sector'
+            size: '300M'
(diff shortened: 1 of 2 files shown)


memory/ndctl.py.data/ndctl.yaml
```diff
@@ -2,6 +2,8 @@ config:
     namespace_cnt: 4
     region: 'region0'
     preserve_change: True
+    modes:
```
narasimhan-v: @harish-24 does it makes sense to have the accepted modes multiplexed here, instead of using just one value from user ?
If yes, does the same applicable for size as well ?

harish-24: Yes, that is the idea. But did not want to expand with yaml here as I considered this yaml file just as a placeholder.

narasimhan-v: @harish-24 can you explain more please ?

harish-24: @narasimhan-v modes/size can be multiplexed. I added it here just to let the user know these options can be inputs. It really depends on the user whether it needs to be multiplexed or not. Okay, so I will add these are !mux inputs rather than one value so that user can have a better option to input.

PR #1208, opened 2019-07-18
Fix: size variable in ndctl test

harish-24:
Patch fixes size variable in ndctl test

### DIFF
--- a/memory/ndctl.py
+++ b/memory/ndctl.py
@@ -358,9 +358,9 @@ class NdctlTest(Test):
             self.log.info("Using %s for muliple namespaces", region)
             self.log.info("Creating %s namespaces", self.cnt)
             if not self.size:
-                size = self.get_json_val(self.get_json(
+                self.size = self.get_json_val(self.get_json(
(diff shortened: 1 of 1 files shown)


harish-24:
```
 (07/10) ndctl.py:NdctlTest.test_multiple_ns_modes_region:  PASS (15.43 s)
```

memory/ndctl.py
```diff
@@ -358,9 +358,9 @@ def test_multiple_ns_multiple_region(self):
             self.log.info("Using %s for muliple namespaces", region)
             self.log.info("Creating %s namespaces", self.cnt)
             if not self.size:
-                size = self.get_json_val(self.get_json(
+                self.size = self.get_json_val(self.get_json(
```
PraveenPenguin: one question why we need to change scope of variable as seems me it used only one place or it is bug fix ?

harish-24: yes, it is a fix.

PraveenPenguin: ok sure

PR #1211, opened 2019-07-22
[python3] Move memory tests to run with python3

harish-24:
Move memory tests to run with python3.
Tested on latest RHEL and SuSE releases.

### DIFF
--- a/memory/memhotplug.py
+++ b/memory/memhotplug.py
@@ -69,8 +69,8 @@ def get_hotpluggable_blocks(path, ratio):
         Return number of blocks in chunks of 100
         """
         if num % 2:
-            return num / 100 + 1
-        return num / 100
+            return num // 100 + 1
+        return num // 100
     count = chunks(len(mem_blocks) * ratio)
     return mem_blocks[:count]
 
@@ -150,7 +150,7 @@ class MemStress(Test):
         logs = process.system_output("dmesg -Txl 1,2,3,4").splitlines()
         for error in ERRORLOG:
             for log in logs:
-                if error in log:
+                if error in log.decode():
                     err_list.append(log)
         if "\n".join(err_list):
             collect_dmesg(self)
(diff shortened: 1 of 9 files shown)


PraveenPenguin:
@harish-24 can you please past relevant python3 log

harish-24:
[py3_mem.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/3425061/py3_mem.log) 
One of the hotplug test error is handled with the above push.
 (1/1) memory/memhotplug.py:MemStress.test_hotplug_toggle: PASS (194.45 s)

memory/memhotplug.py
```diff
@@ -189,14 +189,14 @@ def test_hotplug_toggle(self):
 
     def test_dlpar_mem_hotplug(self):
         if 'ppc' in platform.processor() and 'PowerNV' not in open('/proc/cpuinfo', 'r').read():
-            if "mem_dlpar=yes" in process.system_output("drmgr -C", ignore_status=True, shell=True):
+            if b"mem_dlpar=yes" in process.system_output("drmgr -C", ignore_status=True, shell=True):
```
narasimhan-v: wouldn't it be better to compare as string ?

harish-24: I am not sure, because we return bytes with `process.system_output`. Either it has to be decoded as string or used as is as bytes, both works though. Is it better to decode and use as string?

narasimhan-v: To minimise unknown breakages and also maintain a uniformity, it is better we decide either string to byte conversion or the inverse. And since all our tests are working with string better, I thought that would be better. There is nothing else suggesting which is better.

And this was just a suggestion. I am fine with either approach.

PraveenPenguin: i vote for `string to byte conversion ` as in avocado lib also we follow the same

PR #1217, opened 2019-07-25
[python3]Move fs tests to run with python3.

PraveenPenguin:
Move memory tests to run with python3 expect xfstest

### DIFF
--- a/fs/filebench.py
+++ b/fs/filebench.py
@@ -68,7 +68,7 @@ class Filebench(Test):
                                      'workloads', testfile)
         cmd = '%s -f %s' % (binary_path, testfile_path)
         out = process.system_output(cmd)
-        self.log.info("result:" + out)
+        self.log.info(b"result:" + out)
 
 
 if __name__ == "__main__":
--- a/fs/fs-fuzz.py
+++ b/fs/fs-fuzz.py
@@ -34,7 +34,7 @@ class FsFuzz(Test):
         process.run("dmesg -C ", sudo=True)
 
     def verify_dmesg(self):
-        self.whiteboard = process.system_output("dmesg")
+        self.whiteboard = process.system_output("dmesg").decode()
         pattern = ['WARNING: CPU:', 'Oops',
                    'Segfault', 'soft lockup', 'Unable to handle']
         for fail_pattern in pattern:
--- a/fs/fsshrink.py
+++ b/fs/fsshrink.py
@@ -33,7 +33,7 @@ class Fsshrink(Test):
         process.run("dmesg -C ", sudo=True)
 
     def verify_dmesg(self):
-        self.whiteboard = process.system_output("dmesg")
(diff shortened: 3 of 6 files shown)


fs/filebench.py
```diff
@@ -68,7 +68,7 @@ def test(self):
                                      'workloads', testfile)
         cmd = '%s -f %s' % (binary_path, testfile_path)
         out = process.system_output(cmd)
-        self.log.info("result:" + out)
+        self.log.info(b"result:" + out.decode())
```
narasimhan-v: isn't `out.encode()` better ?

harish-24:
> @harish-24 in general, which is better ? b'' or ''.decode() ?

@narasimhan-v IIUC both `b"` and `.encode()` have the same impact in this context.

narasimhan-v:
@harish-24 What I suggest is, use either one so that we do not have confusion later.

fs/filebench.py
harish-24: @PraveenPenguin why is a **byte** `b"result"`appended with a **string** `out.decode()`. Isn't one change sufficient here? Only either of those needs to be converted anyways.

PraveenPenguin: @harish yes you are right , will change and push the changes

fs/fsshrink.py
```diff
@@ -37,7 +37,7 @@ def verify_dmesg(self):
         pattern = ['WARNING: CPU:', 'Oops',
                    'Segfault', 'soft lockup', 'Unable to handle']
         for fail_pattern in pattern:
-            if fail_pattern in self.whiteboard:
+            if fail_pattern in self.whiteboard.decode():
```
harish-24: same here

harish-24:
> @harish-24 What I suggest is, use either one so that we do not have confusion later.

@narasimhan-v @PraveenPenguin Okay, so we can use `b"` moving forward as some of my merged patches already use it.

fs/fs-fuzz.py
```diff
@@ -38,7 +38,7 @@ def verify_dmesg(self):
         pattern = ['WARNING: CPU:', 'Oops',
                    'Segfault', 'soft lockup', 'Unable to handle']
         for fail_pattern in pattern:
-            if fail_pattern in self.whiteboard:
+            if fail_pattern in self.whiteboard.decode():
```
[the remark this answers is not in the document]
PraveenPenguin: @harish-24 yes i know it require library patch as well as you we are in process of send that as well , but test supporting python3 still  be ok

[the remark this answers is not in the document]
harish-24: @PraveenPenguin It does not require library because `genio` expects string as input. So test-case has to change. There is no need for a library patch.

PraveenPenguin:
@harish-24 @narasimhan-v pushed the changes please have look

PR #1233, opened 2019-07-30
Enhancment of scsi_add_remove test script

Naresh-ibm:
existing test script cannot be run on vfc disk which does not has pci_address.
and also, script doesn't has verification part after remove or added the scsi
disk. So modified the script to run using wwid or pci domains and also added
the verification part to it.

### DIFF
--- a/io/disk/scsi_add_remove.py
+++ b/io/disk/scsi_add_remove.py
@@ -14,13 +14,17 @@
 # Copyright: 2017 IBM
 # Author: Narasimhan V <sim@linux.vnet.ibm.com>
 # Author: Venkat Rao B <vrbagal1@linux.vnet.ibm.com>
+# Author: Naresh Bannoth <nbannoth@in.ibm.com>
 """
 This script will perform scsi add and remove test case
 """
 import time
+from avocado import main
 from avocado import Test
 from avocado.utils import process, genio
 from avocado.utils.software_manager import SoftwareManager
+from avocado.utils import multipath
+from avocado.utils import pci
 
 
 class ScsiAddRemove(Test):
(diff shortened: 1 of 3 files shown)


Naresh-ibm:
[scsi_add_remove_job_log.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/3445588/scsi_add_remove_job_log.txt)

io/disk/scsi_add_remove.py.data/README.txt
```diff
@@ -1,7 +1,8 @@
 This Test removes and adds back a scsi device in all the specified PCI domains
-specified in the 'multiplexer' file.
+or wwids specified in the 'multiplexer' file.
```
narasimhan-v: @Naresh-ibm also mention here:
if wwids are provided, pci devices are ignored.

io/disk/scsi_add_remove.py
```diff
@@ -33,44 +37,90 @@ def setUp(self):
...
+            self.log.info("\nDeleting %s = %s\n" % (device, scsi_id))
+            genio.write_file("/sys/block/%s/device/delete" % device, "1")
+            time.sleep(5)
+            if self.is_exists_scsi_device(device) is True:
+                self.err_paths.append(device)
+            else:
+                self.log.info("\n%s = %s deleted\n" % (device, scsi_id))
+            self.log.info("\nadding back %s = %s\n" % (device, scsi_id))
+            part = "scsi add-single-device %s" % scsi_id
+            genio.write_file("/proc/scsi/scsi", part)
```
narasimhan-v: Is this same as `process.run("echo scsi add-single-device %s > /proc/scsi/scsi", scsi_id)` ?

narasimhan-v: Just asking for confirmation.

io/disk/scsi_add_remove.py
```diff
@@ -33,44 +37,90 @@ def setUp(self):
...
+        self.device_list = []
         smm = SoftwareManager()
         if not smm.check_installed("lsscsi") and not smm.install("lsscsi"):
             self.cancel("lsscsi is not installed")
-        self.pci_device = self.params.get("pci_device", default=None)
-        if not self.pci_device:
-            self.cancel("Please provide PCI address for which you \
-                        want to run the test")
+        self.wwids = self.params.get('wwids', default='')
+        self.pci_device = self.params.get("pci_device", default='').split(',')
```
narasimhan-v: The split should happen after checking for presence, similar to self.wwids

io/disk/scsi_add_remove.py
Naresh-ibm: yes

io/disk/scsi_add_remove.py
Naresh-ibm: it works fine even if we have the split at this place. 
So I think it should be fine.

narasimhan-v: @Naresh-ibm 
```
>>> bool(''.split(','))
True
>>> bool('')
False
>>>
```

Naresh-ibm:
closing this in favour of #1249

PR #1260, opened 2019-08-19
 [python3] Enable libhugetlbfs test to run with python3

harish-24:
*  [python3] Enable libhugetlbfs test to run with python3
Patch changes python3-enabled branch of libhugetlbfs and few other
fixes to run test with python3.

* Fix: Removed unused variable
Patch removes an unnecessary line from the test

### DIFF
--- a/memory/libhugetlbfs.py
+++ b/memory/libhugetlbfs.py
@@ -108,7 +108,7 @@ class LibHugetlbfs(Test):
                             % (pages_available, pages_requested))
 
         git.get_repo('https://github.com/libhugetlbfs/libhugetlbfs.git',
-                     destination_dir=self.workdir)
+                     branch='next', destination_dir=self.workdir)
         os.chdir(self.workdir)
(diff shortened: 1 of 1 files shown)


memory/libhugetlbfs.py
```diff
@@ -108,7 +107,7 @@ def setUp(self):
                             % (pages_available, pages_requested))
 
         git.get_repo('https://github.com/libhugetlbfs/libhugetlbfs.git',
-                     destination_dir=self.workdir)
+                     branch='next', destination_dir=self.workdir)
```
PraveenPenguin: @harish-24 so next branch have all the patches which able to run test suite in python 3 ?

harish-24: @PraveenPenguin yes, that is right. Once it is merged into master (expected to be least 3 months) we can move to master back.

PraveenPenguin: @harish-24 sure , will try once and merge

PR #1261, opened 2019-08-22
Add test to verify nvdimm-numa node association

harish-24:
Patch adds tests to verify numa nodes associated per region with its sysfs interface and ndctl command line

### DIFF
--- a/memory/ndctl.py
+++ b/memory/ndctl.py
@@ -20,6 +20,7 @@ This Suite works with various options of ndctl on a NVDIMM device.
 """
 
 import os
+import re
 import json
 import glob
 from avocado import Test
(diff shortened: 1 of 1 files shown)


harish-24:
```
 (1/1) memory/ndctl.py:NdctlTest.test_check_numa: PASS (0.30 s)
```

harish-24:
```
 (1/1) memory/ndctl.py:NdctlTest.test_labels: PASS (1.64 s)
```

memory/ndctl.py
```diff
@@ -418,6 +419,57 @@ def test_check_namespace(self):
...
+
+        self.log.info("Re-storing labels with a namespace")
+        if process.system('%s write-labels %s -i output' % (self.binary, nmem), shell=True):
+            self.fail("Label write failed")
+        self.enable_region(name=region)
+
+        self.log.info("Checking mismatch after restore")
+        new_op = process.system_output('%s check-labels %s' % (self.binary, nmem), shell=True)
+        if new_op != old_op:
+            self.fail("Label read and write mismatch")
```
narasimhan-v: @harish-24 can we have a function which can be called twice (before and after destroying namespace), instead of repeating code for some of the operations ?

harish-24: @narasimhan-v we have the following operations.
```
disable-region
zero-labels
enable-region
create
check-labels
read-labels
disable-namespace
destroy-namespace
disable-region
zero-labels
write-labels
enable-region
check-labels
```
you mean to have functions for individual tasks and call in the same order as above?

Issue #1262, opened 2019-08-22
kernel/kernbench.py found errors while building vmlinux kernel.


Joshhsi:
While running the test case kernel/kernbench.py, there is an error occurred during the process of building a vmlinux kernel.

The error message is as follows :
make[1]: *** No rule to make target 'certs/rhel.pem' , needed by 'certs/x509_certificate_list' . Stop.

And I've learned that the problem could be solved by comment out the lines CONFIG_SYSTEM_TRUSTED_KEY and CONFIG_MODULE_SIG_KEY in boot/config-4.18.0-80.7.2.el8_0.x86_64.

So the olddefconfig would copy the config that I modified and pass the test.
The concern is - is it a reasonable workaround to pass the test?

PraveenPenguin:
@Joshhsi yes that is valid work around as kernbench test is like a  work load as kernel building and in test we are capturing time of kernel build operation ,is this ok to you  send PR as fix let me know either case i will do

Joshhsi:
@PraveenPenguin Thanks for clarifying my thought, issue closed.

PraveenPenguin:
@Joshhsi sure , so are you going to send fix ?

kyle-ibm:
@PraveenPenguin we seem to be hitting this same issue in py2-lts branch
```
[stderr] make[2]: *** No rule to make target `certs/rhel.pem', needed by `certs/x509_certificate_list'.  Stop.
[stderr] make[1]: *** [certs] Error 2
Command '/usr/bin/time -o /var/tmp/avocado_Gdocr6/avocado_job_pSQT8F/6-kernbench.py_Kernbench.test_long-default-custom-636c/linux-master/time_file make -j 1 vmlinux' finished with 2 after 312.751744032s
[stderr] make: *** [sub-make] Error 2

Reproduced traceback from: /usr/lib/python2.7/site-packages/avocado_framework-69.0-py2.7.egg/avocado/core/test.py:844
Traceback (most recent call last):
  File "/root/tests/tests/avocado-misc-tests/kernel/kernbench.py", line 133, in test
    self.time_build(self.threads, timefile, "")
  File "/root/tests/tests/avocado-misc-tests/kernel/kernbench.py", line 60, in time_build
    self.fail("No vmlinux found, kernel build failed")
  File "/usr/lib/python2.7/site-packages/avocado_framework-69.0-py2.7.egg/avocado/core/test.py", line 1030, in fail
[...]

PR #1266, opened 2019-08-29
Fix for issue #1262

Joshhsi:
While running the test case kernel/kernbench.py, there is an error occurred during the process of building a vmlinux kernel.

The error message is as follows :
make[1]: *** No rule to make target 'certs/rhel.pem' , needed by 'certs/x509_certificate_list' . Stop.

This fix is to comment out the lines CONFIG_SYSTEM_TRUSTED_KEY and CONFIG_MODULE_SIG_KEY in boot/config during the test.

So the process "make olddefconfig" would copy the config that already comments out certification sections and building kernel without errors.

### DIFF
--- a/kernel/kernbench.py
+++ b/kernel/kernbench.py
@@ -21,6 +21,7 @@
 import os
 import re
 import platform
+import subprocess
 
 from avocado import Test
 from avocado import main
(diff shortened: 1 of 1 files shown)


kernel/kernbench.py
```diff
@@ -38,7 +39,11 @@ class Kernbench(Test):
     and compile it as per the configuration file. Performance figures will be
     shown in the log file.
     """
-
+    subprocess.call("cp /boot/config* /tmp", shell=True)
```
PraveenPenguin: using avocado library will be preferred  as more granular way as example it provide error handling internally 

example can be 
`process.system("cp /boot/config* /tmp", shell=True , sudo=true)`

PraveenPenguin:
@Joshhsi thanks for providing fix , please make PR as pep8 complaint   as `kernbench.py:46:1: W293 blank line contains whitespace` need to be resolve and please add Signed-off-by tag also in commit message as `git commit -s` can be used

Joshhsi:
@PraveenPenguin Thanks for the guiding, I've changed the library to avocado and meet the pep8 compliant. It seems the remaining problem is "Signed-off-by" tag, right?

PraveenPenguin:
> @PraveenPenguin Thanks for the guiding, I've changed the library to avocado and meet the pep8 compliant. It seems the remaining problem is "Signed-off-by" tag, right?

@Joshhsi yes only issue with "Signed-off-by" and even it has five commit we need to squash all the commit in single commit

PR #1267, opened 2019-08-30
fix build issue on RHEL

Joshhsi:
The error message is as follows :
make[1]: *** No rule to make target 'certs/rhel.pem' , needed by 'certs/x509_certificate_list' . Stop
This fix is to comment out the lines CONFIG_SYSTEM_TRUSTED_KEY and CONFIG_MODULE_SIG_KEY in boot/config during the test.

So the process "make olddefconfig" would copy the config that already comments out certification sections and building kernel without errors.

Signed-off-by: Chih Yuan Hsi <gofar.hsi@gmail.com>

### DIFF
--- a/kernel/kernbench.py
+++ b/kernel/kernbench.py
@@ -98,6 +98,10 @@ class Kernbench(Test):
             deps.extend(['popt', 'glibc', 'glibc-devel', 'libcap-ng',
                          'libcap', 'libcap-devel', 'elfutils-libelf',
                          'elfutils-libelf-devel', 'openssl-devel'])
+            process.system("cp /boot/config* /tmp", shell=True, sudo=True)
+            process.system("sed -i 's/^.*CONFIG_SYSTEM_TRUSTED_KEYS/#&/g' /boot/config*", shell=True, sudo=True)
+            process.system("sed -i 's/^.*CONFIG_SYSTEM_TRUSTED_KEYRING/#&/g' /boot/config*", shell=True, sudo=True)
+            process.system("sed -i 's/^.*CONFIG_MODULE_SIG_KEY/#&/g' /boot/config*", shell=True, sudo=True)
 
         for package in deps:
             if not smg.check_installed(package) and not smg.install(package):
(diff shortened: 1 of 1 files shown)


harish-24:
@Joshhsi I believe these changes appear to be only on `RHEL`, could you handle the same in the patch? `i.e` we should not be doing these steps on other distributions.

kernel/kernbench.py
```diff
@@ -38,6 +38,10 @@ class Kernbench(Test):
     and compile it as per the configuration file. Performance figures will be
     shown in the log file.
     """
+    process.system("cp /boot/config* /tmp", shell=True, sudo=True)
```
PraveenPenguin: it can be like as @harish-24 mentioned 

it can be applicbale for all rhel and rhel origin distro 
```
if detected_distro.name in ['centos', 'fedora', 'rhel']: 
   process.system("cp /boot/config* /tmp", shell=True, sudo=True)
   process.system("sed -i 's/^.*CONFIG_SYSTEM_TRUSTED_KEYS/#&/g' /boot/config*", shell=True, sudo=True)
   process.system("sed -i 's/^.*CONFIG_SYSTEM_TRUSTED_KEYRING/#&/g' /boot/config*", shell=True, sudo=True)
   process.system("sed -i 's/^.*CONFIG_MODULE_SIG_KEY/#&/g' /boot/config*", shell=True, sudo=True)
```

PraveenPenguin: please move this on setup section as well

Joshhsi: Thanks @harish-24 and @PraveenPenguin , I've changed the fix to be rhel or rhel origin distros specific.

harish-24:
@jayarama-ibm make use of the class object `self` to define variable and do not duplicate same code across methods. And also please add comment above the steps on why we can arrive at that workaround.

PraveenPenguin:
> @Joshhsi make use of the class object `self` to define variable and do not duplicate same code across methods. And also please add comment above the steps on why we can arrive at that workaround.

@Joshhsi can you please address @harish-24 comment

PraveenPenguin:
closing in favour of #1337

PR #1280, opened 2019-09-09
[python3]move ras bucket  to python3 compliant

PraveenPenguin:
[python3]move ras bucket  to python3 compliant


ras/ras.py
```diff
@@ -188,13 +192,15 @@ def test9_ofpathname(self):
...
+                                     " tail -1 | cut -d' ' -f1").strip()
+        self.run_cmd("ofpathname -V")
+        disk_name = self.run_cmd_out("df -h | egrep '(s|v)da[1-8]' | "
+                                     "tail -1 | cut -d' ' -f1").strip()
         if disk_name:
             self.run_cmd("ofpathname %s" % disk_name)
-            of_name = process.system_output("ofpathname %s"
-                                            % disk_name).strip()
+            of_name = self.run_cmd_out("ofpathname %s"
+                                       % disk_name).strip()
```
narasimhan-v: Is it better for `run_cmd_out()` to return the output after a `strip()` ?

ras/ras_extended.py
```diff
@@ -206,14 +210,14 @@ def test3_lsvpd(self):
...
+        path_db = self.run_cmd_out("find /var/lib/lsvpd/ -iname vpd.db | "
+                                   "head -1").strip()
         if path_db:
             copyfile_path = os.path.join(self.outputdir, 'vpd.db')
             shutil.copyfile(path_db, copyfile_path)
             self.run_cmd("lsvpd --path=%s" % copyfile_path)
-        path_tar = process.system_output("find /var/lib/lsvpd/ -iname vpd.*.gz"
-                                         " | head -1", shell=True).strip()
+        path_tar = self.run_cmd_out("find /var/lib/lsvpd/ -iname vpd.*.gz"
+                                    " | head -1").strip()
```
narasimhan-v: Is it better for `run_cmd_out()` to return the output after a `strip()` ?

ras/ras.py
PraveenPenguin: sure make sense will adopt it

ras/ras_extended.py
PraveenPenguin: @narasimhan-v here we can not adopt as in `test1_lsdevinfo` we are striping using a pattern if you say I can pass another argument to `run_cmd_out()` but I think earlier implementation hold good

PraveenPenguin:
@narasimhan-v pushed changes where i thought it can adopted , please have a look , for other comment please suggest

ras/ras_extended.py
narasimhan-v: @PraveenPenguin `run_cmd_out()` can still `strip()`, and `test1_lsdevinfo()` can do another `strip('12345')` ?
Let me know if that works fine for you.

PraveenPenguin: @narasimhan-v sure I think it should work, let me try and accordingly update here

PraveenPenguin:
> @PraveenPenguin Also, would be helpful if you can convert all bash commands into python equivalents after this is done.

@narasimhan-v yes that is next step as all ras test cases need to adopt python way will do in some time as right now i am focusing to move it python3 compliant, is it that sounds you ok

ras/ras_extended.py
narasimhan-v: @PraveenPenguin except for this clarification, I am fine with the code.

PraveenPenguin:
@narasimhan-v pushed the changes please have a look

ras/ras_extended.py
PraveenPenguin: works well , please have look

PR #1296, opened 2019-09-25
Porting the module_unload_load test to python from shell

Naresh-ibm:
porting the module_unload_load test from shell to python.
Enhanced few functionality like for FC driver flushing the multipath
and then unload the module which was not there in shell script.


Naresh-ibm:
[module_load_unload_only_io_True_job_log.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/3650552/module_load_unload_only_io_True_job_log.txt)
[module_load_unload_single_module_job_log.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/3650553/module_load_unload_single_module_job_log.txt)

io/driver/module_unload_load.py
```diff
@@ -0,0 +1,139 @@
...
+                            process.system("multipath -F", ignore_status=True)
+                            cmd = "lsmod | grep -i ^%s" % mdl
+                            for _ in range(20):
+                                time.sleep(3)
+                                out = process.getoutput(cmd).split(" ")[-1]
+                                if out.strip() != '0':
+                                    self.log.info("command value : %s\n" % out)
+                                    process.system("multipath -F",
+                                                   ignore_status=True)
+                                    continue
```
narasimhan-v: this is needless here.

io/driver/module_unload_load.py
```diff
@@ -0,0 +1,139 @@
...
+                    for mod in sub_mod:
+                        if mod == 'multipath':
+                            process.system("multipath -F", ignore_status=True)
+                            cmd = "lsmod | grep -i ^%s" % mdl
+                            for _ in range(20):
+                                time.sleep(3)
+                                out = process.getoutput(cmd).split(" ")[-1]
+                                if out.strip() != '0':
+                                    self.log.info("command value : %s\n" % out)
+                                    process.system("multipath -F",
```
narasimhan-v: why doing `multipath -F` twice ?

io/driver/module_unload_load.py
```diff
@@ -0,0 +1,139 @@
...
+                                                   ignore_status=True)
+                                    continue
+                                else:
+                                    break
+                        else:
+                            self.log.info("unloading sub module %s " % mod)
+                            process.system("rmmod %s" % mod)
```
narasimhan-v: no utility to do this ?

io/driver/module_unload_load.py
```diff
@@ -0,0 +1,139 @@
...
+                self.log.info("unloading module %s " % mdl)
+                process.system("rmmod %s" % mdl)
```
narasimhan-v: no utility to do this ?

io/driver/module_unload_load.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        Begining the test here
+        """
+        pci_addrs = []
+        if self.module:
+            for modl in self.module:
+                self.mod_list.append(modl)
+        elif self.only_io is True:
+            pci_addrs = pci.get_pci_addresses()
+            for pci1 in pci_addrs:
```
narasimhan-v:
```
elif self.only_io is True: 
    for pci_addr in pci.get_pci_addresses():
        self.mod_list.append(pci.get_driver(pci_addr))
```

io/driver/module_unload_load.py
```diff
@@ -0,0 +1,139 @@
...
+            if linux_modules.module_is_loaded(mod1) is False:
+                linux_modules.load_module(mod1)
+                time.sleep(self.load_unload_sleep_time)
+
+        for _ in range(0, self.iteration):
+            for mdl in module_list:
+                sub_mod = self.get_depend_modules(mdl)
+                if sub_mod:
+                    sub_mod = sub_mod.split(' ')
+                    for mod in sub_mod:
```
narasimhan-v: for mod in sub_mod.split(' '):

io/driver/module_unload_load.py
narasimhan-v: This block handling multipath can be a separate function. And it can be handled a lot better

io/driver/module_unload_load.py.data/README.txt
```diff
@@ -1,13 +1,9 @@
...
-
-The test will PASS when at least one of the drivers finishes the whole loop. 
+This Program loads and unloads the kernel driver modules. For the kernel driver modules that have dependencies, the dependant modules should  be listed in the config file (config). The script takescare of unloading the dependant modules, if any, before unloading the core module. It runs for given number iterations.
 
 This script should be run as root.
 
 Inputs
 ------
-ITERATIONS -    No of counts to unload and load the module. Defaults to 10.
+ITERATIONS -    No of counts to unload and load the module. Defaults to 1.
```
narasimhan-v: What is the reason for defaulting to 1 ?

io/driver/module_unload_load.py
```diff
@@ -0,0 +1,139 @@
...
+                driver = process.getoutput(line.split('/')[-1])
+                self.mod_list.append(driver.split('.')[0])
+
+        for mod in self.mod_list:
+            if self.built_in_module(mod) is True:
+                self.mod_list.remove(mod)
+        self.module_load_unload(self.mod_list)
+
+        if self.error_modules:
+            self.error("few modules failed to load unload, please check logs")
```
narasimhan-v: self.fail or self.warn is better here

io/driver/module_unload_load.py.data/README.txt
Naresh-ibm: if it is for all the module then doing 10 times is really takes lot of time. hence made it to 1.
I think that should not matter here, as we have given a parameter to give the count.

io/driver/module_unload_load.py
Naresh-ibm: sure will change this to self.fail or self.warn

io/driver/module_unload_load.py
Naresh-ibm: if we use this, this will end up calling " pci.get_pci_addresses():" each time.
so making it to static. so felt making static if better than calling a function multiple time.

io/driver/module_unload_load.py
Naresh-ibm: the utility function is still failing so used the rmmod. will open a issue for same.

io/driver/module_unload_load.py
Naresh-ibm: ok, will make a separate function for this.

io/driver/module_unload_load.py
Naresh-ibm: will handle in separate function.

io/driver/module_unload_load.py
narasimhan-v: yes please

io/driver/module_unload_load.py
narasimhan-v: we will still call it only once.

io/driver/module_unload_load.py.data/README.txt
narasimhan-v: When changing existing values, it is always better to have reason documented. This discussion here is that documentation

io/driver/module_unload_load.py
narasimhan-v: @Naresh-ibm 
check if this works:
```
      cmd = "lsmod | grep -i ^%s" % mdl
      for i in range(20):
          if process.getoutput(cmd).split(" ")[-1] == '0':
              return True
          process.system("multipath -F", ignore_status=True)
          time.sleep(3)
          continue
      return False
```

narasimhan-v: @Naresh-ibm I would also suggest converting the bash command to pythonic way

Naresh-ibm:
closing this in favour of #1335

PR #1304, opened 2019-09-30
Fix perf c2c

shirishaganta:


PraveenPenguin:
@shirishaganta please wrap all commit in single commit

PR #1327, opened 2019-10-07
[Back-port] Added the Manual pre-requisite steps for avago9361 - v3

manvanthar:
Added the Manual pre-requisite steps to Automation to help run script in one go.

### DIFF
--- a/io/disk/Avago_storage_adapter/avago9361_vd.py
+++ b/io/disk/Avago_storage_adapter/avago9361_vd.py
@@ -84,6 +84,20 @@ class Avago9361(Test):
         self.stripe = [64, 128, 256, 512, 1024]
         self.state = ['start', 'stop', 'start', 'pause', 'resume']
         self.list_test = ['online_offline', 'rebuild']
+        cmd = "%s /c%d set jbod=on" % (self.tool, self.controller)
+        if process.system(cmd, shell=True, ignore_status=True) != 0:
+            self.log.info(" Good.. JBOD mode already set, Ignore the Error as it is pre-setup")
+        time.sleep(5)
+        cmd = "%s /c%d/vall delete force" % (self.tool, self.controller)
+        if process.system(cmd, shell=True, ignore_status=True) != 0:
+            self.log.info(" As required,  No Virtual Drives are Available. Ignore the Error as it is pre-setup")
+        cmd = "%s /c%d set jbod=off" % (self.tool, self.controller)
+        if process.system(cmd, shell=True, ignore_status=True) != 0:
+            self.log.info("As required, Controller is already in non-JBOD mode!!, Ignore the Error as it is pre-setup")
+        cmd = "%s /c%d/eall/sall set good force" % (self.tool, self.controller)
+        if process.system(cmd, shell=True, ignore_status=True) != 0:
+            self.log.info("As required Drives are Already Set to good!!, Ignore the Error as it is pre-setup")
+        time.sleep(10)
         if str(self.name) in self.list_test:
             cmd = "%s /c%d set autorebuild=off" % (self.tool, self.controller)
             self.check_pass(cmd, "Failed to set auto rebuild off")
@@ -146,12 +160,8 @@ class Avago9361(Test):
         self.check_pass(cmd, "Failed to set JBOD on")
         cmd = "%s /c%d show jbod" % (self.tool, self.controller)
         self.check_pass(cmd, "Failed to show the JBOD status")
-        cmd = "%s /c%d/eall/sall set jbod" % (self.tool, self.controller)
-        self.check_pass(cmd, "Failed to convert the drives to JBOD")
         cmd = "%s /c%d show " % (self.tool, self.controller)
         self.check_pass(cmd, "Failed to show the adapter details")
-        cmd = "%s /c0/eall/sall set good force" % self.tool
-        self.check_pass(cmd, "Failed to set JBOD drives to good")
 
     def test_online_offline(self):
         """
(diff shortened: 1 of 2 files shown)


narasimhan-v:
@abdhaleegit review please

io/disk/Avago_storage_adapter/avago9361_vd.py
```diff
@@ -146,12 +160,8 @@ def test_jbod(self):
         self.check_pass(cmd, "Failed to set JBOD on")
         cmd = "%s /c%d show jbod" % (self.tool, self.controller)
         self.check_pass(cmd, "Failed to show the JBOD status")
-        cmd = "%s /c%d/eall/sall set jbod" % (self.tool, self.controller)
-        self.check_pass(cmd, "Failed to convert the drives to JBOD")
         cmd = "%s /c%d show " % (self.tool, self.controller)
         self.check_pass(cmd, "Failed to show the adapter details")
-        cmd = "%s /c0/eall/sall set good force" % self.tool
```
narasimhan-v: This test should enable jbod mode, and disable while exiting.

io/disk/Avago_storage_adapter/avago9361_vd.py
```diff
@@ -84,6 +84,20 @@ def setUp(self):
         self.stripe = [64, 128, 256, 512, 1024]
         self.state = ['start', 'stop', 'start', 'pause', 'resume']
         self.list_test = ['online_offline', 'rebuild']
+        cmd = "%s /c%d set jbod=on" % (self.tool, self.controller)
+        if process.system(cmd, shell=True, ignore_status=True) != 0:
```
narasimhan-v: We have a separate function `self.check_pass`. Can you use that please.

io/disk/Avago_storage_adapter/avago9361_vd.py
```diff
@@ -84,6 +84,20 @@ def setUp(self):
...
+        cmd = "%s /c%d/vall delete force" % (self.tool, self.controller)
+        if process.system(cmd, shell=True, ignore_status=True) != 0:
+            self.log.info(" As required,  No Virtual Drives are Available. Ignore the Error as it is pre-setup")
+        cmd = "%s /c%d set jbod=off" % (self.tool, self.controller)
+        if process.system(cmd, shell=True, ignore_status=True) != 0:
+            self.log.info("As required, Controller is already in non-JBOD mode!!, Ignore the Error as it is pre-setup")
+        cmd = "%s /c%d/eall/sall set good force" % (self.tool, self.controller)
+        if process.system(cmd, shell=True, ignore_status=True) != 0:
+            self.log.info("As required Drives are Already Set to good!!, Ignore the Error as it is pre-setup")
+        time.sleep(10)
```
narasimhan-v: Can we have separate functions for jbod enable and disable, so that we can re-use in `test_jbod` also ?

narasimhan-v: Also, I would suggest going for better options than sleep(X)

io/disk/Avago_storage_adapter/avago9361_vd.py
manvanthar: @narasimhan-v  no these are pre-requisuite runs to bring the system in runnable state.  The output of this are not important to exit the tests which are handled via self.check_pass.

However let me know if self.check_pass can be extended to include informational messages only?  that can be done.

io/disk/Avago_storage_adapter/avago9361_vd.py
manvanthar: hmmm No it would still break the code flow.

JBOD=on is used indirectly to clear any leftovers from the RAID creation before running an scenario.

JBOD=off is needed before running RAID level tests.

Eventhough on an outlook it looks as if we really need to do that.  But for the code flow perspective depending on what state the system is this would be essential.

narasimhan-v:
@manvanthar 

1. test_jbod - can this be moved to avago9361.py ?

2. `setting jbod=on, delete force, set jbod=off, set good force` Can this sequence be made into a function ?

io/disk/Avago_storage_adapter/avago9361_vd.py
abdhaleegit: @manvanthar I see `check_pass(self, cmd, errmsg)`  has the option to pass the error message.. so all `if process.system(` conditions here can be replaced by check_pass call

io/disk/Avago_storage_adapter/avago9361_vd.py
abdhaleegit: @manvanthar you can use avocado `wait.py` wait module to wait for command to complete. like

`wait.wait_for(self.check_pass(cmd, errmsg), 10)`

manvanthar: @abdhaleegit   The problem is not with command completion.   command completes immediatly.

  The card takes some random sec to finish of its work before accepting the next of commands.  If we dont wait for few seconds,  the next set of commands fails intermitently.

io/disk/Avago_storage_adapter/avago9361_vd.py
manvanthar: @abdhaleegit i already responded up and explained on what can be done?.   Are you guys ok with the changes to check_pass

io/disk/Avago_storage_adapter/avago9361_vd.py
abdhaleegit: as discussed, the wait here is required for the internal state change of the device, and user could not track the internal state change.. so wait here is valid.

manvanthar:
Fixed in PR# #1354

PR #1329, opened 2019-10-07
[Backport] Fix: Remove unnecessary package

harish-24:
Patch remove package which is not available in rhel8 distro

### DIFF
--- a/fs/xfstests.py
+++ b/fs/xfstests.py
@@ -78,13 +78,15 @@ class Xfstests(Test):
                              'libattr-devel', 'libaio-devel', 'libuuid-devel',
                              'openssl-devel', 'xfsprogs-devel'])
 
-            if self.detected_distro.name == 'rhel' and\
-                    self.detected_distro.version.startswith('8'):
-                packages.remove('indent')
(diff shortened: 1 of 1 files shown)


fs/xfstests.py
```diff
@@ -78,14 +78,16 @@ def setUp(self):
...
 
-            if self.detected_distro.name == 'rhel' and\
-                    self.detected_distro.version.startswith('8'):
-                packages.remove('indent')
             if self.detected_distro.name == 'SuSE':
                 packages.extend(['libbtrfs-devel', 'libcap-progs'])
             else:
                 packages.extend(['btrfs-progs-devel'])
 
+            if self.detected_distro.name == 'rhel' and\
```
narasimhan-v: Why moving the lines ?

harish-24: Package is getting added here: https://github.com/avocado-framework-tests/avocado-misc-tests/blob/master/fs/xfstests.py#L84

narasimhan-v: Ah ok, missed the removal of btrfs...

PR #1332, opened 2019-10-08
Fixing timeout issue with few NIC drivers during ethtool …

manvanthar:
…test.

Fixing timeout issue with few NIC drivers during ethtool test.

### DIFF
--- a/io/net/ethtool_test.py
+++ b/io/net/ethtool_test.py
@@ -23,6 +23,7 @@ This test needs to be run as root.
 """
 
 import netifaces
+import time
 from avocado import main
 from avocado import Test
 from avocado.utils.software_manager import SoftwareManager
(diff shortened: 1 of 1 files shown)


narasimhan-v:
@vaishnavibhat review please.

io/net/ethtool_test.py
```diff
@@ -70,6 +71,16 @@ def interface_state_change(self, interface, state, status):
         cmd = "ip link set dev %s %s" % (interface, state)
         if process.system(cmd, shell=True, ignore_status=True) != 0:
             return False
+        if state == "up":
+            for _ in range(0, 120, 10):
```
abdhaleegit: Can we use wait module in avocado utils to wait for function to return true 
wait.wait_for(check_state(interface), 30)

as per discussion with murilo a 30 sec wait is sufficient for any adapter link to come up

manvanthar: @abdhaleegit this has been tested...  30 sec is an random assumtions.  tg3 driver takes from 20 sec to 2 min to complete based on what ethtool parameter is changed.

manvanthar: btw wait module is not fully functional!!. so cant use it.

manvanthar: the logic checks the status of interface from 10 sec to 2 min. so there is no full wait of the system indefintly or some random 30 sec waiting .

manvanthar: Abdul,  Let me know if you ok with the justification

narasimhan-v: @manvanthar Also, if this is a generic solution, can this be pushed to avocado utils ?

@abdhaleegit your comments ?

abdhaleegit: the wait.wait_for is used to wait for command to complete .. but the logic here is really waiting till the network link is up.. so we can go ahead with this logic.. in future a avocado util to check link up will be better.

abdhaleegit: we can push this till than

narasimhan-v:
@abdhaleegit Thanks for ack. Merging.

PR #1334, opened 2019-10-09
Fixed Bonding tests for infiniband based interface

manvanthar:
Fixed Bonding tests for infiniband based interfaces

### DIFF
--- a/io/net/bonding.py
+++ b/io/net/bonding.py
@@ -96,6 +96,10 @@ class Bonding(Test):
                                                 default=False)
         self.peer_wait_time = self.params.get("peer_wait_time", default=5)
         self.sleep_time = int(self.params.get("sleep_time", default=5))
+        self.ib = 'N'
+        if self.host_interface[0:2] == 'ib':
+            self.ib = 'Y'
+        self.log.info("Test on IB Interface = %s", self.ib)
         self.peer_login(self.peer_first_ipinterface, self.user, self.password)
         self.setup_ip()
         self.err = []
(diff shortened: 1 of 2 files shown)


io/net/bonding.py
```diff
@@ -96,6 +96,10 @@ def setUp(self):
                                                 default=False)
         self.peer_wait_time = self.params.get("peer_wait_time", default=5)
         self.sleep_time = int(self.params.get("sleep_time", default=5))
+        self.ib = 'N'
+        if self.host_interface[0:2] == 'ib':
```
narasimhan-v: Assumption is infiniband interfaces start only with ib ?
Are there any other cases ?

io/net/bonding.py
```diff
@@ -96,6 +96,10 @@ def setUp(self):
                                                 default=False)
         self.peer_wait_time = self.params.get("peer_wait_time", default=5)
         self.sleep_time = int(self.params.get("sleep_time", default=5))
+        self.ib = 'N'
```
narasimhan-v: This can be False / True, so that the checking later can be
`if self.ib` instead of `if self.ib == 'Y'`

narasimhan-v:
@vaishnavibhat review please

io/net/bonding.py
manvanthar: Yeah Atleast hasn't changed in last 10 yrs.

io/net/bonding.py
manvanthar: ok can be done.

io/net/bonding.py
```diff
@@ -172,6 +176,17 @@ def setup_ip(self):
...
+    def bond_ib_conf(self, arg1, arg2):
+        '''
+        configure slaves for IB cards
+        '''
+        if arg2 == "ATTACH":
+            cmd = 'ifenslave %s %s;' % (self.bond_name, arg1)
+        else:
+            cmd = 'ifenslave %s -d %s;' % (self.bond_name, arg1)
+        if process.system(cmd, shell=True, ignore_status=True) != 0:
+            self.log.info("unable to %s IB interface %s", (arg2, arg1))
```
abdhaleegit: @manvanthar  should the test WARN or FAIL if unable to attach ?

manvanthar:
Fixed in PR #1378

PR #1335, opened 2019-10-09
Porting the module_unload_load test to python from shell

Naresh-ibm:
porting the module_unload_load test from shell to python.
Enhanced few functionality like for FC driver flushing the multipath
and then unload the module which was not there in shell script.


Naresh-ibm:
[module_load_unload_single_module_job_log_v4.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/3710692/module_load_unload_single_module_job_log_v4.txt)

io/driver/module_unload_load.py
```diff
@@ -0,0 +1,140 @@
...
+
+        for _ in range(0, self.iteration):
+            for mdl in module_list:
+                sub_mod = self.get_depend_modules(mdl)
+                if sub_mod:
+                    for mod in sub_mod.split(' '):
+                        if mod == 'multipath':
+                            if self.flush_mpath(mdl) is False:
+                                self.error_modules.append(mdl)
+                            continue
```
narasimhan-v: break or continue ?

io/driver/module_unload_load.py
```diff
@@ -0,0 +1,140 @@
...
+                        else:
+                            self.log.info("unloading sub module %s " % mod)
+                            process.system("rmmod %s" % mod)
```
narasimhan-v: why can't we use `linux_modules.unload_module()` here ?

io/driver/module_unload_load.py
```diff
@@ -0,0 +1,140 @@
...
+                self.log.info("unloading module %s " % mdl)
+                linux_modules.unload_module(mdl)
+                time.sleep(self.load_unload_sleep_time)
+                if linux_modules.module_is_loaded(mdl) is True:
+                    self.error_modules.append(mdl)
```
narasimhan-v: Should we break here ?

io/driver/module_unload_load.py
```diff
@@ -0,0 +1,140 @@
...
+                self.log.info("loading module : %s " % mdl)
+                linux_modules.load_module(mdl)
+                time.sleep(self.load_unload_sleep_time)
+                if linux_modules.module_is_loaded(mdl) is False:
+                    self.error_modules.append(mdl)
```
narasimhan-v: Should we break here ?

io/driver/module_unload_load.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def test(self):
+        """
+        Begining the test here
+        """
+        pci_addrs = []
+        if self.module:
+            for modl in self.module:
```
narasimhan-v: `for modl in self.module.split(',')`:

io/driver/module_unload_load.py
```diff
@@ -0,0 +1,140 @@
...
+                driver = process.getoutput(line.split('/')[-1])
+                self.mod_list.append(driver.split('.')[0])
+
+        for mod in self.mod_list:
+            if self.built_in_module(mod) is True:
+                self.mod_list.remove(mod)
+        self.module_load_unload(self.mod_list)
+
+        if self.error_modules:
+            self.fail("few modules failed to load unload, please check logs")
```
narasimhan-v: You can print those error modules here

io/driver/module_unload_load.py
Naresh-ibm: both will work, but continue will give us all failing sub-module list.

io/driver/module_unload_load.py
Naresh-ibm: sure, I missed it here. will change it.

io/driver/module_unload_load.py
narasimhan-v: why will `break` not give the failing sub-module list ?

io/driver/module_unload_load.py
narasimhan-v: Can you clarify why you are doing that please ?

io/driver/module_unload_load.py
Naresh-ibm: it is already splited (line 39)
and it is a list of modules. so not required here

io/driver/module_unload_load.py
Naresh-ibm: sure, I will add the print of error modules here.

io/driver/module_unload_load.py
Naresh-ibm: this is right way, please have a close look.
taking each module from a list of modules and unload_load the modules along with thier sub-modules for those many iteration.

then go for next module.

io/driver/module_unload_load.py
Naresh-ibm: if we give the break, it will come out after checking only one sub-modules.
if we give continue, it will check all the sub-modules and list all failing modules at the end. And I feel this is right way only.
the Aim of script also same i,e it will check for all modules or sub-modules and prints the error once.
And this will not happen for given number of iterations for sub module.
it will check for only one time for all sub-modules.
hope you understand.

io/driver/module_unload_load.py
narasimhan-v: Or we can `continue`

io/driver/module_unload_load.py
narasimhan-v: We have to decide on the preferred approach:
1. break after failing for one sub-module, there by taking less time overall.
(or)
2. exit loop only after iterating all sub-modules, thereby providing all sub-modules that are failing.

Naresh-ibm:
closing this PR in favour of #1343

PR #1345, opened 2019-10-15
Removing btrfs specific tests from py2-lts branch from LVM script -v3

manvanthar:
Removing btrfs specific tests from py2-lts branch from LVM script

### DIFF
--- a/io/disk/lvsetup.py
+++ b/io/disk/lvsetup.py
@@ -57,6 +57,14 @@ class Lvsetup(Test):
         if self.fs_name == 'xfs':
             pkg = 'xfsprogs'
         if self.fs_name == 'btrfs':
+            if distro.detect().name == "rhel":
+                if int(distro.detect().version) > 7:
+                    self.cancel("btrfs is not supported with \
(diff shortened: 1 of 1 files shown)


io/disk/lvsetup.py
```diff
@@ -57,6 +57,14 @@ def setUp(self):
         if self.fs_name == 'xfs':
             pkg = 'xfsprogs'
         if self.fs_name == 'btrfs':
+            if distro.detect().name == "rhel":
+                if int(distro.detect().version) > 7:
+                    self.cancel("btrfs is not supported with \
+                     RHEL 7.4 onwards")
+            if int(distro.detect().version) == 7 and \
```
narasimhan-v: @manvanthar The case should be:
```
if rhel
    if > 7
        cancel
    elif == 7 and >= 4
        cancel
```

PraveenPenguin: can it will be done as below 
```
if d>7 or (d==7 and v >=4):
     cancel

```

manvanthar:
PR fixed in #1348

PR #1371, opened 2019-10-22
deleting bootlist from net and added bootlist to common

smruti77:
bootlist for network and block device and separate yaml file
for both.


io/common/bootlist_test.py
```diff
@@ -18,38 +20,49 @@
...
-                          ignore_status=True):
-            self.cancel("Test not supported on the PowerNV Host platform")
+                                               default=None)
+        self.disk_names = self.params.get("disks", default=None)
+        if self.host_interfaces is not None:
+            if not self.host_interfaces:
+                self.cancel("user should specify host interfaces")
+            self.interfaces = ""
+            interfaces = netifaces.interfaces()
+            for host_interface in self.host_interfaces.split(","):
```
narasimhan-v: self.interfaces = self.host_interfaces.replace(',', ' ')

io/common/bootlist_test.py
```diff
@@ -18,38 +20,49 @@
...
+                if host_interface not in interfaces:
+                    self.cancel("interface is not available")
+                self.interfaces = '%s %s' % (self.interfaces, host_interface)
+        if self.disk_names is not None:
+            if not self.disk_names:
+                self.cancel("user should specify disk name")
+            self.disks = ""
+            for disk in self.disk_names.split(","):
```
narasimhan-v: self.disks = self.disk_names.replace(',', ' ')

io/common/bootlist_test.py.data/bootlist_test_disk.yaml
```diff
@@ -0,0 +1 @@
+disks: "None"
```
narasimhan-v: Is this right ?

io/common/bootlist_test.py.data/bootlist_test_network.yaml
```diff
@@ -0,0 +1 @@
+host_interfaces: "None"
```
narasimhan-v: Is this right ?

io/common/bootlist_test.py.data/bootlist_test_disk.yaml
smruti77: ya, i make default as None so i written here.

narasimhan-v: ok

io/common/bootlist_test.py
```diff
@@ -18,38 +20,46 @@
...
+                self.cancel("user should specify host interfaces")
+            self.interfaces = self.host_interfaces.replace(',', ' ')
+            interfaces = netifaces.interfaces()
+            for host_interface in self.host_interfaces.split(","):
+                if host_interface not in interfaces:
+                    self.cancel("interface is not available")
+        if self.disk_names is not None:
+            if not self.disk_names:
+                self.cancel("user should specify disk name")
+            self.disks = self.disk_names.replace(',', ' ')
```
narasimhan-v: Missed this in first review:

How about this:
```
if self.host_interfaces is not None:
    current code
elif self.disk_names is not None:
    current code
else:
    self.cancel("no interface or disk given")
```

io/common/bootlist_test.py
```diff
@@ -18,38 +20,46 @@
...
 
     def bootlist_mode(self, param):
         '''
         converting into different mode
         '''
-        cmd = "bootlist -m %s %s" % (param, self.interfaces)
+        if self.host_interfaces is not None:
```
abdhaleegit: is it necessary to check `self.host_interfaces` here, as you are already checking in the `setUp`

io/common/bootlist_test.py
```diff
@@ -18,38 +20,46 @@
...
+            cmd = "bootlist -m %s %s" % (param, self.interfaces)
+        elif self.disk_names is not None:
```
abdhaleegit: again checking for disk, as you already checking in setup

io/common/bootlist_test.py.data/bootlist_test_network.yaml
abdhaleegit: this means test will still work if host_interfaces and disks has default None  . if not may be we should have meaningful names instead of None

io/common/bootlist_test.py
smruti77: yes it necessary,because based on that we are running the command for all bootlist mode for both and passing self.interfaces and self.disks based on that.

io/common/bootlist_test.py
smruti77: no in case of None in disks and empty string also it pass the test.

io/common/bootlist_test.py.data/bootlist_test_network.yaml
smruti77: yes test is working,i didn't check that time.i need make this as empty string.

io/common/bootlist_test.py
narasimhan-v: @smruti77 That was why I was asking why "None" is given as default in yaml.
Can you please explain more here ?

PR #1382, opened 2019-11-04
Test for disable/enable namespace

harish-24:
Add test to verify disabling/enabling namespace of a region

### DIFF
--- a/memory/ndctl.py
+++ b/memory/ndctl.py
@@ -164,7 +164,7 @@ class NdctlTest(Test):
             self.fail("Failed to enable %s region(s)" % name)
 
     def disable_namespace(self, namespace='all', region='', bus='',
-                          force=False):
+                          verbose=False):
         """
         Disable namepsaces
         """
@@ -173,12 +173,12 @@ class NdctlTest(Test):
             args = '%s -r %s' % (args, region)
         if bus:
             args = '%s -b %s' % (args, bus)
-        if force:
-            args = '%s -f' % args
+        if verbose:
+            args = '%s -v' % args
 
         if process.system('%s disable-namespace %s' % (self.binary, args),
                           shell=True, ignore_status=True):
-            self.fail('Namespace disble command failed')
+            self.fail('Namespace disable failed for "%s"' % namespace)
 
     def enable_namespace(self, namespace='all', region='', bus='',
                          verbose=False):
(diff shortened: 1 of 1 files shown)


harish-24:
```
# avocado run memory/ndctl.py:test_disable_enable_ns
...
 (1/1) memory/ndctl.py:NdctlTest.test_disable_enable_ns: FAIL: Namespace enable failed for "all" (15.31 s)
RESULTS    : PASS 0 | ERROR 0 | FAIL 1 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 15.37 s
```

memory/ndctl.py
```diff
@@ -173,12 +173,29 @@ def disable_namespace(self, namespace='all', region='', bus='',
             args = '%s -r %s' % (args, region)
         if bus:
             args = '%s -b %s' % (args, bus)
-        if force:
-            args = '%s -f' % args
+        if verbose:
```
PraveenPenguin: any specfic reason to move force to verbose please have that part to commit mesage

memory/ndctl.py
```diff
@@ -322,6 +339,23 @@ def test_namespace(self):
         namespaces = self.get_json('-N')
         self.log.info('Created namespace %s', namespaces)
 
+    def test_disable_enable_ns(self):
+        """
+        Test enable disable namespace
+        """
+        region = self.get_default_region()
+        for _ in range(0, 3):
```
PraveenPenguin: why loop ?

memory/ndctl.py
harish-24: That is the correction we are doing here as it does not have a force option. Will have that as part of commit message

memory/ndctl.py
harish-24: to execute multiple times

PraveenPenguin: :) yes i understand that but the question is test requires to create three namespaces ?

memory/ndctl.py
PraveenPenguin: :+1:

memory/ndctl.py
harish-24: yes to enable/disable namespaces it has to be created first :)

harish-24:
@PraveenPenguin Pushed with commit message changes. Thanks.

PR #1388, opened 2019-11-08
Modifing the packet size to an generic format.

manvanthar:
Modifing the packet size to an generic format.


manvanthar:
Size currently fails if it is default MTU 1500.    Will be handing different MTU via scenarios.

PR #1389, opened 2019-11-08
ip_over_ib tests are obsolete for latest MOFED.

manvanthar:
ip_over_ib tests are obsolete for latest MOFED. For older MOFED versions the code will be retained in the python2 branch.


narasimhan-v:
@manvanthar There is no `Signed-off by` in the commit message, which causes a Travis Failure.

PR #1414, opened 2019-11-18
Enable ndctl selftest from source

harish-24:
Patch enables functional verification test for non-NFIT NVDIMM devices from ndctl source. This patch internally patches source with yet-to-be-upstream accepted proposal to run on non-NFIT devices. Once accepted the patching process would be removed from the test.

### DIFF
--- a/memory/ndctl_selftest.py
+++ b/memory/ndctl_selftest.py
@@ -1,0 +1,122 @@
+#!/usr/bin/env python
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2019 IBM
+# Author: Harish <harish@linux.vnet.ibm.com>
+#
+#
+
+import os
+import json
+
+from avocado import Test
+from avocado import main
+from avocado.utils import process, build, distro, git
+from avocado.utils.software_manager import SoftwareManager
+
+
+class NdctlTest(Test):
+    """
+    Ndctl is a user space tool to manage persistent memory devices
+    This test uses the selftests of the userspace tool to sanity check the
+    persistent memory
+    """
+
+    def get_bus_ids(self):
+        """
+        Get the provider IDs for each pmem device
+        """
+        bus_ids = []
+        os.chdir(self.sourcedir)
+        json_op = json.loads(process.system_output(
+            './ndctl/ndctl list -B', shell=True))
+        for nid in json_op:
+            for key, value in nid.items():
+                if key == 'provider':
+                    bus_ids.append(value)
+        return bus_ids
+
+    def setUp(self):
+        """
+        Prequisite for ndctl selftest on non-NFIT devices
+        """
+        smg = SoftwareManager()
+        self.url = self.params.get(
+            'url', default="https://github.com/pmem/ndctl.git")
+        self.branch = self.params.get('branch', default='master')
+        deps = ['gcc', 'make', 'automake', 'autoconf', 'patch']
+        detected_distro = distro.detect()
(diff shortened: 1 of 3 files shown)


harish-24:
```
linux-jtgh:/home/harish/misc-tests # avocado run memory/ndctl_selftest.py
...
 (1/1) memory/ndctl_selftest.py:NdctlTest.test: FAIL: ['FAIL: clear.sh', 'FAIL: daxdev-errors.sh'] (75.51 s)
RESULTS    : PASS 0 | ERROR 0 | FAIL 1 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
```

memory/ndctl_selftest.py
```diff
@@ -0,0 +1,121 @@
...
+                         'libuuid-devel-static', 'libjson-c-devel',
+                         'systemd-devel', 'kmod-bash-completion'])
+        else:
+            # TODO: Add RHEL when support arrives
+            self.cancel('Unsupported OS %s' % detected_distro.name)
+
+        for package in deps:
+            if not smg.check_installed(package) and not smg.install(package):
+                self.cancel(
+                    "Fail to install %s required for this test." % (package))
```
PraveenPenguin: can we start using python3 based string formate can be like 

`self.cancel("Fail to install {} required for this test." .format(package))`

memory/ndctl_selftest.py
```diff
@@ -0,0 +1,121 @@
...
+
+        process.run('./autogen.sh', sudo=True, shell=True)
+        process.run(
+            "./configure CFLAGS='-g -O2' --prefix=/usr "
+            "--sysconfdir=/etc --libdir=/usr/lib64 "
+            "--enable-destructive", shell=True, sudo=True)
+
+        # TODO: remove patches once merged upstream
+        process.run('patch -p1 < %s' %
+                    self.get_data('upstream.patch'), shell=True)
```
PraveenPenguin: can we get this path hosted somewhere as it sent upstream may be we use that url to get that

memory/ndctl_selftest.py
```diff
@@ -0,0 +1,121 @@
...
+            os.environ['BUS_PROVIDER1'] = bus_ids[1]
+
+    def test(self):
+        """
+        Selftests for non-NFIT devices
+        """
+        self.log.info("Running NDCTL selftests")
+        failed_tests = []
+        output = build.run_make(
+            self.sourcedir, extra_args='check -j 1', process_kwargs={"ignore_status": True})
```
PraveenPenguin: a suggestion just decode once so we need not  to convert every time

memory/ndctl_selftest.py
```diff
@@ -0,0 +1,121 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2019 IBM
+# Author: Harish <harish@linux.vnet.ibm.com>
+#
+#
+
```
PraveenPenguin: please remove this white space

memory/ndctl_selftest.py
harish-24: I think `.format` is supported from python2 time frame as well, not sure how it is related to python3. Anyhow if you are suggesting this as string format specifically it would be good to handle all other files with this change.

PraveenPenguin: yes it supported from python2 but python3 onwards it recomended to use

memory/ndctl_selftest.py
harish-24: makes sense. Will adopt that change for the first patch, Second patch is specific to PPC, so would like to keep it here. I will handle that as before applying as well. Thanks.

memory/ndctl_selftest.py
harish-24: It would be helpful if you could point me to the recommendation link so that I could send a patch addressing other tests as well. Thanks.

PraveenPenguin: you can refer this `https://www.python.org/dev/peps/pep-0498/`

PraveenPenguin: https://realpython.com/python-string-formatting/

harish-24:
@PraveenPenguin Pushed with all other changes. Thanks.

PR #1419, opened 2019-11-19
Correct memtester yaml file

harish-24:
Patch fixes a minor issue with yaml default inputs '1G' which is a string compared with an integer value. Hence making the input an integer, adding a NOTE explicitly and commenting other inputs
to be used on need basis.

### DIFF
--- a/memory/memtester.py.data/memtester.yaml
+++ b/memory/memtester.py.data/memtester.yaml
@@ -1,7 +1,8 @@
 test_memster:
     memtest_opt:
         runs: 1
-        memory: '1G'
+        # NOTE: Specify in MB
+        memory: 1024
         # NOTE: Provide page boundary address
-        physaddr: 0x100000
-        device: '/dev/mem'
+        physaddr:
+        device:


harish-24:
```
# avocado run memory/memtester.py -m memory/memtester.py.data/memtester.yaml 
...
...
 (1/1) memory/memtester.py:Memtester.test;memtest_opt-f75f: PASS (244.16 s)
JOB TIME   : 244.21 s
```

memory/memtester.py.data/memtester.yaml
```diff
@@ -1,7 +1,8 @@
 test_memster:
     memtest_opt:
         runs: 1
-        memory: '1G'
+        # NOTE: Specify in MB
+        memory: 1024
         # NOTE: Provide page boundary address
-        physaddr: 0x100000
-        device: '/dev/mem'
+        # physaddr: 0x100000
```
PraveenPenguin: may be left these as blank # denote that it is not supported

PraveenPenguin: and same need to handle in test script

harish-24: these are supported values, but depends on page boundary on each system, hence commented for reference.

PraveenPenguin: ok I understand but commenting express like  it is not supported that's why requested pass as blank or None which you feel should be better and handle in the script if it None or blank  skip that test

harish-24: OK

harish-24:
@PraveenPenguin Pushed, Thanks.

PraveenPenguin:
@harish-24 please back port same in py2-lts as well

PR #1438, opened 2019-11-25
modifying the multipath_test.py script for scenario testing

Naresh-ibm:
converted the code into functions, based on different scenarios,
So that each scenario can be ran when needed.
script also works fine as a whole.


narasimhan-v:
Merging with job.log from previous PR

PR #1444, opened 2019-11-25
dynamic SMT value and Hugepage memory tests

tfauck:
Default arch SMT is not necessary the on defined at boot time
So we need to dynamically set the SMT to its max with ppc64_cpu --smt=on and retrieve its values
Memory Hugepage tests must be skipped if not defined in kernel

Signed-off-by: Thierry  tfauck@free.fr


PraveenPenguin:
@tfauck please wrap all four commits in two commits as other two commit is not relevnt

PR #1446, opened 2019-11-25
Hugepage may not be defined in kernel

tfauck:
When Hugepage is not defined in kernel, memory.hugepage() is not defined -
Skipping test on that platform


PraveenPenguin:
@tfauck if possible can you please send this patch against py2-lts as well

PR #1447, opened 2019-11-25
SMT max setting

tfauck:
SMT max value may be configured at boot and may not match arch max
We can get dynamically the max value

	modified:   em_cpuhotplug.py
	modified:   ppc64_cpu_test.py


PraveenPenguin:
@tfauck please resolve pep8 issue as there is some indenation issue in script that help travis to green :)

sathnaga:
@tfauck pls squash the commits...

PR #1452, opened 2019-11-26
adding rawread test case

Naresh-ibm:
Adding the basic rawread test script.


Naresh-ibm:
[root@ltczz401-lp4 disk]# avocado run rawread.py -m rawread.py.data/rawread.yaml 
JOB ID     : cbd5ef8d81cc2f215ff91e7b3cf384002d378c01
JOB LOG    : /root/avocado-fvt-wrapper/results/job-2019-11-26T13.00-cbd5ef8/job.log
 (1/1) rawread.py:Rawread.test;run-82f4: PASS (49.41 s)
JOB TIME   : 49.51 s
JOB HTML   : /root/avocado-fvt-wrapper/results/job-2019-11-26T13.00-cbd5ef8/results.html
[root@ltczz401-lp4 disk]#

Naresh-ibm:
[job.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/3893773/job.log)

io/disk/rawread.py
```diff
@@ -0,0 +1,76 @@
...
+        else:
+            deps.extend(['gcc-c++'])
+
+        for package in deps:
+            if not smm.check_installed(package) and not smm.install(package):
+                self.cancel("Fail to install Package: %s" % package)
+
+        tarball_path = os.path.join(os.path.abspath(''),
+                                    "rawread.py.data/rawread.tar")
+        tarball = tarball_path.split("/")[-1]
```
narasimhan-v: `tarball = self.get_data('rawread.tar')`

io/disk/rawread.py.data/rawread.yaml
```diff
@@ -0,0 +1,10 @@
+# Bonnie options:
+# bonnie++ [-d scratch-dir] [-c concurrency] [-s size(MiB)[:chunk-size(b)]]
+#       [-n number-to-stat[:max-size[:min-size][:num-directories[:chunk-size]]]]
+#       [-m machine-name] [-r ram-size-in-MiB]
+#       [-x number-of-tests] [-u uid-to-use:gid-to-use] [-g gid-to-use]
+#       [-q] [-f] [-b] [-p processes | -y] [-z seed | -Z random-file]
+#       [-D]
```
narasimhan-v: Please correct it for rawread

io/disk/rawread.py.data/README
```diff
@@ -0,0 +1,8 @@
+This test attempts to read from the given disk. script test only the 
+basic functionality or disk capability using different values of 
+option t which is a sequential write() test and then reads the same.
+Currenly using locally available suit as we did not found the repo URL
+for latest rawread suit
```
narasimhan-v: This information is not needed here

io/disk/rawread.py
```diff
@@ -0,0 +1,76 @@
...
+        self.disk = self.params.get('disk', default=None)
+
+    def test(self):
+        """
+        Run 'rawread' with its arguments
+        """
+        for val in range(23):
+            cmd = "./rawread -t %s %s " % (val, self.disk)
+            if process.system(cmd, shell=True, ignore_status=True):
+                self.fail("test failed")
```
narasimhan-v: Will a fail in val=5 means it will fail for val=6 and higher values also ?

narasimhan-v:
@Naresh-ibm please attach job.log also, since it is a new test.

io/disk/rawread.py
Naresh-ibm: if the command fail, then it will fail there itself before going to next value.

io/disk/rawread.py.data/README
Naresh-ibm: ok

io/disk/rawread.py.data/rawread.yaml
Naresh-ibm: looks like I have attached the wrong file.
I have the edited file. will attach it. Thanks for it.

io/disk/rawread.py
Naresh-ibm: where we have self.get_data() method?

io/disk/rawread.py
narasimhan-v: @Naresh-ibm I understand that from the code. I was asking the behaviour of the test outside of automation

io/disk/rawread.py
narasimhan-v: @Naresh-ibm
```
self represents the instance of the class. By using the “self” keyword we can access the attributes and methods of the class in python. It binds the attributes with the given arguments.
```

io/disk/rawread.py
Naresh-ibm: NO, Command runs individually on given option at that instance. it does not depend on previous run result.

narasimhan-v: @Naresh-ibm Let me ask my question in a better way.
Is it useful to do something like below, which is `fail last` ?
```
for
    if process.system(cmd, shell=True, ignore_status=True):
        err.append(val)
if err:
    self.fail(val)
```

Naresh-ibm: can be doable. but we have many more option which we have to explore and add them.
So we can change that time. if you want I can add it right away too.

Naresh-ibm:
closing this in favor of #1458

PR #1454, opened 2019-11-27
Hugepage setting on py2 branch

tfauck:
Skip test if Hugepage not found in kernel


PraveenPenguin:
merging as it is already part of master branch

PR #1458, opened 2019-11-27
adding rawread test case

Naresh-ibm:
Adding the basic rawread test script.


Naresh-ibm:
[root@ltczz401-lp4 disk]# avocado run rawread.py -m rawread.py.data/rawread.yaml 
JOB ID     : 859fe0d5c49810382ec398ba3272a63224821930
JOB LOG    : /root/avocado-fvt-wrapper/results/job-2019-11-27T06.48-859fe0d/job.log
 (1/1) rawread.py:Rawread.test;run-82f4: PASS (69.39 s)
JOB TIME   : 69.59 s
JOB HTML   : /root/avocado-fvt-wrapper/results/job-2019-11-27T06.48-859fe0d/results.html
[root@ltczz401-lp4 disk]#

Naresh-ibm:
[job.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/3896949/job.log)

io/disk/rawread.py
```diff
@@ -0,0 +1,80 @@
...
+            deps.extend(['g++'])
+        else:
+            deps.extend(['gcc-c++'])
+
+        for package in deps:
+            if not smm.check_installed(package) and not smm.install(package):
+                self.cancel("Fail to install Package: %s" % package)
+
+        tarball_path = os.path.join(os.path.abspath(''),
+                                    "rawread.py.data/rawread.tar")
```
narasimhan-v: This is not needed. Please go through other examples.

io/disk/rawread.py
```diff
@@ -0,0 +1,80 @@
...
+        Run 'rawread' with its arguments
+        """
+        err_val = []
+        for val in range(23):
+            cmd = "./rawread -t %s %s " % (val, self.disk)
+            if process.system(cmd, shell=True, ignore_status=True):
+                err_val.append(val)
+
+        if err_val:
+            self.fail("test failed for values : %s" % err_val)
```
narasimhan-v: `" ".join(err_val)`

Naresh-ibm: why do we need this ?
err_val is a list

io/disk/rawread.py
Naresh-ibm: I have went through the couple of other test scripts, in those cases we have the url for the tool file, here we use the file from the data directory.
are you saying like below, 

locations = ["https://github.com/mkuoppal/tiobench/archive/master.zip"]
        tarball = self.fetch_asset("tiobench.zip", locations=locations)

I tried but it was erroring out, hence used the above thing.
can you suggest any specific file where can I check.

io/disk/rawread.py
narasimhan-v: You have the question and answer

io/disk/rawread.py
narasimhan-v: @Naresh-ibm 
You have looked at the wrong examples. You are confusing downloading file with local file.
Did you look at examples that use `self.get_data()` ?

Naresh-ibm: got it. will change.
Thanks for it.

io/disk/rawread.py
Naresh-ibm: val is integer here, you cannot join with a string.
if you convert val to string and join then the output what we see on fail is no better than without using join.

following are the different scenarios
-------------------------------------------------
1) with current code : (without any join)

===>> FAIL: test failed for values : [0, 1]
this is very clean.

2) without converting val to string and just use join as u given
===>>> ERROR: sequence item 0: expected str instance, int found

3. converting the val to string while we append to err_val itself
======>>>> FAIL: test failed for values : 0 1

4. converting list itself to string while printing
======>>> FAIL: test failed for values : [ 0 ,   1 ]
this doesn't make big difference.

my opinion: we are making it more complex by adding the space before each element which is by default we have one space in list element.
let me know urs.

narasimhan-v: @Naresh-ibm 

I never talked about converting `val` here. That is your assumption.
I am talking only about `err_val`.

What I want you to do is: when printing a list, format it into string and print, since you are using %s.

Naresh-ibm:
closing this in favor of #1464

PR #1466, opened 2019-11-28
changing the pci_class_name for fc adapter in PowerVMEEH.py

Naresh-ibm:
we use to get fc_host as pci_class_name for fc adapter but
the EEH command uses scsi_host as class name to trigger EEH.
hence changing accordingly.

### DIFF
--- a/io/pci/PowerVMEEH.py
+++ b/io/pci/PowerVMEEH.py
@@ -61,7 +61,8 @@ class PowerVMEEH(Test):
             self.cancel("Processor is not ppc64")
         if 'PowerNV' in genio.read_file("/proc/cpuinfo").strip():
             self.cancel("Test not supported on bare-metal")
-        if '0x1' not in genio.read_file("/sys/kernel/debug/powerpc/eeh_enable").strip():
+        eeh_enable_file = "/sys/kernel/debug/powerpc/eeh_enable"
+        if '0x1' not in genio.read_file(eeh_enable_file).strip():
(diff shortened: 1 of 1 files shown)


Naresh-ibm:
[job.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/3901261/job.log)

Naresh-ibm:
on triggering the EEH system crashed here on first EEH injection itself which was a actual bug.
hence could not got complete job.log and I don't have other setup currently.
the log has the changes that made here taking effect.

io/pci/PowerVMEEH.py
```diff
@@ -85,6 +86,8 @@ def test_eeh_basic_pe(self):
             self.pci_mem_addr = pci.get_memory_address(self.addr)
             self.pci_mask = pci.get_mask(self.addr)
             self.pci_class_name = pci.get_pci_class_name(self.addr)
+            if self.pci_class_name == 'fc_host':
```
narasimhan-v: @Naresh-ibm You need to change https://github.com/avocado-framework/avocado/blob/master/avocado/utils/pci.py#L149

Naresh-ibm: @narasimhan-v  I already went through this method. but this is specific only for EEH test case, i,e we need scsi_host. we may still need the fc_host as class name for other test. hence changed the test script.

let the library function remain same.

narasimhan-v: @Naresh-ibm okay, if you are fine with it.

PR #1470, opened 2019-11-28
Fix some lshw powerpc behaviour

tfauck:
Correct some lshw behaviour on power systems

### DIFF
--- a/ras/lshw.py
+++ b/ras/lshw.py
@@ -66,10 +66,13 @@ class Lshwrun(Test):
         if dist.name == "SuSE" and dist.version < 15:
             self.cancel("lshw not supported on SLES-%s. Please run "
                         "on SLES15 or higher versions only " % dist.version)
-        if (dist.name == 'Ubuntu' and dist.version.version >= 18) or dist.name == "SuSE":
+        if ((dist.name == 'Ubuntu' and dist.version.version >= 18) or
+                (dist.name == "SuSE")):
             packages.extend(['iproute2'])
         else:
             packages.extend(['iproute'])
+        if 'IBM' in process.system_output('lshw -class system', shell='true').decode():
+            packages.extend = ['powerpc-ibm-utils']
 
         for package in packages:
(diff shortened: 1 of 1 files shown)


PraveenPenguin:
@tfauck can you please resolve travis error as resolving pylint or pep8 complaint @Pavithra1602 review please

ras/lshw.py
```diff
@@ -123,9 +126,9 @@ def test_lshw_verification(self):
             self.fail("lshw failed to show correct mac address")
 
         # verify network
-        if self.active_interface\
-                not in self.run_cmd_out("lshw -class network"):
-            self.fail("lshw failed to show correct active network interface")
+        if 'ppc64le' not in distro.detect().arch:
+            if self.active_interface not in self.run_cmd_out("lshw -class network"):
```
PraveenPenguin: i think it is applicable to ppc64le arch as well

ras/lshw.py
```diff
@@ -207,8 +210,8 @@ def test_lshw_options(self):
             self.fail_cmd.append("lshw -enable network|grep -i interface")
 
         self.run_cmd("lshw -class -quiet")
-        if 'PowerVM'\
-                not in self.run_cmd_out("pseries_platform"):
+        if 'PowerNV'\
+                in self.run_cmd_out("pseries_platform"):
```
PraveenPenguin: this condition make this condition should not run in kvm so earlier condition seems me ok

ras/lshw.py
tfauck: agree

ras/lshw.py
tfauck: Got it - MY VMs were not started with -usb enabled meaning that HCI support wasn't there. So I will reverse the test, but I would also suggest to also check that usb is enabled otherwise test is irrelevant - what do you think ?

PraveenPenguin: @tfauck yes good idea

PR #1480, opened 2019-12-03
Enhance perf_24x7_hardware test errors

tfauck:
Add more relevant message test if perf not enabled at BMC level
grep -E returns error if ls returns null string

### DIFF
--- a/perf/perf_24x7_hardware_counters.py
+++ b/perf/perf_24x7_hardware_counters.py
@@ -35,7 +35,8 @@ class EliminateDomainSuffix(Test):
     :avocado: tags=perf,24x7,events
     """
 
-    @skipIf(IS_POWER_NV or IS_KVM_GUEST, "This test is not supported on KVM guest or Power non-virtualized platform")
+    @skipIf(IS_POWER_NV or IS_KVM_GUEST,
+            "This test is not supported on KVM guest or Power non-virtualized platform")
     def setUp(self):
         """
         Setup checks :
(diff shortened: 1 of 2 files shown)


perf/perf_24x7_hardware_counters.py
```diff
@@ -46,12 +47,14 @@ def setUp(self):
         """
         smm = SoftwareManager()
         detected_distro = distro.detect()
-        if 'ppc' not in process.system_output("uname -p", ignore_status=True).decode("utf-8"):
+        if 'ppc' not in process.system_output("uname -a", ignore_status=True).decode("utf-8"):
```
PraveenPenguin: the earlier condition also woks well so i don't think change is requrie

perf/perf_24x7_hardware_counters.py
```diff
@@ -168,10 +174,9 @@ def test_domain_chip_offset(self):
 
     # Helper functions
     def event_helper(self, event):
-        search_suffix = process.run('ls %s/events |grep -E '
-                                    '%s' % (self.event_sysfs, event),
+        search_suffix = process.run('ls %s/events' % (self.event_sysfs),
                                     ignore_status=True)
-        if search_suffix.stdout.decode("utf-8"):
+        if event in search_suffix.stdout.decode("utf-8"):
```
PraveenPenguin: nice catch

perf/perf_24x7_hardware_counters.py
tfauck: No uname -p is not portable so it doesn't work on all distro

PraveenPenguin: ah but i did not see any issue can you please list out which distro you are facing a issue

tfauck: -p sounds supported but may report value unknown - Got that on PowerNV debian system. Going to investigate why.

tfauck: reverting

perf/perf_24x7_hardware_counters.py
tfauck: thx

tfauck:
Did wrap both commits in a single commit for perf/perf_24x7...

PR #1484, opened 2019-12-03
   Fix some lshw powerpc behaviour

tfauck:
lshw --numeric |grep HCI make sense only if USB enabled
    lshw --class network doesn't get the ip name
    add check on ppc64el and ppc64le

### DIFF
--- a/ras/lshw.py
+++ b/ras/lshw.py
@@ -63,13 +63,16 @@ class Lshwrun(Test):
         self.is_fail = 0
         dist = distro.detect()
         packages = ['lshw', 'net-tools', 'pciutils']
-        if dist.name == "SuSE" and dist.version < 15:
+        if dist.name == "SuSE" and int(dist.version) < 15:
             self.cancel("lshw not supported on SLES-%s. Please run "
                         "on SLES15 or higher versions only " % dist.version)
-        if (dist.name == 'Ubuntu' and dist.version.version >= 18) or dist.name == "SuSE":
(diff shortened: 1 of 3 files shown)


PraveenPenguin:
@tfauck seems you still need to address review comment from #1470

ras/lshw.py
```diff
@@ -123,9 +126,9 @@ def test_lshw_verification(self):
             self.fail("lshw failed to show correct mac address")
 
         # verify network
-        if self.active_interface\
-                not in self.run_cmd_out("lshw -class network"):
-            self.fail("lshw failed to show correct active network interface")
+        if distro.detect().arch not in ['ppc64le', 'ppc64el']:
```
PraveenPenguin: this make test not run ppc64le system so not require so earlier test code is fine

ras/lshw.py
```diff
@@ -209,8 +212,8 @@ def test_lshw_options(self):
         self.run_cmd("lshw -class -quiet")
         if 'PowerVM'\
                 not in self.run_cmd_out("pseries_platform"):
-            if not self.run_cmd_out("lshw"
-                                    " -numeric | grep HCI | cut -d':' -f3"):
+            lshw_out = self.run_cmd_out("lshw -numeric")
+            if 'usb' in lshw_out and 'HCI' not in lshw_out:
```
PraveenPenguin: :+1:

PR #1516, opened 2020-01-13
multipath_test.py fails with  bytes-like object is required, not 'str'

tfauck:
if self.wwids is empty call to get_multipath_wwids() fails. Check for
empty string before that call.

### DIFF
--- a/io/disk/multipath_test.py
+++ b/io/disk/multipath_test.py
@@ -72,6 +72,8 @@ class MultipathTest(Test):
         # Check if given multipath devices are present in system
         self.wwids = self.params.get('wwids', default='').split(',')
         system_wwids = multipath.get_multipath_wwids()
+        if self.wwids == ['']:
+            self.cancel("No System Multipath Devices available")
         wwids_to_remove = []
         for wwid in self.wwids:
             if wwid not in system_wwids:
--- a/io/disk/multipath_test.py
+++ b/io/disk/multipath_test.py
@@ -72,7 +72,7 @@ class MultipathTest(Test):
         # Check if given multipath devices are present in system
         self.wwids = self.params.get('wwids', default='').split(',')
         system_wwids = multipath.get_multipath_wwids()
-        if self.wwids == ['']:
+        if system_wwids == ['']:
             self.cancel("No System Multipath Devices available")
         wwids_to_remove = []
         for wwid in self.wwids:


narasimhan-v:
@tfauck call to `get_multipath_wwids()` might still fail in this case also, there are no multipath devices. The right place to fix would be https://github.com/avocado-framework/avocado/blob/master/avocado/utils/multipath.py#L95

tfauck:
I agree that adding a decode() to avocado/utils/multipath.py would handle the case of no multipath device returning a null string. Do you mean that we can have definitions for wwids but no system device so we would call get_multipath_wwids() and that will fail?

tfauck:
but changing avocado/utils/multipath.py would result in cancelling if /etc/multipath/wwids doesn't exist. Which is something we can do in the test also. If we don't want to cancel if file doesn't exist, all commands will fail and we have to hange all process.out behaviour which is not what we want. Waht do you think ?

narasimhan-v:
@tfauck I did not mean that. What you are doing is:  cancelling the test when there are no input values for wwids. If you want get_multipath_wwids() to not return Null, you need to handle it there.

And you can additionally handle it in the test as well, but that would be based on if wwids exist in the system or not. So, even for that case, I would suggest getting get_multipath_wwids() to do that.

tfauck:
Found that my avocado environment was not up to date - the byte error is now fixed in the code. So just updating test code to send a cancel when no system wwids available.

io/disk/multipath_test.py
```diff
@@ -72,6 +72,8 @@ def setUp(self):
         # Check if given multipath devices are present in system
         self.wwids = self.params.get('wwids', default='').split(',')
         system_wwids = multipath.get_multipath_wwids()
+        if self.wwids == ['']:
```
narasimhan-v: @tfauck This is already done in line 84

tfauck: yep - I wanted to say if system_wwids == ... fixing that - thanks

narasimhan-v:
@Naresh-ibm Review please

Naresh-ibm:
@tfauck @narasimhan-v    I feel existing script should be fine and no need to change as the existing script takes care of it but little later. this will help in other scenarios too like below,

if the wwids are provided and say they are not existing in system then also we need to cancel.
so I feel existing scripts handles both the cases very well.

tfauck:
I will check again tomorrow but if if we don't change anything  you don't get a cancel but a 'ERROR: list index out of range' message in File "/usr/local/lib/python3.7/dist-packages/avocado/utils/multipath.py", line 90, in get_mpath_name because multipath -l returns ['']

Naresh-ibm:
@tfauck @narasimhan-v we can handle via library function and as well as the current patch.
@narasimhan-v as we discussed yesterday, merging this also ok for me.
hence approving this.

io/disk/multipath_test.py
```diff
@@ -72,6 +72,8 @@ def setUp(self):
         # Check if given multipath devices are present in system
         self.wwids = self.params.get('wwids', default='').split(',')
         system_wwids = multipath.get_multipath_wwids()
+        if system_wwids == ['']:
```
narasimhan-v: if system_wwids == []:

PR #1538, opened 2020-02-10
Add few python fixes

tfauck:
More debian, ubuntu , centos support
Add -lm at end of compilation line for -lm to be found


tfauck:
I experimented the issue mentionned in https://stackoverflow.com/questions/10059146/gcc-libm-not-working , and moving -lm at end of gcc command it just worked. However I can't explain why in some case like this one -lm is not taken properly.

tfauck:
ALready integrated in code

PR #1555, opened 2020-02-17
Skip test if run on KVM

tfauck:
Command fails on KVM

### DIFF
--- a/ras/diag_encl.py
+++ b/ras/diag_encl.py
@@ -17,11 +17,13 @@
 import os
 import glob
 import xml.etree.ElementTree
-from avocado import Test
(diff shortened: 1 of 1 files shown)


ras/diag_encl.py
```diff
@@ -43,6 +45,7 @@ def run_cmd_out(cmd):
                                      ignore_status=True,
                                      sudo=True).decode("utf-8")
 
+    @skipIf(not IS_POWER_NV, "Skipped on KVM system")
```
PraveenPenguin: may be this string needs to re worded as "Skipped Test case test supported in PowerNV system"

PR #1563, opened 2020-02-19
Added ServiceReport as a Test case

PraveenPenguin:
Added ServiceReport as a Test case , Later will enhance this test case
having more validation

### DIFF
--- a/ras/ServiceReport.py
+++ b/ras/ServiceReport.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2020 IBM.
+# Author: Praveen K Pandey <praveen@linux.vnet.ibm.com>
+
+import os
+
+from avocado import Test
+from avocado import main
+from avocado.utils import process, build
+from avocado.utils import archive
+from avocado.utils.software_manager import SoftwareManager
+
+
+class ServiceReport(Test):
+
+    """
+    ServiceReport is a tool to validate and repair system configuration for
+    specific purposes.Initially envisaged to help setup systems for correct
+    First Failure Data Capture (FFDC),it has now morphed into a plugin based
+    framework which can do more than just FFDC validation.
+
+    :avocado: tags=privileged
+    """
+
+    def setUp(self):
+        smm = SoftwareManager()
+        self.options = self.params.get('option', default='-l')
+        if self.options == "-p":
+            self.plugin = self.params.get('plugin_val', default='kdump')
+            self.options = "%s %s" % (self.options, self.plugin)
+        for package in ['make', 'gcc']:
+            if not smm.check_installed(package) and not smm.install(package):
+                self.cancel("Fail to install %s required for this"
+                            " test." % package)
+        tarball = self.fetch_asset('ServiceReport.zip', locations=[
+                                   'https://github.com/linux-ras/ServiceReport'
+                                   '/archive/master.zip'], expire='7d')
+        archive.extract(tarball, self.workdir)
+        self.sourcedir = os.path.join(self.workdir, 'ServiceReport-master')
+        build.make(self.sourcedir)
+
+    def test(self):
+        os.chdir(self.sourcedir)
+        cmd = "./servicereport %s" % self.options
+        if process.system(cmd, ignore_status=True, sudo=True, shell=True):
+            self.fail("ServiceReport: Failed command is: %s" % cmd)
+
+
+if __name__ == "__main__":
+    main()
--- a/ras/ServiceReport.py.data/option.yaml
+++ b/ras/ServiceReport.py.data/option.yaml
@@ -1,0 +1,15 @@
+setup:
+ opt: !mux
+ basic:
+         option: ''
+ list:
+         option: '-l'
+ recovery:
+         option: '-r'
+ plugin_kdump:
+         option: '-p'
+         ##this only can be used with -p option
+         plugin_val: 'kdump'
+  plugin_htx:
+         option: '-p'
+         plugin_val: 'htx'


ras/ServiceReport.py
```diff
@@ -0,0 +1,75 @@
...
+
+    fail_cmd = list()
+
+    def run_cmd(self, cmd):
+        if process.system(cmd, ignore_status=True, sudo=True, shell=True):
+            self.is_fail += 1
+            self.fail_cmd.append(cmd)
+
+    def error_check(self):
+        if len(self.fail_cmd) > 0:
```
harish-24: `> 0` is unnecessary.

ras/ServiceReport.py
```diff
@@ -0,0 +1,75 @@
...
+            for cmd in range(len(self.fail_cmd)):
+                self.log.info("Failed command: %s" % self.fail_cmd[cmd])
+            self.fail("lshw: Failed commands are: %s" % self.fail_cmd)
```
harish-24: `lshw` typo?

ras/ServiceReport.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def setUp(self):
+        smm = SoftwareManager()
+        self.is_fail = 0
```
harish-24: `is_fail` being a flag is not used elsewhere.

ras/ServiceReport.py
```diff
@@ -0,0 +1,75 @@
...
+        tarball = self.fetch_asset('ServiceReport.zip', locations=[
+                                   'https://github.com/linux-ras/ServiceReport'
+                                   '/archive/master.zip'], expire='7d')
+        archive.extract(tarball, self.workdir)
+        self.sourcedir = os.path.join(self.workdir, 'ServiceReport-master')
+        build.make(self.sourcedir, ignore_status=True)
+
+    def test(self):
+        os.chdir(self.sourcedir)
+        self.run_cmd("./servicereport")
```
harish-24: Suggestion to loop through list of commands to avoid repeating.

ras/ServiceReport.py
```diff
@@ -0,0 +1,75 @@
...
+        for package in ['make', 'gcc']:
+            if not smm.check_installed(package) and not smm.install(package):
+                self.cancel("Fail to install %s required for this"
+                            " test." % package)
+        tarball = self.fetch_asset('ServiceReport.zip', locations=[
+                                   'https://github.com/linux-ras/ServiceReport'
+                                   '/archive/master.zip'], expire='7d')
+        archive.extract(tarball, self.workdir)
+        self.sourcedir = os.path.join(self.workdir, 'ServiceReport-master')
+        build.make(self.sourcedir, ignore_status=True)
```
harish-24: If build fails, the test should not continue right?

ras/ServiceReport.py
```diff
@@ -0,0 +1,75 @@
...
+    fail_cmd = list()
+
+    def run_cmd(self, cmd):
+        if process.system(cmd, ignore_status=True, sudo=True, shell=True):
+            self.is_fail += 1
+            self.fail_cmd.append(cmd)
+
+    def error_check(self):
+        if len(self.fail_cmd) > 0:
+            for cmd in range(len(self.fail_cmd)):
```
harish-24: Isn't simple loop better
```
for cmd in self.fail_cmd:
    self.log.info("Failed command: %s" % cmd)
```

ras/ServiceReport.py
PraveenPenguin: ah yes, will remove this

ras/ServiceReport.py
PraveenPenguin: yes typo will take care on next version

ras/ServiceReport.py
PraveenPenguin: sure

PraveenPenguin:
@harish-24 pushed suggested change, Please have a look

ras/ServiceReport.py
```diff
@@ -0,0 +1,71 @@
...
+        tarball = self.fetch_asset('ServiceReport.zip', locations=[
+                                   'https://github.com/linux-ras/ServiceReport'
+                                   '/archive/master.zip'], expire='7d')
+        archive.extract(tarball, self.workdir)
+        self.sourcedir = os.path.join(self.workdir, 'ServiceReport-master')
+        build.make(self.sourcedir)
+
+    def test(self):
+        os.chdir(self.sourcedir)
+        cmd_list = ['', '-l', '-v', '-r', '-p %s' % self.plugin]
```
narasimhan-v: Why can't this be multiplexed ?

ras/ServiceReport.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    fail_cmd = list()
+
+    def run_cmd(self, cmd):
+        if process.system(cmd, ignore_status=True, sudo=True, shell=True):
+            self.fail_cmd.append(cmd)
+
+    def error_check(self):
+        if len(self.fail_cmd):
```
narasimhan-v: `if self.fail_cmd`

ras/ServiceReport.py.data/option.yaml
```diff
@@ -0,0 +1 @@
+plugin: 'kdump,htx'
```
narasimhan-v: What is the idea behind checking 2 plugins together ? Wouldn't it be more useful if we check one at a time ?

PraveenPenguin: @narasimhan-v it does in a sequential manner as tools give us an option we can pass multiple plugins in time, so seems me should be ok as it user wish if he wants to pass one on multiple

ras/ServiceReport.py
PraveenPenguin: @narasimhan-v no don't think these are basic option need to be checked, anyway, this is basic commit will modify about our need and have more error handling

ras/ServiceReport.py
PraveenPenguin: ok make sense

ras/ServiceReport.py
narasimhan-v: @PraveenPenguin What I meant was, why use for loop when avocado does it for us. We can even get per parameter result.

ras/ServiceReport.py.data/option.yaml
narasimhan-v: @PraveenPenguin Yes, just that as default option, we can provide one option at a time. Since default yaml serves as guideline / template for users.

PraveenPenguin: ok sure

ras/ServiceReport.py
PraveenPenguin: @narasimhan-v ok but still I  feel should be part of the test itself as error processing should be different as based on return output that will be targeted on enhancement as right now it is same for all the cases and don't want to run setup every time so coded this way i know we can skip setUP as well but I  need to modify logic based on enhancement how this sound you?

narasimhan-v: @PraveenPenguin Sure, if you have planned enhancements then this might be fine.
Suggestion: please explore if error handling also can be provided as input via yaml.

PraveenPenguin:
@narasimhan-v pushed the changes as this works as a Test case like our use case as disscued with target wrapper

ras/ServiceReport.py.data/option.yaml
```diff
@@ -0,0 +1,15 @@
+setup:
+ opt: !mux
+ basic:
+         options: ''
+ list:
+         option: '-l'
+ recovery:
+         option: '-r'
+ plugin1:
```
narasimhan-v: plugin-kdump

ras/ServiceReport.py.data/option.yaml
```diff
@@ -0,0 +1,15 @@
...
+         option: '-p'
+         ##this only can be used with -p option
+         plugin_val: 'kdump'
+  plugin1:
```
narasimhan-v: plugin-htx

PraveenPenguin: @narasimhan-v done, please have a look

ras/ServiceReport.py.data/option.yaml
```diff
@@ -0,0 +1,15 @@
+setup:
+ opt: !mux
+ basic:
+         options: ''
```
harish-24: typo here, otherwise looks good.

PraveenPenguin: ah yes, pushed

PraveenPenguin:
@narasimhan-v @harish-24 thanks for ACK , merging

PR #1585, opened 2020-03-16
Nvdimm test update

kvaneesh:
support for using a different branch with ndctl upstream git repo.


kvaneesh:
updated also added a new patch to avoid passing namespace size. @harish-24, please review

PraveenPenguin:
@harish-24 thanks For ACK

PR #1610, opened 2020-04-08
changed input values to a common format

abdhaleegit:
As decided to follow space separated input values passed to any test
made the changes and updated readme to follow the same


PraveenPenguin:
@abdhaleegit why this patch needed as it just introducing another way to pass data @narasimhan-v any thought on this

narasimhan-v:
> @abdhaleegit why this patch needed as it just introducing another way to pass data @narasimhan-v any thought on this

@PraveenPenguin This is needed to unify the input formats across tests

PR #1613, opened 2020-04-09
New test vscsi dlpar

smruti77:
new test for dlpar operation of vscsi device.

### DIFF
--- a/io/disk/dlpar_vscsi.py
+++ b/io/disk/dlpar_vscsi.py
@@ -1,0 +1,173 @@
+#!/usr/bin/python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Author: Bismruti Bidhibrata Pattjoshi<bbidhibr@in.ibm.com>
+
+"""
+DLPAR operations
+"""
+
+try:
+    import pxssh
+except ImportError:
+    from pexpect import pxssh
+from avocado import Test
+from avocado import main
+from avocado.utils import process
+
+
+class CommandFailed(Exception):
+    '''
+    Defines the exception called when a
+    command fails
+    '''
+    def __init__(self, command, output, exitcode):
+        Exception.__init__(self, command, output, exitcode)
+        self.command = command
+        self.output = output
+        self.exitcode = exitcode
+
+    def __str__(self):
+        return "Command '%s' exited with %d.\nOutput:\n%s" \
+               % (self.command, self.exitcode, self.output)
+
+
+class DlparTest(Test):
+    '''
+    DLPAR disk script does vscsi device add,remove.
+    Update the details in yaml file.
+    '''
+
+    def setUp(self):
+        '''
+        Gather necessary test inputs.
+        '''
+        self.disk = self.params.get('disk', default=None)
+        self.num_of_dlpar = int(self.params.get("num_of_dlpar", default='1'))
+        self.vios_ip = self.params.get('vios_ip', '*', default=None)
+        self.vios_user = self.params.get('vios_username', '*', default=None)
+        self.vios_pwd = self.params.get('vios_pwd', '*', default=None)
+        self.login(self.vios_ip, self.vios_user, self.vios_pwd)
+        cmd = "lscfg -l %s" % self.disk
+        for line in process.system_output(cmd, shell=True).decode("utf-8") \
+                                                          .splitlines():
+            if self.disk in line:
+                self.slot = line.split()[-1].split('-')[-2]
+        cmd = "lsdev -slots"
+        output = self.run_command(cmd)
+        for line in output.splitlines():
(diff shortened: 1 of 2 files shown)


smruti77:
[vscsi_dlpar.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/4454986/vscsi_dlpar.log)

io/disk/dlpar_vscsi.py
```diff
@@ -0,0 +1,155 @@
...
+            self.log.info(output)
+        except CommandFailed as cf:
+            self.fail("failed dlpar operation:remove disk, %s" % cf)
+        cmd = "mkvdev -vdev %s -vadapter %s -dev %s" % (self.hdisk_name,
+                                                        self.host, self.vscsi)
+        try:
+            output = self.run_command(cmd)
+            self.log.info(output)
+        except CommandFailed as cf:
+            self.fail("Failed dllpar operation:add disk, %s" % cf)
```
narasimhan-v, line 151: Better to have functions to remove and add device, and call them.
Also, can have number of dlpar operations controlled by yaml parameter

io/disk/dlpar_vscsi.py
```diff
@@ -0,0 +1,155 @@
...
+        for line in output.decode("utf-8").splitlines():
+            if self.disk in line:
+                value = line.split()[0]
+        for line in output.decode("utf-8").splitlines():
+            if value in line:
+                if "dir" in line:
+                    self.disk_dir = line.split()[-1]
+        cmd = r"cat %s/inquiry" % self.disk_dir
+        output = process.system_output(cmd, shell=True)
+        self.hdisk_name = output.split()[2].strip(b'0001').decode("utf-8")
```
narasimhan-v, line 85: It can be decoded to utf-8 and then 0001 can be stripped

io/disk/dlpar_vscsi.py
```diff
@@ -0,0 +1,155 @@
...
+        cmd = "lsscsi -vl | tr -d '[]'"
+        output = process.system_output(cmd, shell=True)
+        for line in output.decode("utf-8").splitlines():
+            if self.disk in line:
+                value = line.split()[0]
+        for line in output.decode("utf-8").splitlines():
+            if value in line:
+                if "dir" in line:
+                    self.disk_dir = line.split()[-1]
+        cmd = r"cat %s/inquiry" % self.disk_dir
```
narasimhan-v, line 83: can't we use `genio.read_file()` ?

io/disk/dlpar_vscsi.py
```diff
@@ -0,0 +1,155 @@
...
+            if self.disk in line:
+                self.slot = line.split()[-1].split('-')[-2]
+        cmd = "lsdev -slots"
+        output = self.run_command(cmd)
+        print(output)
+        for line in output.splitlines():
+            if self.slot in line:
+                self.host = line.split()[-1]
+        self.log.info(self.host)
+        cmd = "lsscsi -vl | tr -d '[]'"
```
narasimhan-v, line 74: please use native python parsing instead of bash.

smruti77:
> @smruti77 What are the differences between vFC and vscsi dlpar ?
> Based on that, we can decide if we need separate tests.

this is for scsi disk.

io/disk/dlpar_vscsi.py
smruti77: ok

io/disk/dlpar_vscsi.py
smruti77: if i will put decode("utf-8") "output = process.system_output(cmd, shell=True)" here getting error, so put after that.

io/disk/dlpar_vscsi.py
smruti77: ok,will use that.

io/disk/dlpar_vscsi.py
smruti77: ok

io/disk/dlpar_vscsi.py
narasimhan-v: oh ok

narasimhan-v:
> > @smruti77 What are the differences between vFC and vscsi dlpar ?
> > Based on that, we can decide if we need separate tests.
> 
> this is for scsi disk.

Is there difference in how they are handled in this context ?

smruti77:
> > > @smruti77 What are the differences between vFC and vscsi dlpar ?
> > > Based on that, we can decide if we need separate tests.
> > 
> > 
> > this is for scsi disk.
> 
> Is there difference in how they are handled in this context ?

i don't know more about vfc.we also thought to kept vfc, vscsi, veth in one file.letter we decided to have separate file, vfc was already their and i sent vscsi so.

io/disk/dlpar_vscsi.py
```diff
@@ -0,0 +1,171 @@
...
+        cmd = r"cat %s/inquiry" % self.disk_dir
+        output = process.system_output(cmd, shell=True)
+        self.hdisk_name = output.split()[2].strip(b'0001').decode("utf-8")
+        self.log.info(self.hdisk_name)
+        cmd = "lsmap -all|grep -p %s" % self.hdisk_name
+        output = self.run_command(cmd)
+        for line in output.splitlines():
+            if "VTD" in line:
+                self.vscsi = line.split()[-1]
+        self.log.info(self.vscsi)
```
narasimhan-v, line 91: @smruti77 Can we cancel the test in case self.vscsi is not got for some reason ?

smruti77: ok

PR #1615, opened 2020-04-10
Add dependencies

tfauck:
Add more dependencies for Debian and Ubuntu


toolchain/oprofile.py
```diff
@@ -31,13 +31,14 @@ def setUp(self):
         sm = SoftwareManager()
         detected_distro = distro.detect()
         deps = ['dejagnu', 'expect']
-        if detected_distro.name == "SuSE":
+        if detected_distro.name in ["SuSE", 'fedora']:
```
PraveenPenguin, line 34: I feel fedora we can install git it will instal `git-core` as well so may be this fix not needed

PR #1617, opened 2020-04-13
Fixing the way lpar name is found

narasimhan-v:
Earlier, we were finding LPAR name by using lsrsrc to find LPAR
host name and then using that to find LPAR name from HMC. But
we came up with lsslot later which was a better way. So, removing
some leftover workarounds from the older method using this commit.


PraveenPenguin:
@abdhaleegit review please

PR #1623, opened 2020-04-17
Added timeout value to runcommand function call in function changehwres

harihare:
On large systems dlpar tests are failing due to timeout, as changehwres
takes more time to release the hardware. Hence added the timeout to
3000 secs. So that tests will pass even on large systems.

Signed-off-by: Hariharan T S <hari@linux.vnet.ibm.com>

### DIFF
--- a/io/pci/dlpar.py
+++ b/io/pci/dlpar.py
@@ -351,7 +351,7 @@ class DlparPci(Test):
                    -o %s --id %s -l %s ' % (server, operation, lpar_id,
                                             drc_index)
         try:
-            cmd = self.run_command(cmd)
+            cmd = self.run_command(cmd, 3000)
         except CommandFailed as cmd_fail:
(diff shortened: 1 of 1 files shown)


io/pci/dlpar.py
```diff
@@ -351,7 +351,7 @@ def changehwres(self, server, operation, lpar_id, lpar, drc_index, msg):
                    -o %s --id %s -l %s ' % (server, operation, lpar_id,
                                             drc_index)
         try:
-            cmd = self.run_command(cmd)
+            cmd = self.run_command(cmd, 3000)
```
narasimhan-v, line 354: @harihare Can this timeout be added to `listhwres()` as well ?

harihare: @narasimhan-v I dont think it is requried there. Bydefault time our for all run_command is 300  seconds. The same worked fine for the large system too.

PR #1625, opened 2020-04-20
Fixed local variable issue in port_bounce script

harihare:
portbounce test was failing due to local variable
referenced out the for loop. Fixed the same moving
the variable out of the loop.

Signed-off-by: Hariharan T S <hari@linux.vnet.ibm.com>


narasimhan-v:
@Naresh-ibm Ack please

Naresh-ibm:
@harihare Thanks for this fix.

PR #1636, opened 2020-04-22
module parameter generic yaml for dirvers

abdhaleegit:
Created a new yaml which is generic for driver param testing

### DIFF
--- a/io/driver/driver_parameter.py.data/driver_parameter.yaml
+++ b/io/driver/driver_parameter.py.data/driver_parameter.yaml
@@ -1,0 +1,12 @@
+module:
+interface:
+host_ip:
+netmask:
+peer_ip:
+sysfs_check_required:
+module_param_name:
+value: !mux
+    value1:
+        module_param_value:
+    value2:
+        module_param_value:


narasimhan-v:
@abdhaleegit Can you explain the use cases a bit more please ?

abdhaleegit:
> @abdhaleegit Can you explain the use cases a bit more please ?
As we cannot use the yaml for different adapter, so wanted to have a template to create yaml for different adapter.. but looks like we need yaml for each driver separately so here is the new one https://github.com/avocado-framework-tests/avocado-misc-tests/issues/1698

abdhaleegit:
can we close this one

abdhaleegit:
@narasimhan-v @vaishnavibhat @smruti77 cant we have single yaml for all drivers

because it will be diffcult for me to handle in cfg bucket... I need to have separate bucket for each driver

narasimhan-v:
@abdhaleegit this is a valid point, we can discuss and decide this locally. Please open an internal issue.

abdhaleegit:
https://github.com/avocado-framework-tests/avocado-misc-tests/issues/1716

PR #1638, opened 2020-04-22
fixed distro_tools for other interface type.

smruti77:
fixed distro_tools for different interfaces.

### DIFF
--- a/io/pci/distro_tools.py
+++ b/io/pci/distro_tools.py
@@ -40,6 +40,7 @@ class DisrtoTool(Test):
         self.option = self.params.get("test_opt", default='')
         self.tool = self.params.get("tool", default='')
         self.pci_device = self.params.get("pci_device", default='')
+        self.adapter_type = pci.get_pci_class_name(self.pci_device)
         smm = SoftwareManager()
         if not smm.check_installed("pciutils") and not smm.install("pciutils"):
             self.cancel("pciutils package is need to test")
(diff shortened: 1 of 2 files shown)


io/pci/distro_tools.py
```diff
@@ -40,6 +40,7 @@ def setUp(self):
         self.option = self.params.get("test_opt", default='')
         self.tool = self.params.get("tool", default='')
         self.pci_device = self.params.get("pci_device", default='')
+        self.adapter_type = self.params.get("adapter_type", default='')
```
narasimhan-v, line 43: Why do we need this ?
Can't this be found via `pci.get_pci_class_name()` https://github.com/avocado-framework/avocado/blob/781adbaa7a0736f10114fb8f5bd107f08f495eb5/avocado/utils/pci.py#L138

io/pci/distro_tools.py.data/distro_tools.yaml
```diff
@@ -1,4 +1,5 @@
 pci_device: ""
+adapter_type: "net"
```
narasimhan-v, line 2: Why only net here ?

smruti77: just for default i gave.

io/pci/distro_tools.py
smruti77: ok i will use this.

io/pci/distro_tools.py.data/distro_tools.yaml
narasimhan-v: This is not needed if you use get_class_name()

Issue #1682, opened 2020-05-04
bridge.py: br0 interface did not clean up after the test


abdhaleegit:
br0 bridge interface is not removed at the end of test resulting 

making the interface config messed up, results in next test failures

[root@ltc-zz1b-lp2 network-scripts]# ls
ifcfg-br0 

ip a
8: br0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc noqueue state DOWN group default qlen 1000
    link/ether 42:c5:f0:75:a1:4c brd ff:ff:ff:ff:ff:ff
    inet 102.10.10.188/24 brd 102.10.10.255 scope global noprefixroute br0
       valid_lft forever preferred_lft forever
3: enP19p96s0f0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 98:be:94:07:3a:68 brd ff:ff:ff:ff:ff:ff
    inet 102.10.10.188/24 brd 102.10.10.255 scope global noprefixroute enP19p96s0f0
       valid_lft forever preferred_lft forever
    inet6 fe80::9abe:94ff:fe07:3a68/64 scope link noprefixroute 
       valid_lft forever preferred_lft forever

smruti77:
@abdhaleegit i am not getting this issue.

abdhaleegit:
after deleting br0,,  ifup is bringing up the bridge interface

abdhaleegit:
please delete br0 script in  /etc/sysconfig/network-scripts/ifcfg-br0
in teardown

smruti77:
#1701

abdhaleegit:
closing this

PR #1701, opened 2020-05-07
added code to remove bridge config file

smruti77:
after delete bridge interface also if we will do ifup bridge interface,
that interface is coming back.

### DIFF
--- a/io/net/bridge.py
+++ b/io/net/bridge.py
@@ -16,6 +16,7 @@
 # Bridge interface test
 
 
+import os
 import netifaces
 from avocado import main
 from avocado import Test
@@ -80,9 +81,12 @@ class Bridging(Test):
         local = LocalHost()
         networkinterface = NetworkInterface(self.bridge_interface, local,
                                             if_type="Bridge")
-        networkinterface.add_ipaddr(self.ipaddr, self.netmask)
+        try:
+            networkinterface.add_ipaddr(self.ipaddr, self.netmask)
+            networkinterface.save(self.ipaddr, self.netmask)
+        except Exception:
+            networkinterface.save(self.ipaddr, self.netmask)
         networkinterface.bring_up()
-        networkinterface.save(self.ipaddr, self.netmask)
         if networkinterface.ping_check(self.peer_ip, count=5) is not None:
             self.fail('Ping using bridge failed')
         networkinterface.remove_ipaddr(self.ipaddr, self.netmask)
@@ -92,6 +96,10 @@ class Bridging(Test):
         Set to original state
         '''
         self.check_failure('ip link del dev %s' % self.bridge_interface)
+        # TODO:need to get this functionality into avocado utils interfcae.py
+        path = "/etc/sysconfig/network-scripts/ifcfg-%s" \
+               % self.bridge_interface
+        os.remove(path)
 
 
 if __name__ == "__main__":


smruti77:
[bridge_fix.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/4592004/bridge_fix.log)

PraveenPenguin:
@abdhaleegit I think this PR address issue https://github.com/avocado-framework-tests/avocado-misc-tests/issues/1682 please have a look

io/net/bridge.py
```diff
@@ -92,6 +95,9 @@ def test_bridge_delete(self):
         Set to original state
         '''
         self.check_failure('ip link del dev %s' % self.bridge_interface)
+        cmd = "rm -rf /etc/sysconfig/network-scripts/ifcfg-%s" \
+              % self.bridge_interface
+        process.system(cmd, shell=True)
```
narasimhan-v, line 100: What is the reason to do this in test instead of utils ?

smruti77: @narasimhan-v this one is for remove cfg file for bridge interface.after link delete also if we do ifup of thta interface name its coming back.

PraveenPenguin: i can say use `os.remove` @narasimhan-v this we are in utils it just delete a file

narasimhan-v: @smruti77 Did we try to add this to utils ? Would be better if handled via utils

narasimhan-v: @PraveenPenguin I was talking about avocado network utils

PraveenPenguin: may we can add in genio utils like generic to remove as I did not feel wider use case for n/w utils either as I say just we can use `os.remove` from test

smruti77: ok then i will use os.remove

smruti77:
JOB ID     : 447015f863d5f04fd1bd7caa2bd9634a7e1e9d2e
JOB LOG    : /root/avocado-fvt-wrapper/results/job-2020-05-08T04.46-447015f/job.log
 (1/3) bridge.py:Bridging.test_bridge_create;run-6007: PASS (0.61 s)
 (2/3) bridge.py:Bridging.test_bridge_run;run-6007: PASS (4.80 s)
 (3/3) bridge.py:Bridging.test_bridge_delete;run-6007: PASS (0.76 s)
RESULTS    : PASS 3 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 6.31 s
JOB HTML   : /root/avocado-fvt-wrapper/results/job-2020-05-08T04.46-447015f/results.html

io/net/bridge.py
narasimhan-v: @PraveenPenguin @smruti77 I am confused here.

Why should network configure utils not handle it ? How is it handled in other tests ?

PraveenPenguin: @narasimhan-v i am saying it just removing a file  , it can be part of utils but adding one line u feel we need to be part of utils as other case we need backup and we restore it from back but here we need to remove interface file

io/net/bridge.py
```diff
@@ -92,6 +96,9 @@ def test_bridge_delete(self):
         Set to original state
         '''
         self.check_failure('ip link del dev %s' % self.bridge_interface)
+        path = "/etc/sysconfig/network-scripts/ifcfg-%s" \
+               % self.bridge_interface
+        os.remove(path)
```
narasimhan-v, line 101: @smruti77 how is this handled for normal interface, in other tests ?
I want same process to be followed for all types of interfaces, by network utils.

PraveenPenguin: as for other cases, we are doing this way using 
https://github.com/avocado-framework-tests/avocado-misc-tests/blob/11597cc6cd465827b93e61bb00919ad2c587a952/io/net/Network_config.py#L161

as you see restore_backup used backup interface if bridge always be the case to remove we can be code having check interface type but did not feel any harm to using os.remove, any way @smruti77 can take a call :)

narasimhan-v: @PraveenPenguin We have been doing a lot of work in the past months to push a lot of functionality to utils, and then use that to make common parts of code a lot more uniform. This, maintenance of code in future becomes easier. So, I would prefer adding hooks to `restore_from_backup()` for bridge as well.

@smruti77 your views please

PraveenPenguin: @narasimhan-v sure got you, so in bride case, we need to remove always the interface file?

smruti77: @narasimhan-v for all interface why need removed of config file, any way we are restoring the original config.

narasimhan-v: @smruti77 can answer this exactly, since I am unaware of what differences bridges have to normal interfaces, and what hooks are needed.

PraveenPenguin: > @narasimhan-v for all interface why need to removed of config file, any way we are restoring the original config from backup.

@smruti77 no his point is we can resue  `restore_from_backup` having bridge support as well 
so my question is in bride case we always remove interface file or we can tweak another way i can quickly come up with lib patch

narasimhan-v: @smruti77 Can this file removal be done by `restore_from_backup()` or `unset_ip()` ?

narasimhan-v: if not, we should add this to one of those functions.

smruti77: @narasimhan-v it will be not done by any of the function, but we don't want to removed config file always.

smruti77: @narasimhan-v for bridge interface actually we are  facing issue, after remove the interface also, if we will call ifup br0 that interfcae is coming back,because of config file. so i deleted that.but for other interface we don't need to remove config file.

smruti77: @narasimhan-v another thing if i will add remove config file here, restore_from_backup() .
then what is use of move backup file.

narasimhan-v: @smruti77 What I wanted to say is: can we add a special case in one of those 2 util functions (wherever applicable), to remove this file for bridge interfaces ?

Do I make sense ?

smruti77: @narasimhan-v let it be like this for now, if we need  for more test then we will add it to util. that will be better.

PraveenPenguin: > @narasimhan-v for bridge interface actually we are facing issue, after remove the interface also, if we will call ifup br0 that interfcae is coming back,because of config file. so i deleted that.but for other interface we don't need to remove config file.

just thinking is this not a bug ?

narasimhan-v: > > @narasimhan-v for bridge interface actually we are facing issue, after remove the interface also, if we will call ifup br0 that interfcae is coming back,because of config file. so i deleted that.but for other interface we don't need to remove config file.
> 
> just thinking is this not a bug ?

@smruti77 ^^

narasimhan-v: > @narasimhan-v let it be like this for now, if we need for more test then we will add it to util. that will be better.

Okay, though not ideal case, makes sense.

smruti77: > > > @narasimhan-v for bridge interface actually we are facing issue, after remove the interface also, if we will call ifup br0 that interfcae is coming back,because of config file. so i deleted that.but for other interface we don't need to remove config file.
> > 
> > 
> > just thinking is this not a bug ?
> 
> @smruti77 ^^

@narasimhan-v  @PraveenPenguin i checked in abdul machine and in mine machine also same thing is happening.

abdhaleegit: @narasimhan-v @PraveenPenguin 
1. for a normal interface, we expect the cfg file created and always stays in sysfs  so backup restore works as per utils
2. in case of bridge, the bridge is created using bridge utils and not using lib utils... so when bridge interface is created from test, the distro is basically creating a cfg file for br0 (which is odd and not same on all distros). now there is no need for br0 sysfs entry to stay ... as this will effect other interface as br0 will use same ip of backing interface... 

so at the end of bridge test it is cleaning up the br0 config created on the fly.. but we are not cleaning the br0 sysfs entry created by distro

as per your suggestion, yes it is always good to push all the bridge creation/deletion utils to lib.. we will take this as TODO

@smruti77

abdhaleegit: this patch looks good to me for now.. it can be merge

narasimhan-v: @smruti77 please add a TODO, similar to https://github.com/avocado-framework-tests/avocado-misc-tests/blob/master/io/net/virt-net/lpm.py#L256

narasimhan-v: Once done, we can merge it.

smruti77: ok

narasimhan-v: @abdhaleegit thanks for the ACK. Merging

PR #1721, opened 2020-05-12
Fix perf1

PraveenPenguin:
Fix libunwind test
1-as old url dead so changing url to github which is recent and working
2-  add default configure option as without YAML also test works
3- fix python3 issue

### DIFF
--- a/perf/libunwind.py
+++ b/perf/libunwind.py
@@ -19,7 +19,7 @@ import re
 
 from avocado import Test
 from avocado import main
-from avocado.utils import archive, build, distro, process
+from avocado.utils import archive, build, distro, process, cpu
 from avocado.utils.software_manager import SoftwareManager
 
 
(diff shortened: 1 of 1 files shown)


narasimhan-v:
@PraveenPenguin please add additional reviewers, if needed

PraveenPenguin:
@narasimhan-v I don't think so addition reviewer needed :)

perf/libunwind.py
```diff
@@ -48,18 +48,29 @@ def setUp(self):
...
+        https://github.com/libunwind/libunwind
         '''
+        configure_option = ''
+        if cpu.get_vendor() == 'intel':
+            configure_option = 'CC=icc CFLAGS="-g -O3 -ip" CXX=icc CCAS=gcc ' \
+                               'CCASFLAGS=-g LDFLAGS="-L$PWD/src/.libs"'
+        elif cpu.get_vendor() == 'ibm':
+            configure_option = 'CFLAGS="-g -O2 -m64" CXXFLAGS="-g -O2 -m64"'
+        else:
+            self.cancel(
```
narasimhan-v, line 68: Why are we cancelling before reading yaml parameter ?

PraveenPenguin: ah right let me see how I can tweak this

PraveenPenguin:
@narasimhan-v added that check, please have a look

perf/libunwind.py
```diff
@@ -48,18 +48,32 @@ def setUp(self):
...
         '''
         For configure options on different architecture please refer
-        https://github.com/pathscale/libunwind
+        https://github.com/libunwind/libunwind
         '''
         configure_option = self.params.get('configure_option',
-                                           default='configure_option')
+                                           default=None)
+
+        if configure_option:
```
narasimhan-v, line 64: Is this check necessary ?

PR #1722, opened 2020-05-12
Fix issue

PraveenPenguin:
1- As earlier condition never works so having proper check 
2- fixed pep8 issue

### DIFF
--- a/perf/perf_hv_gpci.py
+++ b/perf/perf_hv_gpci.py
@@ -17,7 +17,7 @@
 import platform
 from avocado import Test
 from avocado import main
-from avocado.utils import cpu, distro, process
+from avocado.utils import distro, process, genio
 from avocado.utils.software_manager import SoftwareManager
 
 
@@ -43,7 +43,7 @@ class perf_hv_gpci(Test):
         if 'ppc64' not in detected_distro.arch:
             self.cancel('This test is not supported on %s architecture'
                         % detected_distro.arch)
-        if 'PowerNV' in cpu._get_info():
+        if 'PowerNV' in genio.read_file('/proc/cpuinfo'):
             self.cancel('This test is only supported on LPAR')
 
         deps = ['gcc', 'make']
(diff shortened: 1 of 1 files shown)


perf/perf_hv_gpci.py
```diff
@@ -43,7 +43,7 @@ def setUp(self):
         if 'ppc64' not in detected_distro.arch:
             self.cancel('This test is not supported on %s architecture'
                         % detected_distro.arch)
-        if 'PowerNV' in cpu._get_info():
+        if 'PowerNV' in open('/proc/cpuinfo', 'r').read():
```
harish-24, line 46: Suggestion: `genio` can be used here

perf/perf_hv_gpci.py
```diff
@@ -65,14 +65,17 @@ def setUp(self):
...
+        # the test.
         output = process.run("dmesg -c")
 
     def error_check(self):
         if len(self.fail_cmd) > 0:
             for cmd in range(len(self.fail_cmd)):
-                self.log.info("Failed command: %s" % self.fail_cmd[cmd])
-            self.fail("perf_raw_events: some of the events failed, refer to log")
+                self.log.info("Failed command: %s"
+                              % self.fail_cmd[cmd])
```
harish-24, line 76: Any reason for this line change? I see it is already within the 80 character limit.

PraveenPenguin: ah autopep8 does that I think no harm if have this way

harish-24: Do you mean change is okay with no specific reason? I feel it is unnecessary as it does not add value.

PraveenPenguin: No i am saying pep8 commit just included this cahnge  but it has other changes as well if you say i can change this will do so

harish-24: other changes are fine, just wanted this line change to be removed as it there is no new value to it.

PraveenPenguin:
@harish-24 thanks for review and ACK

PR #1727, opened 2020-05-13
pci_hotplug: cancel test if pci not given

abdhaleegit:
make sure the default value is porper and skip the test
for all non pci devices

### DIFF
--- a/io/pci/PowerVMEEH.py
+++ b/io/pci/PowerVMEEH.py
@@ -65,7 +65,7 @@ class PowerVMEEH(Test):
         if '0x1' not in genio.read_file(eeh_enable_file).strip():
             self.cancel("EEH is not enabled, please enable via FSP")
         self.max_freeze = self.params.get('max_freeze', default=1)
-        self.pci_addr = [self.params.get('pci_device', default='')]
+        self.pci_addr = [self.params.get('pci_device', default="")]
         self.add_cmd = self.params.get('additional_command', default='')
         if not self.pci_addr:
             self.cancel("No PCI Device specified")
--- a/io/pci/pci_hotplug.py
+++ b/io/pci/pci_hotplug.py
@@ -57,10 +57,11 @@ class PCIHotPlugTest(Test):
             if not linux_modules.module_is_loaded("pnv_php"):
                 linux_modules.load_module("pnv_php")
         self.dic = {}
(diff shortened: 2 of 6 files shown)


io/pci/pci_hotplug.py
```diff
@@ -58,10 +58,12 @@ def setUp(self):
             if not linux_modules.module_is_loaded("pnv_php"):
                 linux_modules.load_module("pnv_php")
         self.dic = {}
-        self.device = self.params.get('pci_devices', default=' ').split(" ")
-        self.count = int(self.params.get('count', default='1'))
-        if not self.device:
+        self.device = self.params.get('pci_devices', default=None)
+        if self.device:
+            self.device = self.device.split(" ")
+        else:
```
narasimhan-v, line 64:
```
if not self.device:
    self.cancel("PCI_address not given")
self.device = self.device.split(" ")
```

io/pci/PowerVMEEH.py
```diff
@@ -66,7 +66,7 @@ def setUp(self):
         if '0x1' not in genio.read_file(eeh_enable_file).strip():
             self.cancel("EEH is not enabled, please enable via FSP")
         self.max_freeze = self.params.get('max_freeze', default=1)
-        self.pci_addr = [self.params.get('pci_device', default='')]
+        self.pci_addr = [self.params.get('pci_device', default=None)]
```
narasimhan-v, line 69: Please have this as a separate PR.
We will need to review Readme file to see if space is what it is separated by.

io/pci/pci_hotplug.py
Naresh-ibm, line 64: the existing logic is taken care of cancelling if pci_devices are not given, then may I know why we are changing this ?

narasimhan-v:
@abdhaleegit Is this still being pursued ?

io/pci/pci_hotplug.py
abdhaleegit: self.device = self.params.get('pci_devices', default=' ').split(" ")  >> was failing with error if no pci device was given.

so first check if pci present and than split is the right way

abdhaleegit:
@vnarasimhan included all the comments, and this can be merged

io/pci/PowerVMEEH.py
abdhaleegit: I now have 3 separate commits

PR #1758, opened 2020-05-28
Python fixes and debian environment handling

tfauck:
Add decode() where needed
    Add debian as known distro
    Add some CFLAGS parameters to avoid compilation errors

### DIFF
--- a/io/net/virt-net/lpm.py
+++ b/io/net/virt-net/lpm.py
@@ -58,6 +58,8 @@ class LPM(Test):
     '''
     @skipUnless("ppc" in distro.detect().arch,
                 "supported only on Power platform")
+    @skipIF(distro.detect().name in ['fedora', 'centos'],
+            "Not supported on distro.detect().name")
     @skipIf(IS_POWER_NV or IS_KVM_GUEST,
             "This test is not supported on KVM guest or PowerNV platform")
     def setUp(self):
@@ -227,7 +229,7 @@ class LPM(Test):
         detected_distro = distro.detect()
         self.log.info("Test is running on: %s", detected_distro.name)
         for pkg in ['ksh', 'src', 'rsct.basic', 'rsct.core.utils',
-                    'rsct.core', 'DynamicRM']:
+                    'rsct.core', 'DynamicRM', 'python3-pexpect']:
             if not smm.check_installed(pkg) and not smm.install(pkg):
                 self.cancel('%s is needed for the test to be run' % pkg)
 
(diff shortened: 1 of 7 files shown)


tfauck:
There is an issue with lpm module that I think I didn't push as it imports something which doesn't exist by default. (pxssh  depending on pexpect). That's a common problem to many recent tests. I would put pexpect as a default prereq for each distro

io/net/virt-net/lpm.py
```diff
@@ -227,7 +229,7 @@ def install_packages(self):
         detected_distro = distro.detect()
         self.log.info("Test is running on: %s", detected_distro.name)
         for pkg in ['ksh', 'src', 'rsct.basic', 'rsct.core.utils',
-                    'rsct.core', 'DynamicRM']:
+                    'rsct.core', 'DynamicRM', 'python3-pexpect']:
```
harish-24, line 232: The test would fail when trying to import the `pxssh` module. So this may not solve any problem. Or am I missing any other value add with this?

toolchain/gcc.py
```diff
@@ -89,7 +93,7 @@ def test(self):
             process_kwargs={'ignore_status': True})
         self.summary = ret.stdout.splitlines()
```
harish-24, line 94: Better to add `decode()` here so that the other part of code can still remain the same.

toolchain/libpfm.py
```diff
@@ -61,7 +61,7 @@ def setUp(self):
                 self.path = softm.get_source('libpfm', sourcedir)
 
         os.chdir(self.path)
-        build.make(self.path)
+        build.make(self.path, env={"-Wno-format-truncation"})
```
harish-24, line 64: Make sure this works for all distributions.

toolchain/papiTest.py
```diff
@@ -51,7 +51,8 @@ def setUp(self):
             self.path = os.path.join(self.path, 'src')
 
         os.chdir(self.path)
-        process.run('./configure', shell=True)
+        process.run('./configure', shell=True,
+                    env={"CC": "gcc -Wno-override-init -Wno-format-truncation"})
```
harish-24, line 55: same here

toolchain/papiTest.py
```diff
@@ -65,6 +66,7 @@ def test(self):
         errors = 0
         warns = 0
         for line in result.stdout.splitlines():
```
harish-24, line 68: here too `result.stdout.decode().splitlines()` would be better

tfauck:
closed with #1760 using comments

io/net/virt-net/lpm.py
narasimhan-v: User is expected to have pexpect installed before starting this test.

PR #1765, opened 2020-06-09
Add test to validate numa associativity for pmem device

harish-24:
Patch adds test to validate device-tree node associativity against OS level node associativity for pmem devices

### DIFF
--- a/memory/pmem_dt_check.py
+++ b/memory/pmem_dt_check.py
@@ -45,9 +45,12 @@ class NdctlDeviceTreeCheck(Test):
                          [1:]).strip().split()
         return values
 
-    def get_ref_point_index(self):
-        val = self.get_hex_list(
-            "/proc/device-tree/rtas/ibm,associativity-reference-points")
+    def get_ref_point_index(self, legacy=False):
+        path = "/proc/device-tree/rtas/ibm,associativity-reference-points"
+        if legacy:
+            path = "/proc/device-tree/ibm,opal/"\
+                   "ibm,associativity-reference-points"
+        val = self.get_hex_list(path)
         return int(val[0], 16)
 
     def get_node_id_from_associativity(self, assoc_ref_index, filename):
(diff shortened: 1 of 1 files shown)


harish-24:
``` 
(1/1) memory/pmem_dt_check.py:NdctlDeviceTreeCheck.test: PASS (2.68 s)
JOB TIME   : 2.74 s
```

memory/pmem_dt_check.py
```diff
@@ -0,0 +1,161 @@
...
+                                "Exp:%s, Got:%s" % (reg_no, dt_node, node))
+            else:
+                self.log.info("Working as expected. "
+                              "ndctl node: %s, DT node: %s", node, dt_node)
+        if failures:
+            self.fail(failures)
+
+    @avocado.fail_on(pmem.PMemException)
+    def tearDown(self):
+        self.plib.disable_region()
```
narasimhan-v, line 157:
```
if self.plib:
    self.plib.disable_region()
```

memory/pmem_dt_check.py
```diff
@@ -0,0 +1,161 @@
...
+        for val in buses:
+            provider = self.plib.run_ndctl_list_val(val, 'provider')
+            region = self.plib.run_ndctl_list_val(val, 'regions')
+            reg_no = self.plib.run_ndctl_list_val(region[0], 'dev')
+            node_path = '/sys/bus/nd/devices/%s/target_node' % reg_no
+            if not os.path.exists(node_path):
+                # Fallback to numa_node if target_node support does not exist
+                node_path = '/sys/bus/nd/devices/%s/numa_node' % reg_no
+            node = genio.read_one_line(node_path)
+            dt_node = str(self.parse_pmem_dt(assoc_ref_index, provider))
```
narasimhan-v, line 145: Just for better readability:
```
            region = self.plib.run_ndctl_list_val(val, 'regions')
            reg_no = self.plib.run_ndctl_list_val(region[0], 'dev')
            node_path = '/sys/bus/nd/devices/%s/target_node' % reg_no
            if not os.path.exists(node_path):
                # Fallback to numa_node if target_node support does not exist
                node_path = '/sys/bus/nd/devices/%s/numa_node' % reg_no
            node = genio.read_one_line(node_path)

            provider = self.plib.run_ndctl_list_val(val, 'provider')
            dt_node = str(self.parse_pmem_dt(assoc_ref_index, provider))
```

memory/pmem_dt_check.py
```diff
@@ -0,0 +1,161 @@
...
+        return values
+
+    def get_ref_point_index(self):
+        val = self.get_hex_list(
+            "/proc/device-tree/rtas/ibm,associativity-reference-points")
+        return int(val[0], 16)
+
+    def get_node_id_from_associativity(self, assoc_ref_index, filename):
+        values = self.get_hex_list(filename)
+        return int(values[assoc_ref_index], 16)
```
narasimhan-v, line 56: Suggestion: would it not be better if decimal conversion is done in `get_hex_list()` itself ?
You could have a default value to the list index as 0, and override it when calling from `get_node_id_from_associativity()` with `assoc_ref_index`

harish-24: It is possible, but I would like to keep this way for two reasons.
1. better readability, so that it is straightforward to understand what is used for what
2. In future, `get_hex_list` may be used to get values which does not need any decimal conversion.

narasimhan-v: okay, agree !!

PR #1761, opened 2020-06-04
passwordless ssh for hmc

smruti77:
removed login() and run_command(), handle through ssh module
for running command on hmc in dlpar.py and network_virtualization.py

Signed-off-by: bismurti bidhibrata pattajoshi <bbidhibr@in.ibm.com>

PR #1770, opened 2020-06-11
passwordless ssh for hmc

smruti77:
removed login() and run_command(), handle through ssh module
for running command on hmc in dlpar.py and network_virtualization.py


narasimhan-v:
Merging based on reviews from  #1761

PR #1793, opened 2020-06-29
Adding 24 hours yamls for HTX tests

narasimhan-v:
Adding 24 hours yamls for HTX tests.

### DIFF
--- a/io/disk/htx_block_devices.py.data/htx_all_devices_24_hours.yaml
+++ b/io/disk/htx_block_devices.py.data/htx_all_devices_24_hours.yaml
@@ -1,4 +1,4 @@
 htx_disks: ''
 all: True
-time_limit: 1
+time_limit: 1440
 mdt_file: 'mdt.hd'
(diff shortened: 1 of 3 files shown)


narasimhan-v:
@PraveenPenguin These are just yaml changes. I hope they can be merged ?

narasimhan-v:
> @narasimhan-v I feel YAML file can be maintained via user in upstream we can have only sample copy

@PraveenPenguin I do not understand. Can you please explain a bit more.

PraveenPenguin:
> > @narasimhan-v I feel YAML file can be maintained via user in upstream we can have only sample copy
> 
> @PraveenPenguin I do not understand. Can you please explain a bit more.

what i meant as we can have sample YMAL file or if test case has nature to pass to have another YAML  that time also we have sample YAML, if  user want to his custom YAML he/she need to have it saved in local as i don't feel it should be part of the upstream repo  as in this case we have two YAML 
1-htx_nic_devices_24_hours.yaml (from the block device  we can have commented part what need to be there )
2- htx_all_devices.yaml

narasimhan-v:
@PraveenPenguin Are we discussing 2 things here ?
1. multiple yamls
2. upstream vs local

narasimhan-v:
@PraveenPenguin I just realised I don't need the duplicate yamls. Will send an update by changing the default timing on them. Thanks

PR #1794, opened 2020-06-30
add connect() from ssh.py to avoid ssh-keygen

smruti77:
to avoid password during command run in peer machine.

### DIFF
--- a/io/net/bonding.py
+++ b/io/net/bonding.py
@@ -118,6 +118,8 @@ class Bonding(Test):
         self.log.info("Bond Test on IB Interface? = %s", self.ib)
         self.session = Session(self.peer_first_ipinterface, user=self.user,
                                password=self.password)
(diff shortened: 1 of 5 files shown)


io/net/bonding.py
```diff
@@ -118,6 +118,8 @@ def setUp(self):
         self.log.info("Bond Test on IB Interface? = %s", self.ib)
         self.session = Session(self.peer_first_ipinterface, user=self.user,
                                password=self.password)
+        if not self.session.connect():
+            self.fail("failed connecting to peer")
```
narasimhan-v, line 122: let us use `cancel` in setUp()

smruti77: ok

narasimhan-v:
@smruti77 please resolve conflicts also here.

PR #1818, opened 2020-07-09
cpu 0 is not offline

tfauck:
no need to online cpu 0 as it is not offlined


cpu/em_cpuhotplug.py
```diff
@@ -90,7 +90,8 @@ def test(self):
                 self.log.info("Onlining the threads : %s for "
                               "the core : %s" % (cpu_list, core))
                 for cpu_num in cpu_list:
-                    self.online_cpu(cpu_num)
+                    if cpu_num != 0:
```
harish-24, line 93: `cpu_list.remove(0)` before loop?

tfauck:
sorry - created 1819 by mistake - so closing this one - took your point in account

PR #1839, opened 2020-07-27
Fix: Tests binaries access for LTP

harish-24:
Like temporary directories' access, there are few binaries which are part of the build directory that needs access when tests are
utilizing them. Hence replacing teststmpdir as the build directory so that tests still have access to the entire path of binary.


harish-24:
@PraveenPenguin Re-based fixing the conflicts.

PR #1850, opened 2020-07-31
Fix perf install command by linux-perf for debian

utest2020:
Fix perf install command by linux-perf for debian as bellow files

perf_basic.py
perf_c2c.py
perf_fuzzer.py

Signed-off-by:  xiaohaiyan <xiaohaiyan@uniontech.com>

### DIFF
--- a/perf/perf_basic.py
+++ b/perf/perf_basic.py
@@ -53,11 +53,13 @@ class PerfBasic(Test):
     def setUp(self):
         smg = SoftwareManager()
         dist = distro.detect()
-        if dist.name in ['Ubuntu', 'debian']:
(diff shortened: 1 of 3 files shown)


tfauck:
Doesn't understand why Ubuntu is not included in the changes.
There are other tests requiring perf to be installed so I thought it was going to be installed at a higher level.

utest2020:
hi,tfauck,thanks for your reply.As our work environment just base on debian system, we not adjust the test cases on Ubuntu.if necessary，we will add this test later.At the same time,we will try  other tests requiring perf to be installed and give another submit.Thanks again.

PraveenPenguin (approved):
LGTM

PR #1860, opened 2020-08-17
support loongson kernel-4.19.90

zhengxingwang123:
support loongson kernel-4.19.90


zhengxingwang123:
support loongson kernel-4.19.90

PR #1868, opened 2020-08-25
ndctl: Remove "all" from disable/enable test case

santoshs:
Removing "all" is valid because if create-namespace fails, the pmem lib will
throw an error. And if one of the namespace disable fails, it will still throw
an error. So we got all the cases covered. This "all" is redundant.

If in future, ndctl treats "all" to behave differently for zero namespaces, we
will again have to change the test cases. Future proof it.

Signed-off-by: Santosh S <santosh@fossix.org>

### DIFF
--- a/memory/ndctl.py
+++ b/memory/ndctl.py
@@ -255,7 +255,6 @@ class NdctlTest(Test):
         ns_names = []
         for ns in namespaces:
             ns_names.append(self.plib.run_ndctl_list_val(ns, 'dev'))
-        ns_names.append('all')
 
         for namespace in ns_names:
             self.plib.disable_namespace(namespace=namespace)


harish-24:
> Removing "all" is valid because if create-namespace fails, the pmem lib will
> throw an error. And if one of the namespace disable fails, it will still throw
> an error. So we got all the cases covered. This "all" is redundant.
> 

That is not completely right. `create-namespace` does throw the error and so does `disable` for one of the namespaces. But in the past we have seen instances where disable/enable works for one-off namespaces' but failed when using `all`. 

> If in future, ndctl treats "all" to behave differently for zero namespaces, we
> will again have to change the test cases. Future proof it.
> 

Also in this test we are not doing this operation for zero namespaces. We create N namespaces, try disable/enable each of them, followed by disable/enable all. So we need not remove it as it would not consider zero namespaces here.

@santoshs If you have any other comments opposing this please let know, otherwise I suggest closing this.

PR #1878, opened 2020-09-08
update perf/perfmon.py perf/perftool.py

bych-github:
…fails.The 'apt-cache search ncurses' command will solve this problem.

update:perf/perftool.py Compatible with linux-tools and linux-perf

perf/perfmon.py
perf/perftool.py

Signed-off-by: Jerry <b447467606@163.com>

### DIFF
--- a/perf/perfmon.py
+++ b/perf/perfmon.py
@@ -37,6 +37,7 @@ class Perfmon(Test):
 
         deps = ["gcc", "make"]
         if dist.name in ['Ubuntu', 'debian']:
+            process.system('apt-cache search ncurses')
             deps.extend(['libncurses-dev'])
         elif dist.name in ['rhel', 'SuSE']:
             deps.extend(['ncurses-devel'])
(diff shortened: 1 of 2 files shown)


PraveenPenguin:
@nasastry  can you please review this PR

perf/perfmon.py
```diff
@@ -37,6 +37,7 @@ def setUp(self):
 
         deps = ["gcc", "make"]
         if dist.name in ['Ubuntu', 'debian']:
+            process.system('apt-cache search ncurses')
```
PraveenPenguin, line 40: wondering why apt search require for package install, maybe in your setup package list not refreshed

bych-github: apt-get update？useless.

PraveenPenguin: no, what I meant i did not face this issue in my setup so I think it can be setup issue so aksing please revisit your setup so code change not require

bych-github: You can run the test in the newly installed operating system. May trigger this issue.

PR #1926, opened 2020-10-13
changes for multiple vnic with different adapter id

smruti77:
script will create mutiple vnic with adapter id and different port.
because we can't create bond interfcae with same slot num vnic
interfaces.

Signed-off-by: Bismruti Pattjoshi <bismruti@oc7660885027.ibm.com>

### DIFF
--- a/io/driver/driver_parameter_block_device.py.data/driver_parameter_block_device_ipr.yaml
+++ b/io/driver/driver_parameter_block_device.py.data/driver_parameter_block_device_ipr.yaml
@@ -1,0 +1,29 @@
+module: 'ipr'
+multipath_enabled: False
+disk: ''
+Test: !mux
+    fastfail:
+        module_param_name: "fastfail"
+        value: !mux
+            fastfail_180:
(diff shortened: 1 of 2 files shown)


smruti77:
[vnic.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/5371010/vnic.log)

abdhaleegit (changes requested):
Please push the sample yaml for these changes to be used and also add the definitions in Readme

1. yaml for single vnic device (no backing device)
2. yaml for single vnic with single backing device
3. yaml for multiple vnic device with out backing device
4. yaml for multiple vnic with multiple backing device

also add which of the option do not work for now in README.

TODO list

abdhaleegit:
@smruti77  also a regression run with other variation to make sure nothing breaks.. paste the logs for all the working scenario

smruti77:
@abdhaleegit  for now test is not support for multiple vnic with multiple backing device.

smruti77:
#1948

PR #1920, opened 2020-10-05
adding physical_adapter check

vaishnavibhat:
adding physical_adapter param to check of the tested
interface is physical or virtual adapter

Signed-off-by: Vaishnavi Bhat <vaishnavi@linux.vnet.ibm.com>

PR #1938, opened 2020-10-23
bonding: test changes the peer's interface MTU indiscriminatley

cforno12:
Currently the bonding test would change the MTU of the peer's interface from
2000-9000. If the peer's interface is an ibmvnic interface, vNIC doesn't support
MTU values that aren't 1500 or 9000 and the entire test fails since it cannot
change the MTU.

This fix checks if the peer's interface is an ibmvnic interface and if it is,
the test will only change the MTU if it is 9000.

### DIFF
--- a/io/net/bonding.py
+++ b/io/net/bonding.py
@@ -296,8 +296,17 @@ class Bonding(Test):
             for interface in self.peer_interfaces:
                 peer_networkinterface = NetworkInterface(interface,
                                                          self.remotehost)
-                if peer_networkinterface.set_mtu(mtu) is not None:
-                    self.cancel("Failed to set mtu in peer")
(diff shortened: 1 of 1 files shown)


cforno12:
I acknowledge that the `self.session.cmd(cmd).stdout.decode("utf-8")` line is costly. The only way I know how to check if the interface is using the ibmvnic driver is either through sysfs or lsdevinfo, both requiring to send a command.

I could also refactor the code such that the outer for loop is iterating through all the peer's interface instead of the MTUs, resulting in less `self.session.cmd()` calls. Please comment on your thoughts!

vaishnavibhat:
Waiting on PR addressing this issue. https://github.com/avocado-framework-tests/avocado-misc-tests/pull/1920

cforno12:
Closing this PR as Vaishnavi has a better solution to this problem.

PR #1960, opened 2020-11-10
virt-net: Fail test if VIOS could not connect

cforno12:
Currently, when using avocado's ssh utils to create a session with the VIOS, it
does not check if the connection was successful. This fix fails thetest if
connecting to the VIOS has failed.

### DIFF
--- a/io/net/virt-net/network_virtualization.py
+++ b/io/net/virt-net/network_virtualization.py
@@ -116,10 +116,6 @@ class NetworkVirtualization(Test):
         self.vios_ip = self.params.get('vios_ip', '*', default=None)
         self.vios_user = self.params.get('vios_username', '*', default=None)
         self.vios_pwd = self.params.get('vios_pwd', '*', default=None)
-        if 'vios' in str(self.name.name):
-            self.session = Session(self.vios_ip, user=self.vios_user,
-                                   password=self.vios_pwd)
-            self.session.connect()
         self.count = int(self.params.get('vnic_test_count', default="1"))
         self.num_of_dlpar = int(self.params.get("num_of_dlpar", default='1'))
         self.device_ip = self.params.get('device_ip', '*',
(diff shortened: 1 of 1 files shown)


io/net/virt-net/network_virtualization.py
```diff
@@ -113,7 +113,8 @@ def setUp(self):
         if 'vios' in str(self.name.name):
             self.session = Session(self.vios_ip, user=self.vios_user,
                                    password=self.vios_pwd)
-            self.session.connect()
+            if not self.session.connect():
+                self.fail("Failed connecting to VIOS")
```
narasimhan-v, line 117: @cforno12 Only one test (`test_rmdev_viosfailover`) uses VIOS login, so only that test should fail.

We should move this part of the code inside the test.

narasimhan-v (changes requested):

io/net/virt-net/network_virtualization.py
cforno12: Right, but there could be more tests in the future that use a VIOS connection, therefore we should check if the connection is successful when setting up.

narasimhan-v: @cforno12 What I meant is, we can push the vios session setup itself into those tests that use it. Else, we would end up failing tests that do not use a vios session at all.

Can you include that change in your patch please ?

cforno12: Sure, thanks Narasimhan

io/net/virt-net/network_virtualization.py
```diff
@@ -382,6 +378,12 @@ def test_rmdev_viosfailover(self):
         '''
         using mrdev and mkdev command to check vios failover works
         '''
+
+        self.session = Session(self.vios_ip, user=self.vios_user,
+                               password=self.vios_pwd)
+        if not self.session.connect():
+                self.fail("Failed connecting to VIOS")
```
smruti77, line 385: @cforno12 this check is needed but don't kept inside the test, keep in setUp() it self.

smruti77 (changes requested):

io/net/virt-net/network_virtualization.py
narasimhan-v: @smruti77 I asked for this to be moved inside test since there was only one test using it.

smruti77: @narasimhan-v  ok fine, i didn't saw that comments.

smruti77 (approved):

narasimhan-v:
@cforno12 Once you fix the indentation issue pointed by Travis, we can merge this.

narasimhan-v (approved):

PR #1966, opened 2020-11-26
enhancing the vfc-dlpar.py to work with vfc bootable disk

Naresh-ibm:
The script was failing when we have one of the vfc interfaces as bootable disk.
Hence added an option to skipp the bootable vfc interface and perform the dlpar
on rest of the vfc interfaces.

### DIFF
--- a/io/disk/vfc-tests/vfc_dlpar.py
+++ b/io/disk/vfc-tests/vfc_dlpar.py
@@ -52,6 +52,7 @@ class VirtualFC(Test):
         self.hmc_pwd = self.params.get("hmc_pwd", '*', default=None)
         self.hmc_username = self.params.get("hmc_username", '*', default=None)
         self.count = self.params.get("count", default=1)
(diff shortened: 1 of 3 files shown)


io/disk/vfc-tests/vfc_dlpar.py
```diff
@@ -52,6 +52,7 @@ def setUp(self):
         self.hmc_pwd = self.params.get("hmc_pwd", '*', default=None)
         self.hmc_username = self.params.get("hmc_username", '*', default=None)
         self.count = self.params.get("count", default=1)
+        self.skip_drc = self.params.get("skip_drc_name", default=None)
```
narasimhan-v, line 55: @Naresh-ibm Skipping by drc is the correct approach here.

But, it would be preferred to find that boot disk drc without user intervention.

narasimhan-v (approved):

PR #1978, opened 2020-12-09
virt-net: test interface every time DLPAR remove and add occurs

cforno12:
Currently, the interface is only tested after X amount of DLPAR remove/add operations.
This fix adds an ip addr, brings the link up, and pings the peer with the interface
after every remove and add.

### DIFF
--- a/io/net/virt-net/network_virtualization.py
+++ b/io/net/virt-net/network_virtualization.py
@@ -424,7 +424,7 @@ class NetworkVirtualization(Test):
 
     def test_vnic_dlpar(self):
         '''
-        Perform vNIC device hot add and hot remove using drmgr command
+        Perform vNIC device hot add and hot remove
         '''
         for slot_no, device_ip, netmask, mac, peer_ip in zip(self.slot_num,
                                                              self.device_ip,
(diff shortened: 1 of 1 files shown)


io/net/virt-net/network_virtualization.py
```diff
@@ -433,30 +433,40 @@ def test_vnic_dlpar(self):
...
                     for _ in range(self.num_of_dlpar):
                         self.drmgr_vnic_dlpar('-r', slot)
+                        if self.interface_present(device_name):
+                            self.fail("DLPAR remove did not remove interface")
                         self.drmgr_vnic_dlpar('-a', slot)
-                        self.wait_intrerface(device_name)
+                        self.wait_interface(device_name)
+                        try:
+                            networkinterface.add_ipaddr(device_ip, netmask)
+                            networkinterface.bring_up()
```
cforno12, line 447: Please note that I added the command to bring up the interface up. When the interface is added back, it is in a DOWN state and the function `networkinterface.is_link_up` does not bring the interface up, rather checks if the interface is up.

abdhaleegit: @cforno12 looks fine, could you also add a line to log the iteration count, it will be easy to know if dlpar fails for which iteration step?

io/net/virt-net/network_virtualization.py
```diff
@@ -433,30 +433,40 @@ def test_vnic_dlpar(self):
...
-                if not wait.wait_for(networkinterface.is_link_up, timeout=120):
-                    self.fail("Unable to bring up the link on the Network \
-                              virtualized device")
-                if networkinterface.ping_check(peer_ip, count=5) is not None:
-                    self.fail("dlpar has affected Network connectivity")
             else:
                 self.fail("slot not found")
         self.check_dmesg_error()
 
+    def interface_present(self, device_name):
```
abdhaleegit, line 462: can this function be moved to nework utils

abdhaleegit (changes requested):
could you paste the run log files for these changes

cforno12:
Will update this PR when the utils PR is complete: https://github.com/avocado-framework/avocado/pull/4485

io/net/virt-net/network_virtualization.py
smruti77: @cforno12 suggest to add networkinterface.bring_up() after try...except block so it will bring up interface after ip added.

smruti77 (changes requested):

cforno12:
Logs are attached: [dlpar_logs.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/6418369/dlpar_logs.log)

PR #1980, opened 2020-12-11
Trinity test capture kernel tainting.

yiwei075:
Trinity test can tainting the kernel, add this error message capture.

Signed-off-by: yiwei <yiwei@uniontech.com>


PraveenPenguin:
@yiwei075 please wrap all commit in logical commit

PR #1982, opened 2020-12-15
Update trinity.py

yiwei075:
Patch adds checks for tainting kernel errors,
for example: Using fanotify permission c hecks may lead to deadlock;tainting kernel

Signed-off-by: yiwei <yiwei@uniontech.com>


PraveenPenguin (approved):
LGTM

PR #1991, opened 2021-01-11
virt-net: parsing for system name fix

cforno12:
Found a bug where I had a machine ltczep10 and another machine ltczep1. The test
would act on the latter machine because of the line `if line in self.lpar`.
This fix assumes the format of an LPAR
`<name_of_the_system>-lp<partiton_number>`.

### DIFF
--- a/io/net/virt-net/network_virtualization.py
+++ b/io/net/virt-net/network_virtualization.py
@@ -77,11 +77,12 @@ class NetworkVirtualization(Test):
             self.cancel("failed connecting to HMC")
         cmd = 'lssyscfg -r sys  -F name'
         output = self.session_hmc.cmd(cmd)
-        self.server = ''
-        for line in output.stdout_text.splitlines():
-            if line in self.lpar:
-                self.server = line
-                break
+        self.server = self.params.get("server", "*", default=None)
+        if not self.server:
+            for line in output.stdout_text.splitlines():
+                if line in self.lpar:
+                    self.server = line
+                    break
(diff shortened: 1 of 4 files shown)


cforno12:
I noticed that you have some CECs in your CI with the name `ltc-zz**-lp#`. How do you propose we expect the name of the CEC? Do we just add the name of the CEC as a param instead of trying to look for it?

io/net/virt-net/network_virtualization.py
```diff
@@ -78,8 +78,11 @@ def setUp(self):
         cmd = 'lssyscfg -r sys  -F name'
         output = self.session_hmc.cmd(cmd)
         self.server = ''
+        lpar_split = self.lpar.split('-', maxsplit=1)
+        if(len(lpar_split) != 2):
+            self.cancel("LPAR name incorrectly formatted. Expecting: <name>-lp<#>")
```
narasimhan-v, line 83: @cforno12 I don't think it is right to force lpar names. <name>-lp<#> is just a common nomenclature we follow.

narasimhan-v (changes requested):

cforno12:
You're right, we shouldn't enforce naming policy. What do you think of this change?

PraveenPenguin (approved):

PraveenPenguin:
@narasimhan-v  new changes looks good to me can you also please review

PraveenPenguin:
@abdhaleegit review Please

PR #1993, opened 2021-01-13
test_backingdev_add: remove check_slot_availablilty

cforno12:
Since the backing device shares the slot number with the newly added device,
there is no need to check if the slot_num given is available since it has
already been taken.


narasimhan-v:
@cforno12 Since the tests can be run separately, it is better we have this check.
Even better would be to handle it in setUp with test names

PR #1994, opened 2021-01-14
bridge.py: check if config file exists before deleting

cforno12:
Check if the config file exists before deleting it.


cforno12:
In SUSE, the network scripts are stored in `/etc/sysconfig/network/`. Need to add that path to the bridge cleanup test.

narasimhan-v (approved):

narasimhan-v:
> In SUSE, the network scripts are stored in `/etc/sysconfig/network/`. Need to add that path to the bridge cleanup test.

@cforno12 Are you planning to handle SuSE path also here ?

PR #1995, opened 2021-01-15
htx_nic_devices.py: get pkg manager of peer

cforno12:
Currently, the test assumes that the peer machine has the same package manager.
This fix retrieves the peer's OS and uses the appropriate package manager.

### DIFF
--- a/io/net/htx_nic_devices.py
+++ b/io/net/htx_nic_devices.py
@@ -94,7 +94,11 @@ class HtxNicTest(Test):
         Build 'HTX'
         """
         packages = ['git', 'gcc', 'make']
+        peer_packages = packages.copy()
+        peer_pkg_manager = None
         detected_distro = distro.detect()
+
+        #Host package detection
         if detected_distro.name in ['centos', 'fedora', 'rhel', 'redhat']:
             packages.extend(['gcc-c++', 'ncurses-devel', 'tar'])
         elif detected_distro.name == "Ubuntu":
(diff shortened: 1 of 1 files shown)


smruti77 (approved):

smruti77:
@narasimhan-v we have plan to add this in to utils,what about that.? or else this is fine.

narasimhan-v:
Most of our use cases have been with same distro on peer, so this was not a necessity so far. We need to evaluate the use cases.

Regarding utils part, @smruti77 any pointers ? I could not recollect the discussion.

smruti77:
> Most of our use cases have been with same distro on peer, so this was not a necessity so far. We need to evaluate the use cases.
> 
> Regarding utils part, @smruti77 any pointers ? I could not recollect the discussion.

@narasimhan-v didn't sent any PR for that. my point is like this PR we have also done but later we decided because of this code, don't want to include in script, so we have did for same distro on peer. and if we need package installation for different distro peer decided to include that in utils. my question is can we allow this in script..?

narasimhan-v:
> > Most of our use cases have been with same distro on peer, so this was not a necessity so far. We need to evaluate the use cases.
> > 
> > Regarding utils part, @smruti77 any pointers ? I could not recollect the discussion.
> 
> @narasimhan-v didn't sent any PR for that. my point is like this PR we have also done but later we decided because of this code, don't want to include in script, so we have did for same distro on peer. and if we need package installation for different distro peer decided to include that in utils. my question is can we allow this in script..?

Yes, right. The best place to include this is in utils

cforno12:
Would I have to contribute to this https://github.com/avocado-framework/avocado/blob/master/avocado/utils/distro.py ? Just want to be sure I'm looking in the right place...

I'm a bit unsure on what my action item is, do I want to enable the distro.py to detect the distro of a peer machine?

cforno12:
Started distro detection on peer on avocado utils. https://github.com/avocado-framework/avocado/pull/4783 please let me know if you have any concerns @narasimhan-v @smruti77

PraveenPenguin:
@cforno12 sorry for late  seems me it failing in trvais check , even if you @narasimhan-v perusing this Request please open new request

PR #1998, opened 2021-01-21
netperf: make UDP/TCP_STREAM option have dynamic socket buffer arg

cforno12:
Currently the netperf stream tests have a hard value of 63000 bytes given as an
argument when it should depend on each system. This fix reads the default socket
buffer size from /proc/ and sends it as an argument.

### DIFF
--- a/io/net/netperf_test.py
+++ b/io/net/netperf_test.py
@@ -148,7 +148,18 @@ class Netperf(Test):
         cmd = "timeout %s %s -H %s" % (self.timeout, self.perf,
                                        self.peer_ip)
         if self.option != "":
-            cmd = "%s -t %s" % (cmd, self.option)
+            if "TCP_STREAM" in self.option:
+                socket_size = process.run("cat /proc/sys/net/ipv4/tcp_rmem",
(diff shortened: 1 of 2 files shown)


io/net/netperf_test.py
```diff
@@ -98,6 +98,7 @@ def setUp(self):
                                             password=self.peer_password)
         self.peer_public_networkinterface = NetworkInterface(self.peer_interface,
                                                              self.remotehost_public)
+        self.networkinterface.bring_up()
```
smruti77, line 101: @cforno12  everything  is fine. but why its needed here ..?

smruti77 (changes requested):

io/net/netperf_test.py
abdhaleegit: self.networkinterface.bring_up() is already called in setup at line 66

abdhaleegit (approved):
LGTM

cforno12:
Removed the bring_up() line from the setup routine, please review.

smruti77 (approved):

narasimhan-v (approved):

PR #2008, opened 2021-02-17
migrate_pages.py: Handle return code if test needs to skip

harish-24:
Patch handles return code if test needs to skip


PraveenPenguin (approved):
LGTM

PR #2015, opened 2021-02-23
Trace testcases

kalshett:


perf/trace.py
```diff
@@ -0,0 +1,140 @@
...
+class Linuxtrace(Test):
+
+    """
+    Linux Trace tests
+    """
+
+    @staticmethod
+    def run_cmd_out(cmd):
+        return process.system_output(cmd, shell=True, ignore_status=True,
+                                     sudo=True)
```
nasastry, line 34: Output is not parsed to set the condition for failure. Please add this part.
If not your test case will always PASS irrespective of the actual test result.

kalshett: Okay, will incorporate this change.

kalshett: @nasastry Below is the code I will add with checking for failed case.
    def run_cmd_out(self, cmd):
        results = process.system_output(cmd, shell=True, sudo=True)
        match = re.search(br'fail', results, re.M | re.I)
        if match:
           self.log.info("Test failed: %s" % match)
        return results

Let me know if this would sufficient.

perf/trace.py
```diff
@@ -0,0 +1,140 @@
...
+                install_rpm_cmd = ("rpm -i %s/%s.rpm" %
+                                   (self.workdir, kernel_src))
+                self.run_cmd_out("%s" % install_rpm_cmd)
+
+                src_def_path = "/root/rpmbuild/SOURCES"
+                linux_name = "linux-%s.tar.xz" % self.output
+                tarball = "%s/%s" % (src_def_path, linux_name)
+                archive.extract(tarball, self.workdir)
+                os.chdir("%s/linux-%s/%s" % (self.workdir, self.output,
+                         self.testdir))
```
harish-24, line 74: This entire code can be largely reduced by get_source library from avocado
```
            if detected_distro.name in ['centos', 'rhel']:
                src_name = 'kernel'
                self.buldir = smg.get_source(src_name, self.workdir)
                self.buldir = os.path.join(
                    self.buldir, os.listdir(self.buldir)[0])
                os.chdir(os.path.join(self.buldir, self.testdir))
```

perf/trace.py
```diff
@@ -0,0 +1,140 @@
...
+            elif 'SuSE' in detected_distro.name:
+                if not smg.check_installed("kernel-source") and not\
+                        smg.install("kernel-source"):
+                    self.cancel(
+                        "Failed to install kernel-source for this test.")
+                if not os.path.exists("/usr/src/linux"):
+                    self.cancel("kernel source missing after install")
+                self.output = "/usr/src/linux"
+                os.chdir("%s/%s" % (self.output, self.testdir))
```
harish-24, line 83: please use `os.path.join` here to concat the paths

perf/trace.py
```diff
@@ -0,0 +1,140 @@
...
+
+            match = next(
+                (ext for ext in [".zip", ".tar"] if ext in location), None)
+            if match:
+                tarball = self.fetch_asset("kselftest%s" % match,
+                                           locations=[location], expire='1d')
+                archive.extract(tarball, self.workdir)
+            else:
+                git.get_repo(location, destination_dir=self.workdir)
+            os.chdir("%s/linux-%s/%s" % (self.workdir, self.output,
```
harish-24, line 97: also the actual  resultant test directory can be stored in a variable on each if block and a common chdir can be used after all the conditions

perf/trace.py
```diff
@@ -0,0 +1,140 @@
...
+                     self.testdir))
+
+    def test_ftrace_basic(self):
```
harish-24, line 100: instead of repeating each test with same steps each of the scenario can be make yaml configurable.

harish-24 (changes requested):

kalshett:
@nasastry pointed out that I have sent the updated PR on a master branch, which seems incorrect. Will close this and send in a new PR on trace.

PR #2017, opened 2021-02-25
virt-net: dynamic debug should only be turned off when the vNIC adapt…

cforno12-zz:
…er is removed

Currently, dynamic debug is turned off after every vNIC related test.
Dynamic debug should remain on when running other tests in the test
bucket. This patch turns off dynamic debug only when the vNIC adapter is
removed.

### DIFF
--- a/io/net/virt-net/network_virtualization.py
+++ b/io/net/virt-net/network_virtualization.py
@@ -840,7 +840,8 @@ class NetworkVirtualization(Test):
         self.session_hmc.quit()
         if 'vios' in str(self.name.name):
             self.session.quit()
-        cmd = "echo 'module ibmvnic -pt; func send_subcrq -pt' > /sys/kernel/debug/dynamic_debug/control"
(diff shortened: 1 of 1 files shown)


narasimhan-v:
@cforno12 The setUp() and tearDown() are run for every testX() tests, so the dynamic debug is turned on and off for every test. Do you think that is not enough ?

cforno12:
Yes, I don't think that's enough. 

Say we test_add -> run a bond test that enslaves the newly added vNIC, suppose it fails and the vNIC driver is at fault. We won't know what happened because dynamic debug was disabled after test_add was run. Am I making sense?

So my thinking is that it should only be disabled when we are done with the vNIC interface (at the end of test_remove).

vaishnavibhat (approved):
This looks like a good option , the dynamic_debug can be on through out the tests and removed only when the last device gets removed. 

LGTM

abdhaleegit (approved):
LGTM

PR #2029, opened 2021-03-24
uperf test followed by nping

vaishnavibhat:
nping with tcp packets is done after uperf test.

Signed-off-by: Vaishnavi Bhat <vaishnavi@linux.vnet.ibm.com>

### DIFF
--- a/io/net/uperf_test.py
+++ b/io/net/uperf_test.py
@@ -69,9 +69,11 @@ class Uperf(Test):
             self.cancel("failed connecting to peer")
         smm = SoftwareManager()
         detected_distro = distro.detect()
-        pkgs = ["gcc", "autoconf", "perl", "m4", "git-core", "automake"]
+        pkgs = ["gcc", "gcc-c++", "autoconf", "perl", "m4", "git-core", "automake"]
         if detected_distro.name == "Ubuntu":
             pkgs.extend(["libsctp1", "libsctp-dev", "lksctp-tools"])
+        elif detected_distro.name == "rhel":
+            pkgs.extend(["nmap"])
         else:
             pkgs.extend(["lksctp-tools", "lksctp-tools-devel"])
         for pkg in pkgs:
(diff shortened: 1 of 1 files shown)


io/net/uperf_test.py
```diff
@@ -136,6 +167,7 @@ def test(self):
         cmd = "h=%s proto=tcp ./src/uperf -m manual/throughput.xml -a" \
             % self.peer_ip
         result = process.run(cmd, shell=True, ignore_status=True)
+        nping_result = self.nping()
```
smruti77, line 170: @vaishnavibhat  what i feel this line should be added after throughput compare, here is no point to keep anyways nping is checking after uperf test is complete.

smruti77 (changes requested):

io/net/uperf_test.py
```diff
@@ -150,6 +182,11 @@ def test(self):
                               ", Throughput Actual value - %s "
                               % ((tput*100)/speed, self.expected_tp,
                                  str(tput)+'Mb/sec'))
+        for line in nping_result.stdout.decode("utf-8").splitlines():
+            if 'Raw packets' in line:
+                lost = int(line.split("|")[2].split(" ")[2])*10
+                if lost > 60:
+                    self.fail("FAIL: Ping fails after iperf test")
```
smruti77, line 189: here just a normal change, "uperf" in place of iperf .

io/net/uperf_test.py
vaishnavibhat: @smruti77 Want to check if nping sees a packet loss during performance tests and base the results on both uperf out put and nping value. So doing a nping check immediately.

smruti77: ok,but can you add this after ```exit_status``` check to make sure command run fine or uperf not failed. sometime we see uperf failed,so that time no use to check nping.

narasimhan-v: @vaishnavibhat Can you take care of this comment please ?

Naresh-ibm:
@vaishnavibhat can you please take a look on the changes requested?

abdhaleegit (changes requested):
@vaishnavibhat Once we hav these 2 changes.. I can merge

1. the nmap pkg shall be appended to exiting list of packages
2. move  nping_result = self.nping()  call to just above this line
for line in nping_result.stdout.decode("utf-8").splitlines():

Naresh-ibm (approved):

Naresh-ibm:
@vaishnavibhat I think you need to re-base this branch. please check.

abdhaleegit (approved):
LGTM

PR #2038, opened 2021-04-15
producer-consumer.py: Parse results, store in json

harish-24:
Patch adds result parsing and stores it to json file

### DIFF
--- a/cpu/producer_consumer.py
+++ b/cpu/producer_consumer.py
@@ -14,8 +14,10 @@
 #
 # Based on code by Gautham Shenoy <ego@linux.vnet.ibm.com>
 
+import json
 import os
 import platform
+import re
 
 from avocado import Test
 from avocado.utils import process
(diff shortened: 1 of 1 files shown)


harish-24:
```
# cat /home/harish/avocado-fvt-wrapper/results/job-2021-04-15T20.20-18916c9/test-results/1-cpu_producer_consumer.py_Producer_Consumer.test_run-consumer-stats-default-consumer-duration-load-perf-precompute-producer-random-statistics-30bf/time_log.json 
{"iterations": "50856", "iter_time": "43357", "access_time": "42"}
```

PraveenPenguin (commented):
@harish-24 looks good to this patch , one suggestion can we  use this Json out put to verify some performance matrix as example  it can be user defined input which we can match with json fields

harish-24:
> @harish-24 looks good to this patch , one suggestion can we use this Json out put to verify some performance matrix as example it can be user defined input which we can match with json fields

Sorry, dint get the point. can you elaborate?

harish-24:
> @harish-24 looks good to this patch , one suggestion can we use this Json out put to verify some performance matrix as example it can be user defined input which we can match with json fields

I kind of get your point. The whole purpose of storing as a human/machine readable format is to make use of that in  various follow up scripts. It is hard for a user to find a user-defined baseline metric considering the dynamic change in environment. Rather this can be run across different environments and compared against each other. And that again cannot be part of this standalone test.

PraveenPenguin:
> > @harish-24 looks good to this patch , one suggestion can we use this Json out put to verify some performance matrix as example it can be user defined input which we can match with json fields
> 
> I kind of get your point. The whole purpose of storing as a human/machine readable format is to make use of that in various follow up scripts. It is hard for a user to find a user-defined baseline metric considering the dynamic change in environment. Rather this can be run across different environments and compared against each other. And that again cannot be part of this standalone test.

@harish-24 thanks for info , yes make sense we can make this in

PraveenPenguin (approved):

PR #2050, opened 2021-05-02
Update skipIf condition check for em_quad_level_frequency_transition test case

sacsant:
em_quad_level_frequency_transition test case is supported only on PowerNV.
Adjust the skipIf() statement to reflect the same.

The patch also updates the error message to reflect missing cpufreq sysfs attribute.


sacsant:
With the fix:

(1/1) cpu/em_quad_level_frequency_transitions.py:freq_transitions.test: SKIP: This test is not supported on PowerVM
RESULTS    : PASS 0 | ERROR 0 | FAIL 0 | SKIP 1 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 0.15 s

PraveenPenguin (approved):
LGTM

harish-24 (approved):

cpu/em_quad_level_frequency_transitions.py
```diff
@@ -35,7 +35,7 @@ class freq_transitions(Test):
 
     :avocado: tags=cpu,power,privileged
     """
-    @skipIf(IS_POWER_NV, "This test only supported on Power9  platform")
+    @skipIf(not IS_POWER_NV, "This test is not supported on PowerVM")
```
harish-24, line 38: Nitpick: this code runs for each variant, better to stick to adding `not` at the top so that no processing takes place here.

sacsant: Thanks. Should I send an updated patch? Or can this be merged as is?

harish-24: you can send an updated patch.

PR #2052, opened 2021-05-03
ipistorm test case fixup for get_family error

sacsant:
ipistorm test case fails to run with following error:
DBLIPIStrom.test: ERROR: 'module' object has no attribute 'get_family'

Use distro.detect().arch statement to check for ppc architecture
Also fix pycodestyle warning against line 91.


sacsant:
This is a incomplete fix. Will send a V2 shortly.

PR #2056, opened 2021-05-04
Update hwinfo test case to allow it to run on RHEL variant

sacsant:
The test case fails on RHEL with following error
Hwinfo.test_list: CANCEL: Hwinfo not supported on RHEL (0.00 s)

Given that hwinfo tool is available with recent releases of Red Hat
Enterprise Linux, remove the check for rhel from the test case.


sacsant:
will send a new pull request

PR #2069, opened 2021-05-18
New test for strace a Linux command line utility.

sacsant:
New test for strace a Linux command line utility.
Strace is a command line utility in Linux used for diagnosis and debugging.

This test downloads the source code, compiles it and finally
invokes the built-in tests.

### DIFF
--- a/toolchain/strace.py
+++ b/toolchain/strace.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2021 IBM
+# Author: Sachin Sant <sachinp@linux.vnet.ibm.com>
+
+import os
+import re
+
+from avocado import Test
+from avocado.utils import build, process, git
+from avocado.utils.software_manager import SoftwareManager
+
+
+class Strace(Test):
+
+    """
+    Strace is a diagnostic and debugging command line utility for Linux
+    :avocado: tags=os,testsuite
+    """
+
+    def setUp(self):
+        """
+        Build strace
+
+        Source:
+        http://github.com/strace/strace.git
+        """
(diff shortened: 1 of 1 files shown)


sacsant:
Run log:

# avocado run toolchain/strace.py 
JOB ID     : e12aa2cad5906180ac7c6e389ce14fe41a2a494d
JOB LOG    : /home/sachin/results/job-2021-05-18T07.44-e12aa2c/job.log
 (1/1) toolchain/strace.py:Strace.test: PASS (308.02 s)
JOB HTML   : /home/sachin/results/job-2021-05-18T07.44-e12aa2c/results.html
JOB TIME   : 327.36 s

toolchain/strace.py
```diff
@@ -0,0 +1,68 @@
...
+
+        Source:
+        http://github.com/strace/strace.git
+        """
+        smm = SoftwareManager()
+        packages = ['make', 'gcc', 'autoconf', 'automake']
+        for package in packages:
+            if not smm.check_installed(package) and not smm.install(package):
+                self.cancel(' %s is needed for the test to be run' % package)
+        git.get_repo('https://github.com/strace/strace.git',
```
PraveenPenguin, line 44: can we use zip file as in that case we can leverage  `fetch_asset`  API ?

toolchain/strace.py
```diff
@@ -0,0 +1,68 @@
...
+    def setUp(self):
+        """
+        Build strace
+
+        Source:
+        http://github.com/strace/strace.git
+        """
+        smm = SoftwareManager()
+        packages = ['make', 'gcc', 'autoconf', 'automake']
+        for package in packages:
```
PraveenPenguin, line 41: can use package list directory here like 

`for package in ['make', 'gcc', 'autoconf', 'automake']:`  as it will help for garbage collector management well

PraveenPenguin (changes requested):
@sacsant thanks for test , just some minor Review comment , One help can we extend test case to use  `stress` rpm from native Distro

toolchain/strace.py
sacsant: strace project hosted on github does not allow to download a zip/tarball, so I think this is not an option.

strace's program version is maintained using git-version-gen script, therefore, building strace depends on "git describe" (a full git clone is needed for this)or, alternatively, a .tarball-version file.  Neither github nor gitlab can
provide the latter, they also do not allow to inject a .tarball-version file during export archiving.

toolchain/strace.py
sacsant: will make this change.

harish-24 (approved):

PR #2072, opened 2021-05-24
[net] bonding.py: use public ip address when setting up LACP

cforno12:
When running the setup test, the test attempts to create an ssh session using
the private ip address assigned to one of the interfaces. This attempt fails
because the interfaces are configured for LACP and they can only communicate
once they are bonded.


cforno12:
[job.log (test_setup)](https://github.com/avocado-framework-tests/avocado-misc-tests/files/6534929/job.1.log)

io/net/bonding.py
```diff
@@ -116,16 +116,46 @@ def setUp(self):
...
+                for _ in range(5):
+                    if self.session.connect():
+                        connect = True
+                        self.log.info("Was able to connect to peer.")
+                        break
+                    time.sleep(5)
+                if not connect:
+                    self.cancel("failed connecting to peer")
+            else:
+                self.cancel("failed connecting to peer")
```
cforno12, line 149: @abdhaleegit added this code block to fix the failing of connecting to peer. Let me know what you think!

abdhaleegit: looks good

abdhaleegit (approved):

PR #2073, opened 2021-05-24
[net] virt-net: logport assigned out of scope

cforno12-zz:
Test should cancel if it cannot find a logical port of a backing device. Often
the test fails with a message like "logport was assigned before declaration".
This gives it a more descriptive message.


abdhaleegit (approved):
LGTM

PR #2081, opened 2021-06-08
virt-net: improve vNIC DLPAR test

cforno12-zz:
Currently, the interface is only tested after X amount of DLPAR remove/add
operations. This fix adds an ip addr, brings the link up, and pings the peer
with the interface after every dlpar add. On remove, it checks to see if the
interface is still present.

Additionally, the interface is removed via chhwres on the HMC instead
of using the drmgr on the LPAR as drmgr is used for developer use only and is not used in production.

### DIFF
--- a/io/net/virt-net/network_virtualization.py
+++ b/io/net/virt-net/network_virtualization.py
@@ -424,39 +424,54 @@ class NetworkVirtualization(Test):
 
     def test_vnic_dlpar(self):
         '''
-        Perform vNIC device hot add and hot remove using drmgr command
+        Perform vNIC device hot add and hot remove
         '''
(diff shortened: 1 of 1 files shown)


cforno12:
This has been tested on multiple vNICs without backind devices and a single vNIC with one backing device.

Naresh-ibm:
@cforno12 are you still looking to get this into upstream?
if yes, can you please attach the latest log and code if any changes done so far?
@vaishnavibhat @abdhaleegit ^^

abdhaleegit:
already covered here https://github.com/avocado-framework-tests/avocado-misc-tests/pull/2309

PR #2100, opened 2021-07-09
ndctl: Fix umount error with the test

kvaneesh:
With the latest avocado framework (commit "utils/partition.py: Fix fetching the
exact mount point") umount expect self.mountpoint to be set. Update
the test to track mount points so that an unmount is successful.

Signed-off-by: Aneesh Kumar K.V <aneesh.kumar@linux.ibm.com>


PraveenPenguin (approved):
LGTM

PR #2105, opened 2021-07-14
Added support for retrying setting the MTU on vnic

otisdog8:
When changing the mtu on vnic, the interface will reset which
can be solved by repeatedly setting the MTU and waiting. I use
run_command instead of process because that will run the command
over ssh also.

Signed-off-by: Jacob Root <otis@otisroot.com>
[job_retrymtu.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/6812606/job_retrymtu.log)

### DIFF
--- a/io/net/tcpdump.py
+++ b/io/net/tcpdump.py
@@ -19,6 +19,7 @@ Tcpdump Test.
 """
 
 import os
+from time import sleep
 import netifaces
 from avocado import Test
 from avocado.utils import process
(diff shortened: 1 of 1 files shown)


cforno12:
Should we add this retry mtu feature to NetworkInterfaces in utils? Looking at it again we would have to add this to every networking test...

vaishnavibhat:
Will the wait_for() function not suffice for resetting mtu ? we use wait_for() for other functions is_link_up.

cforno12:
> Will the wait_for() function not suffice for resetting mtu ? we use wait_for() for other functions is_link_up.

@vaishnavibhat It will not suffice as setting the mtu on vnic will sometimes fail with `Command 'ip link set env8 mtu 1500' failed.
│ │  stderr: b'RTNETLINK answers: Device or resource busy\n'` 

Thus, the test should retry setting the mtu upto 5 times. Should this be another parameter in `set_mtu` in utils? So,

```
def set_mtu(self, mtu, timeout=30, retry=1):
    ...
```

PraveenPenguin:
@otisdog8 as seems this PR has conflict , if you still pursue this please raise new , closing this one

PR #2140, opened 2021-10-04
[io/net] vlan.py: update the vlan test to use the latest utils

cforno12-zz:
Refactored and reorganized vlan test to use the latest avocado utils.

Signed-off-by: Cris Forno <fornosoccer12@gmail.com>

### DIFF
--- a/io/net/vlan_test.py
+++ b/io/net/vlan_test.py
@@ -15,12 +15,14 @@
 # VLAN Testcase
 
 import time
-import paramiko
 
 from avocado import Test
 from avocado.utils import process
-from avocado.utils.ssh import Session
+from avocado.utils.network.hosts import LocalHost, RemoteHost
+from avocado.utils.network.interfaces import NetworkInterface
 from avocado.utils.process import CmdError
+
+import paramiko
 
 
 class VlanTest(Test):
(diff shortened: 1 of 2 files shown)


io/net/vlan_test.py
```diff
@@ -175,32 +129,29 @@ def test_vlan_1_2230(self):
...
         time.sleep(5)
-        if self.ping_check_host(self.host_intf, self.ip_dic[self.peer_intf]):
-            self.fail("Ping test failed for vlan 1 & 2230 in host")
-        self.log.info("Ping test passed for vlan 1 % 2230 in host")
-        if self.ping_check_peer(self.peer_intf, self.ip_dic[self.host_intf]):
-            self.fail("Ping test failed for vlan 1 & 2230 in peer")
-        self.log.info("Ping test passed for vlan 1 % 2230 in peer")
+
+        self.host_interface.ping_check(peer_ip=self.peer_ip, count=5)
+        self.peer_interface.ping_check(peer_ip=self.peer_ip, count=5)
```
Naresh-ibm, line 134: is this something your doing self ping check?
Or ping from peer to host? if that is the case it should be,
self.peer_interface.ping_check(peer_ip=<host_ip>, count=5)

abdhaleegit (changes requested):
@cforno12  Job logs please ?

io/net/vlan_test.py
Naresh-ibm: @vaishnavibhat can you please re-check on this and attached the latest logs?
So that we can ask for next steps.

abdhaleegit:
No activity from long time, Please reopen if you want to proceed further, also the PR does not work on latest switches as the cli is different, so not good idea to hardcode for one switch

PR #2153, opened 2021-11-05
On the Feiteng 2500 machine, the normal page size is 64K, huge_ The page size is 512M, which solves Nr_ Too many pages will lead to insufficient memory

liuyafei8:
On the Feiteng 2500 machine, the normal page size is 64K, huge_ The page size is 512M, which solves Nr_ Too many pages will lead to insufficient memory


PraveenPenguin:
@liuyafei8 i suggest please use the YAML file to overwrite these values as having value just can be for reference

PraveenPenguin:
@liuyafei8 is above suggestion helped you or still need to pursue this PR

PR #2203, opened 2021-12-15
fixing the timing issue in numa_test.py

Naresh-ibm:
The retunr value after a command execution taking longer time
to get reflect the exact value. it varies each time the
command is executed. So added wait.for() to resolve this
issue and the script works fine now


Naresh-ibm:
issue is inconsistence, will re-push with exact fix. closing as of now

PR #2223, opened 2022-01-26
fix run numa  testcases faild 

tongxintest:
hello, when I test numa is problem
getpagesize() return is byte,but file /sys/devices/system/node/node1/meminfo is kb
so testcase is faild

when test_thp_compare and test_movepages togehter 
file /sys/devices/system/node/node1/meminfo  is not lockd

my modify result 
企业微信截图_16431667985086

### DIFF
--- a/memory/numa_test.py
+++ b/memory/numa_test.py
@@ -18,6 +18,7 @@
 
 import os
 import shutil
+import time
 
 from avocado import Test
 from avocado import skipIf
@@ -116,6 +117,7 @@ class NumaTest(Test):
         """
         Test PFN's before and after offlining
         """
+        time.sleep(200)
         self.nr_pages = self.params.get(
             'nr_pages', default=100)
         os.chdir(self.teststmpdir)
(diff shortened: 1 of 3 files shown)


tongxintest:
@kvaneesh

memory/numa_test.py
```diff
@@ -116,6 +117,7 @@ def test_thp_compare(self):
         """
         Test PFN's before and after offlining
         """
+        time.sleep(200)
```
PraveenPenguin, line 120: looks ugly to add sleep in test case , can we not avoid this

PraveenPenguin (commented):
over all patch looks good , have one suggestion please have a look

memory/numa_test.py
tongxintest: ok  ,i need learn how to locked file, wait me

Naresh-ibm: @tongxintest any update here?

PR #2272, opened 2022-04-12
Fix for em_cpuhotplug.py Test failing issue.

SamirMulani:
In present implementention for CPU online and offline function in utils/cpu.py file below cases are present,
online function implementation:

Case-1: In case the CPU is already online, it will return 1 if it is.
Case-2: Upon checking CPU status if the CPU is offline, we write 1 into the "online" file. Then, after writing into the "online" file, we check if the CPU is online and if it is, we return 0.
-> Above two different cases we are returning 1 & 0 when our CPU is online.

Note: Logically when the CPU is online it should return 1.
Added code changes to utils/cpu.py in order to return 1 when CPU is
online.
To support above code changes added support in em_cpuhotplug.py.


SamirMulani:
This code changes depend on **#https://github.com/avocado-framework/avocado/pull/5317**

PraveenPenguin (approved):

PR #2277, opened 2022-04-19
Fix for perf_c2c.py test broken issue.

SamirMulani:
The current implementation of process.run does not create perf.data file after execution of perf record command. The perf.data file is the input to the perf report command. The test failed because the current implementation failed to create the perf.data file.

To execute the perf record command, added code changes to perf_c2c.py.

### DIFF
--- a/perf/perf_c2c.py
+++ b/perf/perf_c2c.py
@@ -15,6 +15,8 @@
 # Author: Nageswara R Sastry <rnsastry@linux.vnet.ibm.com>
 
 import os
+import shlex
+import subprocess
 import platform
 import tempfile
 from avocado import Test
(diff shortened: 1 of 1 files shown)


perf/perf_c2c.py
```diff
@@ -88,20 +89,41 @@ def run_cmd(self, cmd):
         except process.CmdError as details:
             self.fail("Command %s failed: %s" % (cmd, details))
 
+    def run_record_cmd(self, cmd):
+        """
+        Run record command
+        """
+        try:
+            cmd_output = subprocess.Popen(cmd,
```
PraveenPenguin, line 97: ah why not use `process` API from avocado,as I know this is a data drainer issue in avocado process API , can  you please have look

Naresh-ibm:
@SamirMulani any update here.

PraveenPenguin:
closing this as as it won't pursue more

PR #2283, opened 2022-04-26
Patch address False alarm

PraveenPenguin:
As recetly in SLES15-SP4 test case is failing and that is due to actual build failure
but while running it shows Unsupported OS and that give False positive as test case was
working earlier


Geet2:
Thanks for rectifying the error message @PraveenPenguin

PraveenPenguin:
@Geet2 if looks good changes to you can please ACK this

Geet2 (approved):

PR #2285, opened 2022-04-28
Add support to use upstream fsprogs while testing

OjaswinM:
Modify fs/xfstests.py to pull and make fsprogs like e2fsprogs,
xfsprogs and btrfsprogs from upstream. This feature is useful
when testing upstream kernels and especially to be able to
correctly test btrfs subpage support, since the distro provided 
btrfs versions don't have the latest features/patches needed.

To test for upstream, we can use the run_type option in the
yaml. We also need to specify the upstream URLs of the fsprogs
we want to install, eg:

...
upstream:
  run_type: 'upstream'		# Test with upstream fsprogs
  xfsprogs_url: 'https://git.kernel.org/pub/scm/fs/xfs/xfsprogs-dev.git'
  e2fsprogs_url: 'https://git.kernel.org/pub/scm/fs/ext2/e2fsprogs.git'
  btrfsprogs_url: 'https://git.kernel.org/pub/scm/linux/kernel/git/kdave/btrfs-progs.git'
distro:
  run_type: 'distro'		# Test with distro provided fsprogs

Signed-off-by: Ojaswin Mujoo <ojaswin@linux.ibm.com>

### DIFF
--- a/fs/xfstests.py
+++ b/fs/xfstests.py
@@ -173,7 +173,10 @@ class Xfstests(Test):
             packages.extend(
                 ['xfslibs-dev', 'uuid-dev', 'libuuid1',
                  'libattr1-dev', 'libacl1-dev', 'libgdbm-dev',
-                 'uuid-runtime', 'libaio-dev', 'fio', 'dbench'])
+                 'uuid-runtime', 'libaio-dev', 'fio', 'dbench',
+                 'gettext', 'libinih-dev', 'liburcu-dev', 'libblkid-dev',
+                 'liblzo2-dev', 'zlib1g-dev', 'e2fslibs-dev', 'asciidoc',
+                 'xmlto', 'libzstd-dev', 'libudev-dev'])
             if self.detected_distro.version in ['14']:
                 packages.extend(['libtool'])
             elif self.detected_distro.version in ['18', '20']:
@@ -189,12 +192,17 @@ class Xfstests(Test):
             packages.extend(['acl', 'bc', 'dump', 'indent', 'libtool', 'lvm2',
                              'xfsdump', 'psmisc', 'sed', 'libacl-devel',
                              'libattr-devel', 'libaio-devel', 'libuuid-devel',
-                             'openssl-devel', 'xfsprogs-devel'])
+                             'openssl-devel', 'xfsprogs-devel', 'gettext',
+                             'libblkid-devel', 'lzo-devel', 'zlib-devel',
+                             'e2fsprogs-devel', 'asciidoc', 'xmlto',
+                             'libzstd-devel', 'systemd-devel', 'meson',
+                             'gcc-c++'])
 
             if self.detected_distro.name == 'SuSE':
-                packages.extend(['libbtrfs-devel', 'libcap-progs'])
+                packages.extend(['libbtrfs-devel', 'libcap-progs',
+                                'liburcu-devel', 'libinih-devel'])
             else:
-                packages.extend(['btrfs-progs-devel'])
+                packages.extend(['btrfs-progs-devel', 'userspace-rcu-devel'])
 
             packages_remove = ['indent', 'btrfs-progs-devel']
             if self.detected_distro.name == 'rhel' and\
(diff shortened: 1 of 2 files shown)


OjaswinM:
Before this patch (or when not using the `upstream` option), we can see that the mkfs.btrfs version used is an older version that is supplied by the distro (using `yum install ...` or `apt-get install`)

```
INFO | Running 'yes | mkfs -t btrfs  -s 4096 -f /dev/loop0'
DEBUG| [stdout] btrfs-progs v5.4.1
```

When using `upstream` option, btrfs is now pulled from the `btrfsprogs_url` supplied, and is the latest:

```
INFO | Running 'yes | mkfs -t btrfs  -s 4096 -f /dev/loop0'
DEBUG| [stdout] btrfs-progs v5.17
```

fs/xfstests.py
```diff
@@ -214,13 +222,72 @@ def setUp(self):
...
+
+                # Build xfs progs
+                xfsprogs_dir = self.teststmpdir + "/xfsprogs"
+                os.makedirs(xfsprogs_dir)
+                xfsprogs_url = self.params.get('xfsprogs_url')
+                git.get_repo(xfsprogs_url, destination_dir=xfsprogs_dir)
+                os.chdir(xfsprogs_dir)
+                #process.run("./configure", verbose=True)
+                process.run("make", verbose=True)
+                process.run("make install", verbose=True)
```
PraveenPenguin, line 270: please use avocado utils for same 
```

build.make(xfsprogs_dir)
build.make(xfsprogs_dir,extra_args='install')
```

fs/xfstests.py
```diff
@@ -214,13 +222,72 @@ def setUp(self):
...
+                # Build btrfs progs
+                btrfsprogs_dir = self.teststmpdir + "/btrfsprogs"
+                os.makedirs(btrfsprogs_dir)
+                btrfsprogs_url = self.params.get('btrfsprogs_url')
+                git.get_repo(btrfsprogs_url, destination_dir=btrfsprogs_dir)
+                os.chdir(btrfsprogs_dir)
+                process.run("./autogen.sh", verbose=True)
+                process.run("./configure --prefix=/usr/local/ --disable-documentation", verbose=True)
+                process.run("make", verbose=True)
+                process.run("make install", verbose=True)
```
PraveenPenguin, line 282: same applies here

fs/xfstests.py
```diff
@@ -214,13 +222,72 @@ def setUp(self):
...
+
+        # Check versions
+        e2fs_ver = process.system_output("mkfs.ext4 -V", ignore_status=True, shell=True).decode("utf-8")
+        print(e2fs_ver)
+        xfs_ver = process.system_output("mkfs.xfs -V", ignore_status=True, shell=True).decode("utf-8")
+        print(xfs_ver)
+        btrfs_ver = process.system_output("mkfs.btrfs -V", ignore_status=True, shell=True).decode("utf-8")
+        print(btrfs_ver)
```
PraveenPenguin, line 290: please push logging  in self.log.info() instead of print

PraveenPenguin (changes requested):
@OjaswinM thanks for the patch,  have a couple of  suggestions please have a look

OjaswinM:
Hi Praveen, thank you for the review! I have modified the code accordingly.

fs/xfstests.py
```diff
@@ -221,6 +229,73 @@ def setUp(self):
...
+                build.make(btrfsprogs_dir)
+                build.make(btrfsprogs_dir, extra_args='install')
+
+        # Check versions of fsprogs
+        e2fs_ver = process.system_output("mkfs.ext4 -V", ignore_status=True, shell=True).decode("utf-8")
+        self.log.info(e2fs_ver)
+        xfs_ver = process.system_output("mkfs.xfs -V", ignore_status=True, shell=True).decode("utf-8")
+        self.log.info(xfs_ver)
+        btrfs_ver = process.system_output("mkfs.btrfs -V", ignore_status=True, shell=True).decode("utf-8")
+        self.log.info(btrfs_ver)
```
sacsant, line 298: Can you use self.fs_to_test and make the code generic (instead of repetitive) ? 
```
fs_ver = process.system_output("mkfs.%s -V" % self.fs_to_test, ignore_status=True, shell=True).decode("utf-8")
self.log.info(fs_ver)
```

sacsant: In fact the same logic can be applied to the above code block used to compile the <fs>progs code.

OjaswinM: Yes that sounds better Sachin, I've addressed this.

Thanks

fs/xfstests.py
```diff
@@ -31,7 +31,7 @@
 from avocado.utils import process, build, git, distro, partition
 from avocado.utils import disk, data_structures, pmem
 from avocado.utils import genio
-from avocado.utils.software_manager import SoftwareManager
+from avocado.utils.software_manager.manager import SoftwareManager
```
PraveenPenguin, line 34: ah @OjaswinM I missed this, as still, our repository is runner compliant and due to this change our CI/CR break, as we are planning to have this change across the repository soon , can you please revert this

OjaswinM: Ahh alright, done :)

PraveenPenguin (approved):
LGTM

fs/xfstests.py
```diff
@@ -221,6 +229,72 @@ def setUp(self):
...
+                    os.makedirs(e2fsprogs_dir)
+                e2fsprogs_url = self.params.get('e2fsprogs_url')
+                git.get_repo(e2fsprogs_url, destination_dir=e2fsprogs_dir)
+                e2fsprogs_build_dir = os.path.join(e2fsprogs_dir, 'build')
+                if not os.path.exists(e2fsprogs_build_dir):
+                    os.makedirs(e2fsprogs_build_dir)
+                os.chdir(e2fsprogs_build_dir)
+                process.run("../configure", verbose=True)
+                build.make(e2fsprogs_build_dir)
+                build.make(e2fsprogs_build_dir, extra_args='install')
```
sacsant, line 248: Would this command install the binaries in /usr/ or /usr/local ?

fs/xfstests.py
```diff
@@ -221,6 +229,72 @@ def setUp(self):
...
+                                      libini. Please check the logs.')
+                # Build xfs progs
+                xfsprogs_dir = os.path.join(self.teststmpdir, 'xfsprogs')
+                if not os.path.exists(xfsprogs_dir):
+                    os.makedirs(xfsprogs_dir)
+                xfsprogs_url = self.params.get('xfsprogs_url')
+                git.get_repo(xfsprogs_url, destination_dir=xfsprogs_dir)
+                os.chdir(xfsprogs_dir)
+                build.make(xfsprogs_dir)
+                build.make(xfsprogs_dir, extra_args='install')
```
sacsant, line 278: Same comment. Would this command install the binaries in /usr/ or /usr/local ?

fs/xfstests.py
```diff
@@ -221,6 +229,72 @@ def setUp(self):
...
+            if self.fs_to_test == "btrfs":
+                # Build btrfs progs
+                btrfsprogs_dir = os.path.join(self.teststmpdir, 'btrfsprogs')
+                if not os.path.exists(btrfsprogs_dir):
+                    os.makedirs(btrfsprogs_dir)
+                btrfsprogs_url = self.params.get('btrfsprogs_url')
+                git.get_repo(btrfsprogs_url, destination_dir=btrfsprogs_dir)
+                os.chdir(btrfsprogs_dir)
+                process.run("./autogen.sh", verbose=True)
+                process.run("./configure --prefix=/usr/local/ --disable-documentation", verbose=True)
```
sacsant, line 289: This would ensure all the binaries with make install would be copied to /usr/local, which is good.

fs/xfstests.py
```diff
@@ -221,6 +229,72 @@ def setUp(self):
...
+                build.make(btrfsprogs_dir)
+                build.make(btrfsprogs_dir, extra_args='install')
+
+        # Check versions of fsprogs
+        fsprogs_ver = process.system_output("mkfs.%s -V" % self.fs_to_test,
+                                            ignore_status=True,
+                                            shell=True).decode("utf-8")
+        self.log.info(fsprogs_ver)
 
         if process.system('which mkfs.%s' % self.fs_to_test,
```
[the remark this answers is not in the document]
OjaswinM: Hey Sachin, all the distros I've tested seem to have `/usr/local/bin` before `/usr/sbin` in their $PATH. In my tests, the script is correctly able to pick up the updated mkfs.btrfs from `/usr/local/bin`. Can you please provide the avocado run logs and $PATH value of the system you observe this behavior on. I can take a look.

fs/xfstests.py
OjaswinM: I noticed that this installs in `/sbin` (although docs suggested `/usr/local/bin` however I do agree that it would be better to explitly pass `--prefix=/usr/local` as `/usr/local/bin` seems to precede `/sbin` in `$PATH`. 

I'll do the same for xfsprogs as well.

fs/xfstests.py
OjaswinM: I'll set this explicitly to `/use/local/bin`. _See above comment._

fs/xfstests.py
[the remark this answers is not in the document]
sacsant: > Hey Sachin, all the distros I've tested seem to have `/usr/local/bin` before `/usr/sbin` in their $PATH. In my tests, the script is correctly able to pick up the updated mkfs.btrfs from `/usr/local/bin`. Can you please provide the avocado run logs and $PATH value of the system you observe this behavior on. I can take a look.

On SLES it's the other way around.
 
~ echo $PATH
/sbin:/usr/sbin:/usr/local/sbin:/root/bin:/usr/local/bin:/usr/bin:/bin
~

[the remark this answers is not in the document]
OjaswinM: Just to summarize the discussing we had on Slack regarding this:

Basically, there are times when the xfstests also call mkfs  and when it does so, they use the one in `$PATH` and I couldn't find a way to change this. So I think we will have to make sure the binaries are installed at right path instead of explicitly using `/usr/local/bin/mkfs`

So we agreed upon something like the follows:
```
if SLES:
    --prefix=/sbin
else RHEL:
    --prefix=/usr/local
```

I've updated this PR to reflect this change. The behavior is now something as follows:
- Install libs and other stuff in `/usr/local` for both RHEL and SuSE
- Install the binaries (mkfs etc) in `/usr/local/{bin,sbin}` for RHEL and `/sbin` for SuSE

fs/xfstests.py
```diff
@@ -221,6 +229,90 @@ def setUp(self):
         self.test_mnt = self.params.get('test_mnt', default='/mnt/test')
         self.disk_mnt = self.params.get('disk_mnt', default='/mnt/loop_device')
         self.fs_to_test = self.params.get('fs', default='ext4')
+        self.run_type = self.params.get('run_type', default='distro')
+
+        if self.run_type == 'upstream':
+            prefix = "/usr/local"
+            bin_prefix = "/usr/local/bin"
+            sbin_prefix = "/usr/local/bin"
```
sacsant, line 237: Did you actually mean /usr/local/bin and /usr/local/sbin ? If not then just one variable would suffice.

fs/xfstests.py
```diff
@@ -221,6 +229,90 @@ def setUp(self):
...
+
+            if self.detected_distro.name == 'SuSE':
+                # SuSE has /sbin at a higher priority than /usr/local/bin
+                # in $PATH, so install all the binaries in /sbin to make
+                # sure they are picked up correctly by xfstests.
+                #
+                # We still install in /usr/local but binaries are kept in
+                # /sbin
+                bin_prefix = "/sbin"
+                sbin_prefix = "/sbin"
```
sacsant, line 247: Same comment. Did you actually mean /bin and /sbin ? If not then just one variable would suffice.

fs/xfstests.py
```diff
@@ -221,6 +229,90 @@ def setUp(self):
...
+                if not os.path.exists(e2fsprogs_dir):
+                    os.makedirs(e2fsprogs_dir)
+                e2fsprogs_url = self.params.get('e2fsprogs_url')
+                git.get_repo(e2fsprogs_url, destination_dir=e2fsprogs_dir)
+                e2fsprogs_build_dir = os.path.join(e2fsprogs_dir, 'build')
+                if not os.path.exists(e2fsprogs_build_dir):
+                    os.makedirs(e2fsprogs_build_dir)
+                os.chdir(e2fsprogs_build_dir)
+                process.run("../configure --prefix=%s --bindir=%s --sbindir=%s"
+                            % (prefix, bin_prefix, sbin_prefix), verbose=True)
```
sacsant, line 261: If both bin_prefix and sbin_prefix are one and the same then a single variable can be used instead of 2

sacsant (changes requested):

fs/xfstests.py
OjaswinM: Right, I think we can get rid of the duplication. Will do. Thanks for pointing it out

sacsant (approved):
Thanks. Latest change looks good.

PR #2286, opened 2022-04-29
Fix for Distro: NDCTL Upstream Test Failing issue.

SamirMulani:
Removed implementations for upstream.patch and ppc.patch.

### DIFF
--- a/memory/ndctl_selftest.py
+++ b/memory/ndctl_selftest.py
@@ -86,16 +86,6 @@ class NdctlTest(Test):
 
         self.sourcedir = self.teststmpdir
         os.chdir(self.sourcedir)
-
-        # TODO: remove patches once merged upstream
-        upstream_patch = self.fetch_asset("upstream.patch", locations=[
-            "https://patchwork.kernel.org/series/177255/mbox/"], expire='7d')
-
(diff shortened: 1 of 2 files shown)


memory/ndctl_selftest.py
```diff
@@ -86,16 +86,6 @@ def setUp(self):
...
-
-        # TODO: remove patches once merged upstream
-        upstream_patch = self.fetch_asset("upstream.patch", locations=[
-            "https://patchwork.kernel.org/series/177255/mbox/"], expire='7d')
-
-        process.run('patch -p1 < %s' % upstream_patch, shell=True)
-        if detected_distro.arch in ["ppc64", "ppc64le"]:
-            process.run('patch -p1 < %s' %
-                        self.get_data('ppc.patch'), shell=True)
-
```
Naresh-ibm, line 98: @SamirMulani are we removing the upstream support here?

PraveenPenguin: no, it like patch already part for ndctl upstream

rohan-r-deshpande: Tried and tested the code, removing the above block of code snippet. 
Changes look good & we can safely go ahead for the merge.

rohan-r-deshpande (approved):

PR #2297, opened 2022-05-10
Style Fixes

PraveenPenguin:
Moved Print statement to logger as it will log statemnt in logger file


PraveenPenguin:
as this is just a style fix so merging this

PR #2305, opened 2022-05-17
Use distro as a default target for toolchain component tests

sacsant:
A number of tests under toolchain directory has support for
both distro and upstream as a target (as a YAML input).

Lot of these toolchain tests have a dependency on compilers and
libraries installed on the system where they are executed.

In order to avoid test errors due to mismatch in levels being
tested vs what is available on the system, switch to using
distro as the default target for all the applicable tests.

### DIFF
--- a/toolchain/binutils.py
+++ b/toolchain/binutils.py
@@ -58,7 +58,7 @@ class Binutils(Test):
             needed_deps.extend(['build-essential'])
         for pkg in needed_deps:
             self.check_install(pkg)
-        run_type = self.params.get("type", default="upstream")
+        run_type = self.params.get("type", default="distro")
         # Extract - binutils
         # Source: https://ftp.gnu.org/gnu/binutils/binutils-2.26.tar.bz2
         if run_type == "upstream":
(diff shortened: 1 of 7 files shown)


PraveenPenguin:
@sacsant couple of suggestions/observation

As we are defaulting distro as a target so it will force the user to configure src repo,  sometimes not an easy case, even as having YAML support we can override the target, so thinking we should have default as upstream version. 
As having using upstream we see less user error and aligned with the upstream (use case) effort

sacsant:
> @sacsant couple of suggestions/observation
> 
> As we are defaulting distro as a target so it will force the user to configure src repo, sometimes not an easy case, even as having YAML support we can override the target, so thinking we should have default as upstream version. As having using upstream we see less user error and aligned with the upstream (use case) effort

Right agree. To use distro as a target corresponding source code software repository has to be setup. At the same time I have come across two user error reports(glibc and valgrind) when upstream was used as a target. Agree that these options can be overridden via YAML, new users of this test may or may not do that. 

May be we don't need a big hammer approach. How about we change default target to distro only for valgrind and glibc and leave the remaining tests untouched. 
In longer run we should also enhance glibc and valgrind tests (for upstream as a target) to make it more robust. For glibc with upstream as a target we will also need to pull in latest gcc and use the newly built gcc to compile glibc code.

sacsant:
On second thought let's fix the individual test cases for doth distro and upstream as a target so that irrespective of the target they work correctly.

@PraveenPenguin thanks for your inputs. Will close this PR.

PR #2308, opened 2022-05-20
Fix SoftwareManager import

dsilakov:
software_manager/__init__.py is empty now (https://github.com/avocado-framework/avocado/commit/1b8360c7ffb0e6874eb4f9cdb9c593337482bbd4) so imports don't work.

Let's change them to match examples:

https://github.com/avocado-framework/avocado/commit/1b8360c7ffb0e6874eb4f9cdb9c593337482bbd4#diff-759aaf7d53119997ec7909dc4710a0ab2191335bcebd30347b0e03bc2f4c61b8

### DIFF
--- a/cpu/cpuidle-latency.py
+++ b/cpu/cpuidle-latency.py
@@ -21,7 +21,7 @@ import shutil
 from avocado import Test
 from avocado.utils import process
 from avocado.utils import build, distro, git
-from avocado.utils.software_manager import SoftwareManager
+from avocado.utils.software_manager.manager import SoftwareManager
 
 
 class Cpuidle_latency(Test):
--- a/cpu/cpupower_monitor.py
+++ b/cpu/cpupower_monitor.py
@@ -22,7 +22,7 @@ from avocado import skipIf
 from avocado.utils import archive
 from avocado.utils import build, distro
 from avocado.utils import process, cpu
-from avocado.utils.software_manager import SoftwareManager
+from avocado.utils.software_manager.manager import SoftwareManager
 
 
 class CpupowerMonitor(Test):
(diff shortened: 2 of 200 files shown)


PraveenPenguin:
@dsilakov thanks for the patch, we can wait for this patch as still https://github.com/avocado-framework-tests/avocado-misc-tests not complaint to run against the latest avocado , we are in plan around to moving to latest , we can pursue this change that time

PraveenPenguin:
> > @dsilakov thanks for the patch, we can wait for this patch as still https://github.com/avocado-framework-tests/avocado-misc-tests not complaint to run against the latest avocado , we are in plan around to moving to latest , we can pursue this change that time
> 
> Thanks, got it. Actually I am trying to run the tests with the latest avocado on Ubuntu now:) Besides software manager issue, the largest source of failures is inconsistent setUp / tearDown functions in some tests for unsupported OS - when in setUp one just calls test cancel but then in tearDown tries to use variable which is undefined because we cancelled the test before initializing the variable. I wonder if this is something known or can I also send PRs to fix such cases?

@dsilakov ah, not sure as I did not come across this, seems it is well-needed Fix those instances if you planning to send PR for the same please do so that is great addition , if looking my help, please  let me know

PraveenPenguin:
@dsilakov sorry for late , one request can you please resolve conflict so we should able to merge this

abdhaleegit:
@dsilakov a similar one is also raised https://github.com/avocado-framework-tests/avocado-misc-tests/pull/2347

dsilakov:
I'd suggest to merge #2347 since it's up to date and has no conflicts.

PR #2318, opened 2022-05-25
Fix unclear failure messages when running tests on unsupported OS (Ubuntu)

dsilakov:
Some tests are only intended for RHEL / SLES, and produce confusing failure messagescaused by the fact that some variables are not initialized in setUp but used in the tearDown (so we get an missing attribute Exception instead of a clear statement about unsupported system).

Proposed solution is like the following:

- In memory/pmem* and ndctl, let's check if we have corresponding attributes set before trying to use them
- In selinux-tests, we can initiate the problematic variable to None in the beginning and then use it only if it is not empty
- ras/lparstat is just missing explicit test cancelation for the OS other than rhel/suse.

Signed-off-by: Denis Silakov <dsilakov@gmail.com>

### DIFF
--- a/memory/ndctl.py
+++ b/memory/ndctl.py
@@ -1008,15 +1008,16 @@ class NdctlTest(Test):
 
     @avocado.fail_on(pmem.PMemException)
     def tearDown(self):
-        if self.part:
+        if hasattr(self, 'part') and self.part:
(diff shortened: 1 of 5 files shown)


abdhaleegit (approved):
LGTM

PR #2319, opened 2022-05-25
Revert "Revert "Including the set-channel options in the ethool test""

abdhaleegit:
Reverts avocado-framework-tests/avocado-misc-tests#2290


PraveenPenguin:
@abdhaleegit I say we go with the old codebase as the still library also has changes like function name, so we can have fresh PR as CR/CI failing with library change

PR #2331, opened 2022-05-30
Cancel libhugetlbfs tests on RHEL9*

spoorthys1303:
libhugetlbfs package is not available in RHEL9.* hence the test case will be canceled

Signed-off-by: spoorthy <spoorts2@in.ibm.com>

### DIFF
--- a/memory/libhugetlbfs.py
+++ b/memory/libhugetlbfs.py
@@ -47,6 +47,8 @@ class LibHugetlbfs(Test):
         # Check for basic utilities
         smm = SoftwareManager()
         detected_distro = distro.detect()
+        if (detected_distro.name == 'rhel' and detected_distro.version == '9'):
+            self.cancel("libhugetlbfs is not available RHEL 9.x onwards")
         deps = ['gcc', 'make', 'patch']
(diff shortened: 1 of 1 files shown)


memory/libhugetlbfs.py
```diff
@@ -47,6 +47,11 @@ def setUp(self):
         # Check for basic utilities
         smm = SoftwareManager()
         detected_distro = distro.detect()
+        
+        if detected_distro.name == 'rhel':
+            if (detected_distro.version == '9' and detected_distro.release == '1'):
```
PraveenPenguin, line 52: this test case not support RHEL9-- onwards even  may you can use `skip` global decorate

PraveenPenguin (changes requested):
@spoorthys1303, thanks for patch ,  make script pylint compliant and have valid signed off

memory/libhugetlbfs.py
spoorthys1303: These scenarios are cancelled in earlier test cases rather than Skipping, hence we can continue with canceling the tests

PraveenPenguin (approved):

PR #2334, opened 2022-06-06
Resolving cleanup issue for fiotest.py

Naresh-ibm:
Similar to tiobench other tests under disk dir we have claenup
issue i,e the mounted dir and disk or fs created on disk or sraids
are not getting cleanup up properly after each varient ran.
so this code will resolve this issue


Naresh-ibm:
changes are not there in this file

PR #2351, opened 2022-06-22
enhancing ltp_fsstress and fixing pre_clean issue

Naresh-ibm:
fixing the pre_cleanup issue and did some enhancement
to the existing script for better usage of script

### DIFF
--- a/io/disk/ltp_fsstress.py
+++ b/io/disk/ltp_fsstress.py
@@ -13,6 +13,7 @@
 #
 # Copyright: 2016 IBM
 # Author: Vaishnavi Bhat <vaishnavi@linux.vnet.ibm.com>
+#       : Naresh Bannoth <nbannoth@linux.vnet.ibm.com>
 #
 # https://github.com/autotest/autotest-client-tests/tree/master/ltp
 
@@ -20,12 +21,13 @@
 LTP fsstress test
 """
 
-
 import os
+import time
 from avocado import Test
 from avocado.utils import build
 from avocado.utils import disk
 from avocado.utils import lv_utils
+from avocado.utils import wait
 from avocado.utils import softwareraid
 from avocado.utils import distro
 from avocado.utils import process, archive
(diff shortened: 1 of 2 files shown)


io/disk/ltp_fsstress.py
```diff
@@ -139,38 +173,166 @@ def create_fs(self, l_disk, mountpoint, fstype):
...
+            process.system(cmd, shell=True, ignore_status=True)
+            if disk.fs_exists(l_disk):
+                return False
+            return True
+
+        def is_disk_unmounted():
+            cmd = "umount %s" % l_disk
+            cmd1 = 'umount /dev/mapper/avocado_vg-avocado_lv'
+            process.system(cmd, shell=True, ignore_status=True)
+            process.system(cmd1, shell=True, ignore_status=True)
```
abdhaleegit, line 269: Why duplicate commands ? cant we pass disk value and a single process.system(cmd, can take care of both ?

abdhaleegit: param initialization should be outside the function always

io/disk/ltp_fsstress.py
```diff
@@ -139,38 +173,166 @@ def create_fs(self, l_disk, mountpoint, fstype):
...
+                return False
+            return True
+
+        self.log.info("checking if disk is mounted.")
+        if disk.is_disk_mounted(l_disk):
+            self.log.info("%s is mounted, unmounting it ....", l_disk)
+            if wait.wait_for(is_disk_unmounted, timeout=10):
+                self.log.info("%s unmounted successfully" % l_disk)
+            else:
+                self.err_mesg.append("%s unmount failed", l_disk)
```
abdhaleegit, line 287: log only if fails.  too much logging is not needed

io/disk/ltp_fsstress.py.data/ltp_fsstress.yaml
```diff
@@ -1,7 +1,8 @@
-disk:
-script: 'fsstress'
-dir: '/mnt/nfs'
-fsstress_loop: 1
+disk: ''
+dir: ''
+fsstress_loop:
+nval: '250'
+pval: '250'
```
abdhaleegit, line 5: Does readme explains the values to be given fsstress_loop:
nval: '250'
pval: '250'

abdhaleegit (changes requested):
@Naresh-ibm can we avoid too much loginfo.. I would suggested to not log if things pass.. log only when things fail.. this applies all cleanup code PRs

io/disk/ltp_fsstress.py
Naresh-ibm: will put them outside of the method.

io/disk/ltp_fsstress.py.data/ltp_fsstress.yaml
Naresh-ibm: we do not have the Readme file. will add one with all required info

PR #2360, opened 2022-06-29
Automated htx installation

rohan-r-deshpande:
- Provided support for picking up the latest available htx version
  for the respective distro.
- Added a function install_latest_htx_rpm() for the same.

Signed-off-by: Rohan Deshpande <rohan_d@linux.ibm.com>

### DIFF
--- a/generic/htx_test.py
+++ b/generic/htx_test.py
@@ -21,6 +21,8 @@ HTX Test
 import os
 import time
 import shutil
+import re
+
 from avocado import Test
 from avocado.utils.software_manager.manager import SoftwareManager
 from avocado.utils import build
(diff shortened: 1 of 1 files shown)


generic/htx_test.py
```diff
@@ -21,8 +21,10 @@
 import os
 import time
 import shutil
+import re
+
 from avocado import Test
-from avocado.utils.software_manager.manager import SoftwareManager
+from avocado.utils.software_manager import SoftwareManager
```
Naresh-ibm, line 27: Please clone the latest misc test and then do the modification.
.manager was recent change and should not be removed.

generic/htx_test.py
```diff
@@ -40,6 +42,25 @@ class HtxTest(Test):
     :params time_limit: how much time(hours) you want to run this stress.
     """
 
+    def install_latest_htx_rpm(self):
```
Naresh-ibm, line 45: Add the docstring tho the method

Naresh-ibm (changes requested):

generic/htx_test.py
rohan-r-deshpande: Done

PraveenPenguin:
@Naresh-ibm can you please check this

generic/htx_test.py
```diff
@@ -40,6 +42,29 @@ class HtxTest(Test):
...
+            htx_rpm for htx_rpm in matching_htx_versions
+            if distro_pattern in htx_rpm]
+        distro_specific_htx_versions.sort(reverse=True)
+        self.latest_htx_rpm = distro_specific_htx_versions[0]
+
+        if process.system('rpm -ivh --nodeps %s%s '
+                          '--force' % (self.rpm_link, self.latest_htx_rpm),
+                          shell=True, ignore_status=True):
+            self.cancel("Installion of rpm failed")
+
```
Naresh-ibm, line 67: code wise it looks good to me.
@PraveenPenguin @rohan-r-deshpande can we move this part to common code.
we will be needing this in most of our testing like, for block devices, NIC/vNIC
So I feel, it should be like just pass distro version and then it should be able to install the respective HTX RPM.
if not exact to the util part, can we have this in locally in avocado-misc-tests itself?


@abdhaleegit @vaishnavibhat ^^

PraveenPenguin: @Naresh-ibm yes  good idea, if @rohan-r-deshpande you not started  we can pursue this as an  enhancement  as this logic we need tp propagate in other area as well

PraveenPenguin: @rohan-r-deshpande ah ok then we can pursue this as enhancement, @Naresh-ibm is this looks ok to you?

PR #2366, opened 2022-06-30
removing the test "remove_all_path" from multipath.py

Naresh-ibm:
removing the test "remove_all_path" from multipath.py as it is
bringing down the multipath services. this can be enhanced in future
if we have some work_around for this. Hence commented the particular
test to get skipped from running for now.


abdhaleegit:
this must be fixed locally

PR #2380, opened 2022-07-26
perf/perf_pmu: Remove PMU register test

sacsant:
test_PMU_register subtest searches for a particular text string
in dmesg log. This may not work if the dmesg log has been cleared
by previous test.

There are other subtests related to PMU sysfs check as well as
kernel config check this test seem to be unnecessary.

Given the error prone nature remove this subtest.

### DIFF
--- a/perf/perf_pmu.py
+++ b/perf/perf_pmu.py
@@ -15,7 +15,6 @@
 
 import os
 import glob
-import re
 from avocado import Test
 from avocado.utils import cpu, dmesg, genio, linux_modules, process
 from avocado import skipIf, skipUnless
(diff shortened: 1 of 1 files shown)


sacsant:
without the fix:

(1/9) /home/tests/avocado-misc-tests/perf/perf_pmu.py:PerfBasic.test_PMU_register: STARTED
 (1/9) /home/tests/avocado-misc-tests/perf/perf_pmu.py:PerfBasic.test_PMU_register:  FAIL: No registered PMUs found. (0.48 s)

After fix since this test is removed there is no corresponding output.

PraveenPenguin (approved):

PR #2394, opened 2022-08-16
netperf: provide the tar file name instead full path

abdhaleegit:
when untaring netperf suite on peer, make sure you give
the tarball name instead a wrong path, this was a typo fix
from previous PR


Naresh-ibm (approved):
LGTM

PR #2438, opened 2022-10-10
memory: Add 5-level page table tests

kalpanashtty:
* Test will detect CPU feature i.e la57.
      * Check for kernel support
      * Run series of 5-level page tests from "pg-table_tests.git" that covers
            - heap, mmap, shmat tests.

### DIFF
--- a/memory/page_table.py
+++ b/memory/page_table.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2022 AMD 
+# Author: Kalpana Shetty <kalpana.shetty@amd.com>
+#
+
+import os
+from avocado import Test
+from avocado import skipUnless
+from avocado.utils import cpu, git, genio, build, process, linux_modules
+from avocado.utils.software_manager.manager import SoftwareManager
+
+
+class PageTable(Test):
+    """
+    Tests 5-level, 4-level page table depending on the supported platform.
+          * Test will detect CPU feature, 5 Level page table.
+          * Check for kernel support
+          * Run series of 5-level page tests from "pg-table_tests.git" that covers
+	        - heap, mmap, shmat tests.
+    :avocado: tags=memory
+    """
+
+    @skipUnless('x86_64' in cpu.get_arch(),
+                "This test runs on x86-64 platform.\
+        If 5-level page table supported on other platform then\
+        this condition can be removed")
+    def setUp(self):
+        '''
+        Install pre-requisites packages. 
+        Setup pa-table_tests.git
+        '''
+        smm = SoftwareManager()
+        for package in ['gcc', 'make']:
+            if not smm.check_installed(package) and not smm.install(package):
+                self.cancel('%s is needed for the test to be run' % package)
+
+        self.url = self.params.get('url', default=None)
+        git.get_repo(self.url, destination_dir=self.teststmpdir)
+        os.chdir(self.teststmpdir)
+        build.make(self.teststmpdir)
+
+    def test_detect_5lvl(self):
+        '''
+        Detect 5-Level page table CPU feature
+        '''
+        cpu_info = genio.read_file("/proc/cpuinfo")
+        if 'la57' in cpu_info:
+           self.log.info("Detected 5-Level page table cpu support")
+        else:
+           self.fail("5-Level page table - Unsupported platform")
+
+    def test_kernel(self):
+        '''
+        Check 5-Level page table support at kernel.
+        '''
+        cfg_param = "CONFIG_X86_5LEVEL"
+        result = linux_modules.check_kernel_config(cfg_param)
+        if result == linux_modules.ModuleConfig.NOT_SET:
+           self.fail("%s is not set in the kernel." % cfg_param)
+        else:
+           self.log.info("Detected 5-Level page table config - CONFIG_X86_5LEVEL set in the kernel")
+
+    def test_pg_table_tests(self):
+        '''
+        Run series of 5-level page tests from "pg-table_tests.git" that covers
+                - heap, mmap, shmat tests.
+        '''
+        output = process.run('./run-tests')
+        err_msg = []
+        for line in output.stdout.decode('utf-8').splitlines():
+            if "failed" in line:
+                err_msg.append(line)
+        if err_msg:
+            self.fail("Page Table tests failed: %s" % err_msg)
+        else:
+            self.log.info("Page Table tests passed.")
--- a/memory/page_table.py.data/page_table.yaml
+++ b/memory/page_table.py.data/page_table.yaml
@@ -1,0 +1,1 @@
+url: 'https://github.com/sanskriti-s/pg-table_tests.git'


memory/page_table.py
```diff
@@ -0,0 +1,75 @@
...
+        If 5-level page table supported on other platform then\
+        this condition can be removed")
+    def setUp(self):
+        smm = SoftwareManager()
+        for package in ['gcc', 'make']:
+            if not smm.check_installed(package) and not smm.install(package):
+                self.cancel('%s is needed for the test to be run' % package)
+
+        self.url = self.params.get(
+            'url', default="https://github.com/sanskriti-s/pg-table_tests.git")
```
Naresh-ibm, line 46: @kalpanashtty moving this to yaml will help more when we need to change the url.
So suggest to have this in yaml. ok to hard_coded in yaml if needed.

memory/page_table.py
```diff
@@ -0,0 +1,75 @@
...
+           self.cancel("%s is not set in the kernel." % cfg_param)
+        else:
+           self.log.info("Detected 5-Level page table config - CONFIG_X86_5LEVEL set in the kernel")
+
+    def test_pg_table_tests(self):
+        output = process.run('./run-tests')
+        failed_tests = []
+        for line in output.stdout.decode('utf-8').splitlines():
+            if "failed" in line:
+                failed_tests.append(line)
```
Naresh-ibm, line 71: looks like your not adding the failed test name here, so check if we can add the name else having the more meaningful variable name like err = [] or err_msg = [] etc.

memory/page_table.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def test_kernel(self):
+        cfg_param = "CONFIG_X86_5LEVEL"
+        result = linux_modules.check_kernel_config(cfg_param)
+        if result == linux_modules.ModuleConfig.NOT_SET:
+           self.cancel("%s is not set in the kernel." % cfg_param)
+        else:
+           self.log.info("Detected 5-Level page table config - CONFIG_X86_5LEVEL set in the kernel")
+
+    def test_pg_table_tests(self):
```
Naresh-ibm, line 66: @kalpanashtty add doc_strings for all the methods.

Naresh-ibm (changes requested):

memory/page_table.py
kalpanashtty: Actually, I forgot to create yaml folder... will do.

memory/page_table.py
kalpanashtty: Done

memory/page_table.py
kalpanashtty: Added comments.

memory/page_table.py
kalpanashtty: Done

memory/page_table.py
kalpanashtty: Done

memory/page_table.py
Naresh-ibm: adding the hard_coded value for default url is not required here as we are already doing it in yaml. 
so please remove this and add `default=None`

memory/page_table.py
```diff
@@ -0,0 +1,89 @@
...
+        git.get_repo(self.url, destination_dir=self.teststmpdir)
+        os.chdir(self.teststmpdir)
+        build.make(self.teststmpdir)
+
+    def test_detect_5lvl(self):
+        '''
+        Detect 5-Level page table CPU feature
+        '''
+        cpu_info = genio.read_file("/proc/cpuinfo")
+        if 'la57' in cpu_info:
```
Naresh-ibm, line 60: I feel this should be exact match and may give the false True here if this is the part of some other string.
so please re-check. I am ok if you think this is a unique string.

Naresh-ibm (changes requested):

memory/page_table.py
kalpanashtty: Yes, it's a unique string in "/proc/cpuinfo" for 5-Level page table support.
Ref: https://lwn.net/Articles/716324/

Naresh-ibm:
@kalpanashtty please add the job/debug.log for reference along with run logs.

memory/page_table.py
kalpanashtty: > adding the hard_coded value for default url is not required here as we are already doing it in yaml. so please remove this and add `default=None`

Done.

kalpanashtty:
> @kalpanashtty please add the job/debug.log for reference along with run logs.

Attached...
[job.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/9751763/job.log)

[1-page_table.py_PageTable.test_detect_5lvl_run-6e47_debug.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/9751765/1-page_table.py_PageTable.test_detect_5lvl_run-6e47_debug.log)

[2-page_table.py_PageTable.test_kernel_run-6e47_debug.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/9751768/2-page_table.py_PageTable.test_kernel_run-6e47_debug.log)

[3-page_table.py_PageTable.test_pg_table_tests_run-6e47_debug.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/9751769/3-page_table.py_PageTable.test_pg_table_tests_run-6e47_debug.log)

kalpanashtty:
root@xxxxx:/home/xxxxx/avocado-misc-tests/memory# avocado run --test-runner runner page_table.py -m /home/xxxxx/avocado-misc-tests/memory/page_table.py.data/page_table.yaml
JOB ID     : 5ed873c544465f89a518bc602b67d7da02a2c5f9
JOB LOG    : /home/xxxx/tests/results/job-2022-10-11T04.53-5ed873c/job.log
 (1/3) page_table.py:PageTable.test_detect_5lvl;run-6e47: PASS (2.19 s)
 (2/3) page_table.py:PageTable.test_kernel;run-6e47: PASS (1.72 s)
 (3/3) page_table.py:PageTable.test_pg_table_tests;run-6e47: PASS (2.77 s)
RESULTS    : PASS 3 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /home/xxxxx/tests/results/job-2022-10-11T04.53-5ed873c/results.html
JOB TIME   : 7.68 s

Naresh-ibm:
Thanks @kalpanashtty 
LGTM

kalpanashtty:
@narasimhan-v  @PraveenPenguin:  @Naresh-ibm  has done the review, so, whenever you have time please review it.

narasimhan-v (commented):
@kalpanashtty Just a minor question / suggestion. Will `test_detect_5lvl()` run if `test_detect_5lvl()` and `test_detect_5lvl()` cancels ?
If they don't run in that case, we need to move the 2 functions into `setUp()`

kalpanashtty:
> @kalpanashtty Just a minor question / suggestion. Will `test_detect_5lvl()` run if `test_detect_5lvl()` and `test_detect_5lvl()` cancels ? If they don't run in that case, we need to move the 2 functions into `setUp()`

Yes; it would run other tests... primarily, adding manual tests as different test cases so that if the cpu feature "la57" is present it would pass. So, thinking we can make the test fail instead of cancel might give more meaning.

But somehow adding it as part of "setUp()" will not differentiate different test cases. Also, in the future, we will add additional test cases and that would repeat 'setUp' for each of those tests.
So, in my opinion, if we can make the test fail would make more sense instead of adding 1st 2 cases as part of 'setUp'.

kalpanashtty:
> > @kalpanashtty Just a minor question / suggestion. Will `test_detect_5lvl()` run if `test_detect_5lvl()` and `test_detect_5lvl()` cancels ? If they don't run in that case, we need to move the 2 functions into `setUp()`
> 
> Yes; it would run other tests... primarily, adding manual tests as different test cases so that if the cpu feature "la57" is present it would pass. So, thinking we can make the test fail instead of cancel might give more meaning.
> 
> But somehow adding it as part of "setUp()" will not differentiate different test cases. Also, in the future, we will add additional test cases and that would repeat 'setUp' for each of those tests. So, in my opinion, if we can make the test fail would make more sense instead of adding 1st 2 cases as part of 'setUp'.

Discussed with Narasimhan and we will make test fail if we don't see system is capable of 5-level feature instead cancel the tests. So pushed the changes for review.

PraveenPenguin (approved):

narasimhan-v (approved):

Naresh-ibm (approved):
@PraveenPenguin @narasimhan-v Thanks for reviews
@kalpanashtty Thanks for accommodating all the changes suggested.

PR #2441, opened 2022-10-14
memory/transparent_hugepages*.py: Add THP support check based on sysf…

kalpanashtty:
…s "transparent_hugepage" path

  Current test code uses PAGE SIZE to determine whether THP is supported or not. This doesn't work when we have
  platform with different PAGESIZE (like x86 with 4K).

  The better way is to check, if the sysfs "transparent_hugepage" path exists; i.e check for the path
   "/sys/kernel/mm/transparent_hugepage" and skip the test if path doesn't exists.

### DIFF
--- a/memory/transparent_hugepages.py
+++ b/memory/transparent_hugepages.py
@@ -24,7 +24,7 @@ from avocado.core import data_dir
 from avocado.utils.partition import Partition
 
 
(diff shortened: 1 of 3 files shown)


PraveenPenguin (approved):
LGTM

PR #2461, opened 2022-11-02
SRIOV: check MAC addr mismatch with OS

abdhaleegit:
Added a check to make sure the device mac addr
is not changed across hmc and Linux OS

### DIFF
--- a/io/net/sriov_device_test.py
+++ b/io/net/sriov_device_test.py
@@ -159,6 +159,8 @@ class NetworkSriovDevice(Test):
             if not self.list_device(mac):
                 self.fail("failed to list logical device after add operation")
             device = self.find_device(mac)
+            if not device:
+                self.fail("MAC address differs in linux")
             networkinterface = NetworkInterface(device, self.local)
             networkinterface.add_ipaddr(ipaddr, netmask)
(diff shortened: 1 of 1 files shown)


io/net/sriov_device_test.py
```diff
@@ -158,6 +158,8 @@ def test_add_logical_device(self):
             self.device_add_remove(slot, port, mac, '', 'add')
             if not self.list_device(mac):
                 self.fail("failed to list logical device after add operation")
+            if not self.find_device(mac):
+                self.fail("MAC address differs in linux")
```
PraveenPenguin, line 162: why two  time calling the function it can be performance overhead, can this be used like 

```
device = self.find_device(mac)
if not device:
  self.fail("MAC address differs in linux")
  
  ```

PraveenPenguin (changes requested):

io/net/sriov_device_test.py
abdhaleegit: Done

Naresh-ibm (approved):
@abdhaleegit LGTM

PraveenPenguin (approved):

PR #2476, opened 2022-11-16
Removing the switch credetails details from yaml for port_bounce

Naresh-ibm:
removing the access details of switch from yaml for security reason.


abdhaleegit (approved):
LGTM

PR #2478, opened 2022-11-17
EEH for block devies are update. Added validation to check multipath after EEH

maramsmurthy:
Updated EEH loop to trigger EEH for correct number of times as mentioned in yaml.
Added validation for multipath devices.

Signed-off-by: Maram Srimannarayana Murthy <Maram.Srimannarayana.Murthy@ibm.com>

### DIFF
--- a/io/pci/EEH.py
+++ b/io/pci/EEH.py
@@ -20,13 +20,14 @@ This scripts basic EEH tests on all PCI device
 """
 
 import time
+import re
 from avocado import Test
 from avocado.utils import process
 from avocado.utils import pci
 from avocado.utils import genio
 from avocado.utils import distro
 from avocado.utils import dmesg
-from avocado.utils.software_manager import SoftwareManager
+from avocado.utils.software_manager.manager import SoftwareManager
 
 EEH_HIT = 0
 EEH_MISS = 1
@@ -63,7 +64,22 @@ class EEH(Test):
         eeh_enable_file = "/sys/kernel/debug/powerpc/eeh_enable"
         if '0x1' not in genio.read_file(eeh_enable_file).strip():
             self.cancel("EEH is not enabled, please enable via FSP")
-        self.max_freeze = self.params.get('max_freeze', default=1)
+        self.max_freeze = self.params.get('max_freeze')
+        self.log.info('Reading max supported freeze from file eeh_max_freezes')
+        self.max_supported_freeze = int(process.system_output(
+            'cat /sys/kernel/debug/powerpc/eeh_max_freezes'))
+        if self.max_freeze is None:
+            self.max_freeze = self.max_supported_freeze
+            dmesg_op = process.system_output('dmesg',
+                                             ignore_status=True,
+                                             shell=True).decode("utf-8")
+            if 'times in the last hour' in dmesg_op:
+                eeh_already_hit = re.findall(r"This PCI device has failed "
+                                             "([0-9]+) times in the last hour and will be "
+                                             "permanently disabled after ([0-9]+) failures", dmesg_op)
+                self.log.info('max_freeze value is adjusted as %s eeh hits'
+                              ' already observed', eeh_already_hit[-1][0])
+                self.max_freeze -= int(eeh_already_hit[-1][0])
         self.pci_device = self.params.get('pci_device', default="")
         self.add_cmd = self.params.get('additional_command', default='')
         if not self.pci_device:
@@ -110,6 +126,14 @@ class EEH(Test):
                 self.log.info("Running error inject on pe %s function %s",
                               self.pci_device, func)
                 if num_of_miss < 5:
+                    if self.pci_class_name == 'scsi_host':
+                        paths_before_eeh = self.get_paths(self.pci_device)
+                        host_state_before_eeh = pci.get_port_state(
+                            self.pci_device)
+                        self.log.info("Paths before EEH: %s",
+                                      paths_before_eeh)
+                        self.log.info("hosts state before EEH: %s",
+                                      host_state_before_eeh)
                     if self.is_baremetal():
                         return_code = self.basic_eeh(func, '', '', '', '', '')
                     else:
@@ -134,6 +158,22 @@ class EEH(Test):
                                 break
                             else:
                                 self.log.info("PE recovered successfully")
+                    time.sleep(10)
+                    if (self.pci_class_name == 'scsi_host' and
+                            self.max_freeze >= num_of_hit):
+                        paths_after_eeh = self.get_paths(self.pci_device)
+                        host_state_after_eeh = pci.get_port_state(
+                            self.pci_device)
+                        self.log.info("Paths after EEH: %s",
+                                      paths_after_eeh)
+                        self.log.info("hosts state after EEH: %s",
+                                      host_state_after_eeh)
+                        if paths_before_eeh != paths_after_eeh:
+                            self.fail("All Paths are failed to recover"
+                                      " after EEH")
+                        if host_state_before_eeh != host_state_after_eeh:
+                            self.fail("Hosts state differ after EEH")
+
                 else:
                     self.log.warning("EEH inject failed for 5 times with\
                                function %s" % func)
@@ -261,3 +301,11 @@ class EEH(Test):
         if 'PowerNV' in genio.read_file("/proc/cpuinfo").strip():
             return True
         return False
+
+    def get_paths(self, fc_host):
+        '''
+        returns the list of paths coressponding to the given fc_host
+        '''
+        fc_host = fc_host.split('.')[0]
+        paths = pci.get_disks_in_pci_address(fc_host)
+        return paths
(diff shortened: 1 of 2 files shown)


abdhaleegit (commented):
@maramsmurthy in case of multipath device, should we not check for path state like  is it Active /faulty ?

io/pci/EEH.py
```diff
@@ -105,11 +106,19 @@ def test_eeh_basic_pe(self):
         enter_loop = True
         num_of_miss = 0
         num_of_hit = 0
-        while num_of_hit <= self.max_freeze:
+        while num_of_hit < self.max_freeze:
```
abdhaleegit, line 109: so we are injecting less than the max_freeze, we should still keep eeh more than max_freeze as last scenario @Naresh-ibm any suggestions here ?

Naresh-ibm, line 109: @maramsmurthy it should be `<=` only because the driver fails only on nth time eeh trigger, if logic is with `<` the it will chek for only n-1 time and driver is expected to re-cover all these time.
so keep the original one

io/pci/EEH.py
```diff
@@ -105,11 +106,19 @@ def test_eeh_basic_pe(self):
...
                               self.pci_device, func)
                 if num_of_miss < 5:
+                    if self.pci_class_name == 'scsi_host':
+                        paths_before_eeh = self.get_paths(self.pci_device)
+                        host_state_before_eeh = self.get_host_state(
+                                self.pci_device)
+                        self.log.info("Paths before EEH: %s", 
+                                      paths_before_eeh)
+                        self.log.info("hosts state before EEH: %s", 
+                                      host_state_before_eeh)
```
Naresh-ibm, line 121: I think this has to collected in the setup itself.
why after missing 5 times. move this to setup.
like if 
```


if self.pci_class_name == 'scsi_host':
                        paths_before_eeh = self.get_paths(self.pci_device)
                        host_state_before_eeh = self.get_host_state(
                                self.pci_device)
                        self.log.info("Paths before EEH: %s", 
                                      paths_before_eeh)
                        self.log.info("hosts state before EEH: %s", 
                                      host_state_before_eeh)


```

io/pci/EEH.py
```diff
@@ -134,6 +143,21 @@ def test_eeh_basic_pe(self):
...
+                    time.sleep(10)
+                    if self.pci_class_name == 'scsi_host':
+                        paths_after_eeh = self.get_paths(self.pci_device)
+                        host_state_after_eeh = self.get_host_state(
+                                self.pci_device)
+                        self.log.info("Paths after EEH: %s",
+                                      paths_after_eeh)
+                        self.log.info("hosts state after EEH: %s",
+                                      host_state_after_eeh)
+                        if paths_before_eeh != paths_after_eeh:
```
Naresh-ibm, line 155: I don't think this verification will yield the right validation.
always check the count and status of each file.
number of paths before EEH for that pci_address should be eaqual to number of paths after eeh trigger.
and the stats of each paths before and after should be active.
only in one case paths will not be in active state i,e on nth time eeh trigger. so you have to include that corner case too here

io/pci/EEH.py
```diff
@@ -261,3 +285,31 @@ def is_baremetal():
...
+        host_state = []
+        split_pattern = r':|\.'
+        pci_devices = ':'.join(re.split(split_pattern, pci_devices)[0:3])
+        cmd = 'ls -l /sys/class/fc_host | grep -i %s' % pci_devices
+        for line in process.getoutput(cmd).split('\n'):
+            if "/%s" % pci_devices in line:
+                host_state.append(process.getoutput(
+                'cat /sys/class/fc_host/%s/port_state' % line.split("/")[-1]))
+        return host_state
+
```
Naresh-ibm, line 315: check if we have common utils for these methods.

Naresh-ibm (changes requested):

io/pci/EEH.py
```diff
@@ -261,3 +285,31 @@ def is_baremetal():
...
             return True
         return False
+
+    def get_paths(self, fc_host):
+        '''
+        returns the list of paths coressponding to the given fc_host
+        '''
+        paths = []
+        split_pattern = r':|\.'
+        fc_host = ':'.join(re.split(split_pattern, fc_host)[0:3])
```
abdhaleegit, line 295: are you trying to split 
ls -l /sys/block/ | grep -i host0
lrwxrwxrwx. 1 root root 0 Nov 17 00:17 sda -> ../devices/vio/300000da/host0/rport-0:0-0/target0:0:0/0:0:0:0/block/sda
lrwxrwxrwx. 1 root root 0 Nov 17 00:17 sdb -> ../devices/vio/300000da/host0/rport-0:0-1/target0:0:1/0:0:1:0/block/sdb

just wanted to understand why this lines 
        split_pattern = r':|\.'
        fc_host = ':'.join(re.split(split_pattern, fc_host)[0:3])

when you already passing fc_host  as host0

io/pci/EEH.py
maramsmurthy: @Naresh-ibm I hope we can read this parameter from OS and increment the value by 1 directly instead of passing it in yaml. Intention of this test is full filled only after exceeding supporting number of EEH hits. 
Can I make the changes as suggested above. If so I will add it in setup method itself.

io/pci/EEH.py
maramsmurthy: Added it after every EEH hit as we need to validate paths after every EEH Hit.

io/pci/EEH.py
Naresh-ibm: @abdhaleegit  yes, driver will recover till max_freez_value and goes down on next eeh trigger.
means if we give max_free value as 5 then driver has to recover till 5th time and it will fails to recover on 6th time. hence we have to validate the driver recover till max time value and we have trigger the injection for max+1th time. so the condition for validation to be like below
`while num_of_hit <= self.max_freeze:`

Naresh-ibm: @maramsmurthy this is a mandatory yaml parameter. if any one wants to go for eeh stress, like more than 100 times and see if the driver is recovering, they can always change the /sys dir parameter value to 100 and run this test case. That is what test is doing. it will read the max value from yaml and set it to that partiular value and trigger those many times.

io/pci/EEH.py
maramsmurthy: @abdhaleegit This split I performed on PCI Address(Eg: 001a:01:00.0).
In input file we are passing only one PCI address. When a PCI device reset happens, all the function device will also get reset. So to capture information from other function, I did split on '001a:01:00.0' and forming a new string '001a:01:00', It's used to capture all the disks which are coming from this PCI device.

io/pci/EEH.py
maramsmurthy: Sure. @abdhaleegit and @Naresh-ibm you are correct. I would like to make changes as follows
Read max freeze value from OS if no value is not passed in yaml.
Please let me know your suggestion on this.

io/pci/EEH.py
maramsmurthy: Yeah! There is no method in common utils. I am planning to move it to under utils as it can be used with mulitple devices.

io/pci/EEH.py
maramsmurthy: @Naresh-ibm Array comparison is doing same thing as you said. I will include corner case as suggested.

abdhaleegit:
@maramsmurthy logs for FC, multipath, NVME, nvmf, IPR

io/pci/EEH.py
```diff
@@ -261,3 +285,31 @@ def is_baremetal():
...
+                paths.append(line.split("/")[-1])
+        return paths
+
+    def get_host_state(self, pci_devices):
+        '''
+        returns state of fc hosts corresponding to the given pci devices
+        '''
+        host_state = []
+        split_pattern = r':|\.'
+        pci_devices = ':'.join(re.split(split_pattern, pci_devices)[0:3])
```
abdhaleegit, line 308: pci_devices.split('.')[0] would work instead of re ?

io/pci/EEH.py
PraveenPenguin: even this as well

PraveenPenguin (changes requested):

maramsmurthy:
Adding Logs
FC EEH: fc_eeh.zip
[ipr_eeh.zip](https://github.com/avocado-framework-tests/avocado-misc-tests/files/10082392/ipr_eeh.zip)

[fc_eeh.zip](https://github.com/avocado-framework-tests/avocado-misc-tests/files/10082375/fc_eeh.zip)

IPR EEH: ipr_eeh.zip

io/pci/EEH.py
maramsmurthy: move this definition to utils

maramsmurthy:
Closing this as there is problem with master header, Raised another pull request
https://github.com/avocado-framework-tests/avocado-misc-tests/pull/2492/checks

PR #2481, opened 2022-11-17
vnic: Get backing devices for specific slot

nick-child-ibm:
Hello,

I was hitting issues running vnic add and remove tests (`test_add` and `test_remove`) when there was already an existing vnic device on the lpar. The error was from `update_backing_devices` : "local variable 'index' referenced before assignment".
The reason for the error was that the wrong list of backing devices was returned from `get_backing_devices`. Rather than returning a list of backing devices for the new vnic device, `get_backing_devices`  returned a list of backing devices for the older backing device. This is because `get_backing_devices` does not return backing_devices for a specific slot, it instead returns all the backing devices for all the vnic devices on the lpar.

Therefore, this PR adds an argument `slot` to `get_backing_devices` so that it can retrieve only the backing devices of a specific vnic device.

Thanks.

### DIFF
--- a/io/net/virt-net/network_virtualization.py
+++ b/io/net/virt-net/network_virtualization.py
@@ -755,14 +755,14 @@ class NetworkVirtualization(Test):
             self.fail("lshwres operation failed ")
         return output.stdout_text
 
-    def get_backing_devices(self):
+    def get_backing_devices(self, slot):
         '''
         Lists the Backing devices for a Network virtualized
         device
         '''
         cmd = 'lshwres -r virtualio -m %s --rsubtype vnic --level lpar \
-               --filter lpar_names=%s -F backing_devices' \
-               % (self.server, self.lpar)
+               --filter lpar_names=%s,slots=%s -F backing_devices' \
+               % (self.server, self.lpar, slot)
         try:
             output = self.session_hmc.cmd(cmd)
         except CmdError as details:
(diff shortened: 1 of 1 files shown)


abdhaleegit (approved):
LGTM, Thanks @nick-child-ibm

abdhaleegit:
@nick-child-ibm so you have the test run logs for above changes

nick-child-ibm:
Hi Abdul, thanks for the review.

Which tests would you like me to run and get logs for?

abdhaleegit:
> Hi Abdul, thanks for the review.
> 
> Which tests would you like me to run and get logs for?

if you can run a full test it will be good.. else one run logs for  
`network_virtualization.py:NetworkVirtualization.test_vnic_auto_failover` would be fine

nick-child-ibm:
Sure. The required configuration is one vnic device with an SRIOV backing device, and a second vnic device with an SRIOV backing device which uses a different physical adapter than the first vnic's backing device. I have attached the two logs before and after of `network_virtualization.py:NetworkVirtualization.test_vnic_auto_failover`. I made some edits to the log files to remove personal information.
[after_patch.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/10116988/after_patch.log)

[before_patch.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/10116986/before_patch.log)

PR #2490, opened 2022-11-29
cpu/xive.py: Added new tests

rohan-r-deshpande:
- test_verify_xive_cmdline: test that verifies whether xive is enabled through the kernel command line.
- test_verify_xive_config: test that verifies whether xive is enabled based on the config parameters

Signed-off-by: Rohan Deshpande <rohan_d@linux.ibm.com>

### DIFF
--- a/cpu/xive.py
+++ b/cpu/xive.py
@@ -19,6 +19,7 @@ import re
 
 from avocado import Test
 from avocado.utils import genio
+from avocado.utils import linux_modules
 
 
 class XIVE(Test):
@@ -48,6 +49,8 @@ class XIVE(Test):
             self.intr = 'XICS'
         else:
             self.fail("Unsupported Interrupt Mode")
+
+        self.no_config_parameter = []
 
     def test_intr_mode(self):
         if self.intr == 'XIVE':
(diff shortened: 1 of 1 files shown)


cpu/xive.py
```diff
@@ -90,3 +99,39 @@ def test_storeEOI(self):
...
+
+    def test_verify_xive_cmdline(self):
+        pattern = "xive=on"
+        retval = genio.is_pattern_in_file("/proc/cmdline", pattern)
+        if retval:
+            self.log.info("XIVE is enabled.")
+        else:
+            self.log.info("XIVE is not enabled.")
+
+    def test_verify_xive_config(self):
```
PraveenPenguin, line 111: use avocado module for same like 
```
 ret = linux_modules.check_kernel_config('CONFIG_PPC_XIVE')
 if ret == linux_modules.ModuleConfig.NOT_SET:
            self.cancel("XIVE is not enabled !!\n")
```

cpu/xive.py
```diff
@@ -29,6 +30,14 @@ class XIVE(Test):
     :avocado: tags=cpu,power
     '''
 
+    def read_line_with_matching_pattern(self, filename, pattern):
+        contents = []
+        with open(filename, "r") as file_obj:
+            for line in file_obj.readlines():
+                if pattern in line:
+                    contents.append(line.rstrip("\n"))
+        return contents
```
PraveenPenguin, line 39: not needed as if we are going to use avocado utility

PraveenPenguin (changes requested):

cpu/xive.py
rohan-r-deshpande: Done

PraveenPenguin (approved):

PR #2496, opened 2022-12-05
io/pci/dlpar: Improved error handling

sacsant:
lshwres command can return "No results were found" in response to io specific information query.
In such scenario the test case fails with following error:
  AttributeError: 'DlparPci' object has no attribute 'lpar_id'

The specific message is printed when hardware discovery is not done or should be refreshed.
Handle this case by adding an explicit check in the code.


abdhaleegit (approved):
LGTM

PR #2505, opened 2022-12-26
security/openssl-tests: Fix make and make test errors

sacsant:
While running openssl-tests with different supported targets
there were issues with the test code. With upstream as a target
few perl packages should be installed, else the test fails during
make. With distro as a target even though no tests were
executed overall test is still marked as pass.

This two patch series fixes both these issues.


sacsant:
Before the fix:
   (1/1) security/openssl-tests.py:OpenSSL.test: STARTED
   (1/1) security/openssl-tests.py:OpenSSL.test: ERROR: openssl-tests.py: 'make' command failed. (4.16 s)

After the fix:
   (1/1) security/openssl-tests.py:OpenSSL.test: STARTED
   (1/1) security/openssl-tests.py:OpenSSL.test: PASS (363.51 s)

sacsant:
With distro as a target
Before the fix:
   (1/1) security/openssl-tests.py:OpenSSL.test;run-run_type-distro-8228: STARTED
   (1/1) security/openssl-tests.py:OpenSSL.test;run-run_type-distro-8228: PASS (29.93 s)

Log file shows no tests were executed:
Running 'make test'
make: Nothing to be done for 'test'.

After the fix with following yaml file contents:
run_type: !mux
    distro:
        type: 'distro'
config_args: 'enable-ec_nistp_64_gcc_128 zlib enable-camellia enable-seed enable-rfc3779 enable-sctp enable-cms enable-md2 enable-rc5 enable-ktls  no-mdc2 no-ec2m no-sm2 no-sm4 enable-buildtest-c++ shared linux-ppc64le -Wl,--allow-multiple-definition'

(1/1) security/openssl-tests.py:OpenSSL.test;run-run_type-distro-8161: STARTED
(1/1) security/openssl-tests.py:OpenSSL.test;run-run_type-distro-8161: FAIL: openssl-tests.py: 'make check' failed. (317.39 s)

security/openssl-tests.py
```diff
@@ -58,6 +63,16 @@ def setUp(self):
             self.srcdir = smm.get_source(pkg_name, self.srcdir)
             if not self.srcdir:
                 self.fail("openssl source install failed.")
+        os.chdir(self.srcdir)
+        # Different Linux distro releases use different arguments
+        # for configure command. Provide an option to the user
+        # to specify the same.
```
Naresh-ibm, line 69: @sacsant if it is distro specific then can we hard code them like how we do for packages?

sacsant: Actually the comment is wrong. The configure parameters are tied to the openssl version and not necessarily distro version. For bistros using the default parameters to configure should work. In case a user wants to specify different/additional options they can do so using this newly introduced yaml parameter. This is particularly useful while using latest upstream code.

Thanks for the review comment. I have update the patch and reworded the above comment.

abdhaleegit (approved):
LGTM

PraveenPenguin:
@sacsant have we tried with recent avocado, as we discussed without this patch test will work

sacsant:
> @sacsant have we tried with recent avocado, as we discussed without this patch test will work

With v99.0 test runner second patch security/openssl-tests: False positive test pass is not required. I have updated the PR accordingly.

security/openssl-tests.py
```diff
@@ -31,13 +31,20 @@ def setUp(self):
         Install the basic packages to support openssl
         '''
         # Check for basic utilities
+        run_type = self.params.get('type', default='upstream')
```
PraveenPenguin, line 34: any reason to change move this line up ?

sacsant: I can use this variable to check for additional packages to be installed. Those additional packages are required for upstream only case.

sacsant:
Is this okay for merge?

PR #2512, opened 2022-12-29
security/pam: Fix runtime failures with distro as target

sacsant:
When distro is specified as a value for run_type the tests fail due to
missing source code compilation steps. Add the same.

Depending on pam version autogen.sh script may not be available.
Add a check for the same and skip the step if this script is not
part of source code.


abdhaleegit (approved):
LGTM

abdhaleegit:
@nasastry Please confirm

PR #2447, opened 2022-10-26
libhugetlbfs handling with header flag: RHEL9 

sasingh87:
Since RHEL9 onwards support for libhugetlbfs not there, set a flag HAVE_HUGETLB_HEADER to handle at appropriate places.
Aggregated from  PR -> Add rhel-9 distro check to cancel the test #2435
Signed-off-by: Saurabh Singh <Saurabh.Singh2@amd.com>

PR #2513, opened 2022-12-29
memory/migrate_pages: Handle migrate page tests(libhugetlbfs) for RHE…

kalpanashtty:
…L-9 distro

 Fix is added to handle libhugetlbfs dependency with migrate page tests for RHEL-9 or later rhel distro.

### DIFF
--- a/memory/migrate_pages.py
+++ b/memory/migrate_pages.py
@@ -50,16 +50,23 @@ class MigratePages(Test):
             self.cancel('Test requires two numa nodes to run.'
                         'Node list with memory: %s' % nodes)
 
-        dist = distro.detect()
+        self.dist = distro.detect()
         pkgs = ['gcc', 'make']
-        if dist.name in ["Ubuntu", 'debian']:
+        if self.dist.name in ["Ubuntu", 'debian']:
             pkgs.extend(['libpthread-stubs0-dev',
                          'libnuma-dev', 'libhugetlbfs-dev'])
-        elif dist.name in ["centos", "rhel", "fedora"]:
-            pkgs.extend(['numactl-devel', 'libhugetlbfs-devel'])
-        elif dist.name == "SuSE":
+        elif self.dist.name in ["centos", "rhel", "fedora"]:
+            pkgs.extend(['numactl-devel'])
+            if (self.dist.name == 'rhel' and self.dist.version >= '9'):
+                self.log.info("hugepage tests will not be run, due to\
+                               unavailability of libhugetlbfs packages in\
+                               RHEL-9 or later.")
+            else:
+                os.environ['HAVE_HUGETLB_HEADER'] = '1'
+                pkgs.extend(['libhugetlbfs-devel'])
+        elif self.dist.name == "SuSE":
             pkgs.extend(['libnuma-devel'])
-            if dist.version >= 15:
+            if self.dist.version >= 15:
                 pkgs.extend(['libhugetlbfs-devel'])
             else:
                 pkgs.extend(['libhugetlbfs-libhugetlb-devel'])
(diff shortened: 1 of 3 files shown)


memory/migrate_pages.py
```diff
@@ -50,16 +50,23 @@ def setUp(self):
...
                          'libnuma-dev', 'libhugetlbfs-dev'])
-        elif dist.name in ["centos", "rhel", "fedora"]:
-            pkgs.extend(['numactl-devel', 'libhugetlbfs-devel'])
-        elif dist.name == "SuSE":
+        elif self.dist.name in ["centos", "rhel", "fedora"]:
+            pkgs.extend(['numactl-devel'])
+            if (self.dist.name == 'rhel' and self.dist.version >= '9'):
+                self.log.info("hugepage tests will not be run, due to\
+                               unavailability of libhugetlbfs packages in\
+                               RHEL-9 or later.")
```
Naresh-ibm, line 63: @kalpanashtty I think we need to `cancel` the test instead of just printing message if it is not supported.

memory/migrate_pages.py
```diff
@@ -86,6 +93,8 @@ def test(self):
         cmd = './node_move_pages -n %s' % self.nr_chunks
 
         if self.hpage:
+            if (self.dist.name == 'rhel' and self.dist.version >= '9'):
+                self.cancel("Hugepage tests are cancelled on RHEL-9 and later.")
```
Naresh-ibm, line 97: @kalpanashtty I think this can be handled at line 61-63 itself.
and I feel loginfo is not required there.

memory/migrate_pages.py
kalpanashtty: Earlier we were thinking of this i.e canceling the tests if the distro is RHEL-9 or later.  But later based on Geetika and @sacsant  suggestion we took care of handling huge-page tests to be canceled and run THP tests alone. That's the reason we have not to cancel the tests here.

memory/migrate_pages.py
kalpanashtty: As replied earlier for huge-page tests we cancel and continue for TPH tests.

memory/migrate_pages.py
Naresh-ibm: then this message is not required here. any way test will get cancelled and the same message can be added where we are cancelling the test

kalpanashtty: This is just a log message with additional info, I thought no harm in giving additional data while it tries to install the packages and later handle canceling part. Is it hard that we need to remove this log info? or no harm in adding the info message?

Naresh-ibm: @kalpanashtty nothing harm in having extra info but to go standard of coding it is kind of unnecessary line of code. Thats it, nothing more than that.

PraveenPenguin (approved):
LGTM

memory/migrate_pages.py
kalpanashtty: @Naresh-ibm : Let me know if you need this message to be changed.  
This is primarily an old PR (https://github.com/avocado-framework-tests/avocado-misc-tests/pull/2447), where the submitter left the company so re-sending it. So, you can get the history of the issue from this PR.

memory/migrate_pages.py
sacsant: yeah, this message was added as an information only to let the user know that part of the test is not applicable for RHEL 9. Initial patch cancelled the entire test.

sacsant (commented):
LGTM

Naresh-ibm (approved):
@kalpanashtty Thanks for the Patch.
@PraveenPenguin @sacsant Thanks for the reviews

LGTM

PR #2514, opened 2022-12-30
ras/ras_lsvpd: Add sg3_utils package dependency

sacsant:
With upstream as a value for run_type, code compile fails if
sg3_utils-devel package is not installed.

Make sure the package is installed.

### DIFF
--- a/ras/ras_lsvpd.py
+++ b/ras/ras_lsvpd.py
@@ -76,7 +76,8 @@ class RASToolsLsvpd(Test):
                 deps.extend(['libsgutils-devel', 'sqlite3-devel',
                              'libvpd2-devel'])
             elif self.detected_distro.name in ['centos', 'fedora', 'rhel']:
-                deps.extend(['sqlite-devel', 'libvpd-devel'])
+                deps.extend(['sqlite-devel', 'libvpd-devel',
+                             'sg3_utils-devel'])
             else:
                 self.cancel("Unsupported Linux distribution")
             for package in deps:


sacsant:
Without the patch test fails with following error:

[stderr] configure: error: sg_utils header files are required for building lsvpd
[stdlog] 2022-12-30 04:50:40,246 avocado.utils.process INFO | Command './configure --prefix=/usr' finished with 1 after 1.717416287s

With the fix:

 (1/7) ras/ras_lsvpd.py:RASToolsLsvpd.test_build_upstream;run-run_type-upstream-f0bd: STARTED
 (1/7) ras/ras_lsvpd.py:RASToolsLsvpd.test_build_upstream;run-run_type-upstream-f0bd: PASS (9.89 s)

ras/ras_lsvpd.py
```diff
@@ -76,7 +76,8 @@ def test_build_upstream(self):
                 deps.extend(['libsgutils-devel', 'sqlite3-devel',
                              'libvpd2-devel'])
             elif self.detected_distro.name in ['centos', 'fedora', 'rhel']:
-                deps.extend(['sqlite-devel', 'libvpd-devel'])
+                deps.extend(['sqlite-devel', 'libvpd-devel',
+                             'sg3_utils-devel'])
```
Naresh-ibm, line 80: @sacsant is this from specific distro version?
in that case we have to handle to differently.

if it is for across the distro versions of `['centos', 'fedora', 'rhel']` then is this missed earlier?

sacsant: This is applicable for all the variants. I missed it the previous patch as this particular package was already installed. I ran into this error when I attempted to run the test on a freshly installed system.

PraveenPenguin (approved):

abdhaleegit (approved):
LGTM

PR #2531, opened 2023-01-24
Fix sign extension issue with read_memory function.

kalpanashtty:
In read_memory function in node_move_pages.c, the 'pattern' being compared
 with the 'value' which is a char, this leads to an sign extension issue.

### DIFF
--- a/memory/migrate_pages.py.data/node_move_pages.c
+++ b/memory/migrate_pages.py.data/node_move_pages.c
@@ -139,7 +139,7 @@ void unlock_mem(void *mmap_pointer, unsigned long memory)
 }
 
 /* Read and verify the pattern from given address for given size */
-void read_memory(char *addr, int pattern, unsigned long size, unsigned long pagesize)
+void read_memory(char *addr, char pattern, unsigned long size, unsigned long pagesize)
 {
 	unsigned long iterator;
 	unsigned long read = 0;


kalpanashtty:
##W/o fix
[root@xxxxx migrate_pages.py.data]# ./node_move_pages -n 1
Total memory size 266747158528 bytes
Nodes used in test 0 1

Scenario : Numa Page Migration

Using 26674715852 Bytes for the test


Chunk 1 Memory 26674712576 Bytes

Number of pages: 6512381 Page Size: 4096


Mapped memory at 0x7f0cdafbf000 26674712576
Lock all mapped memory
Write into mapped memory
6512381 pages allocated in Node 0
Unlock all mapped memory
All pages have been allocated in node 0

Move pages from Node 0 to Node 1
Lock all mapped memory 0x7f0cdafbf000 till
Reading from memory
Iterator 0
Correctness failed at loop read ------------------------> Issue comparing "val (char) with pattern(int)"
PATTERN MISMATCH OCCURED

-------------
##with fix
[root@xxxx migrate_pages.py.data]# ./node_move_pages -n 1
Total memory size 266750930944 bytes
Nodes used in test 0 1

Scenario : Numa Page Migration

Using 26675093094 Bytes for the test


Chunk 1 Memory 26675089408 Bytes

Number of pages: 6512473 Page Size: 4096


Mapped memory at 0x7fdf5a26c000 26675089408
Lock all mapped memory
Write into mapped memory
6512473 pages allocated in Node 0
Unlock all mapped memory
All pages have been allocated in node 0

Move pages from Node 0 to Node 1
Lock all mapped memory 0x7fdf5a26c000 till
Reading from memory
6512473 pages read and verified
6512473 pages allocated in Node 1
All pages have been allocated in node 1
Unlock all mapped memory
Unmapping Memory

Test Passed!!

kalpanashtty:
####### w/o fix
[root@xxxx memory]# avocado run --test-runner runner migrate_pages.py
Error running method "pre_tests" of plugin "fetchasset": 'Subscript' object has no attribute 'id'
JOB ID     : 0b8b0d33ce1d901d80b711d5077cc595db6042e7
JOB LOG    : /root/tests/results/job-2023-01-24T10.53-0b8b0d3/job.log
 (1/1) migrate_pages.py:MigratePages.test: CANCEL: Environment prevents test! Check logs for node data (8.52 s)
JOB HTML   : /root/tests/results/job-2023-01-24T10.53-0b8b0d3/results.html
JOB TIME   : 9.43 s

####### with fix
[root@xxxx memory]# avocado run --test-runner runner migrate_pages.py
Error running method "pre_tests" of plugin "fetchasset": 'Subscript' object has no attribute 'id'
JOB ID     : 0a0ce878f8604e4f1b636aef0dea234e61351995
JOB LOG    : /root/tests/results/job-2023-01-24T10.52-0a0ce87/job.log
 (1/1) migrate_pages.py:MigratePages.test: PASS (8.75 s)
JOB HTML   : /root/tests/results/job-2023-01-24T10.52-0a0ce87/results.html
JOB TIME   : 9.67 s

sacsant (commented):
Are we sure using a char variable is the reason for failure? Looking at the code 0xFF value is passed to read_memory() for pattern. 

#define PATTERN         0xff
read_memory(mmap_pointer[i], PATTERN, memory_to_use, page_size);

Inside the read_memory() function, 'val' defined as a character variable is used for comparing with pattern

        char val;
        for(iterator=0; iterator < size; iterator+=pagesize){
                val = addr[iterator];
                if (val != pattern) {

if pattern is changed to int aren't we then comparing int with char ?

kalpanashtty:
> Are we sure using a char variable is the reason for failure? Looking at the code 0xFF value is passed to read_memory() for pattern.
> 
> #define PATTERN 0xff read_memory(mmap_pointer[i], PATTERN, memory_to_use, page_size);
> 
> Inside the read_memory() function, 'val' defined as a character variable is used for comparing with pattern
> 
> ```
>     char val;
>     for(iterator=0; iterator < size; iterator+=pagesize){
>             val = addr[iterator];
>             if (val != pattern) {
> ```
> 
> if pattern is changed to int aren't we then comparing int with char ?

This is what ideal thing to happen, but it fails on RHEL-9.x (not sure of the compiler version need to check). Can you check on powerpc with rhel-9.x distro.

sacsant:
> This is what ideal thing to happen, but it fails on RHEL-9.x (not sure of the compiler version need to check). Can you check on powerpc with rhel-9.x distro.
Yes, the test seems to work on powerpc. 
Previously I read the change incorrectly. I read it as change char to int.
What you are proposing is opposite.I think having a matching check does seem logical.

LGTM.

kalpanashtty:
> 
I thought the issue statement was clear...
We have a signed 8-bit (char) value and compared with a 32-bit value, so the char value
seems extended, which resulted in comparison failure.
When you say the test seems to work on PowerPC, it is with the fix I believe?

sacsant:
> I thought the issue statement was clear...
> We have a signed 8-bit (char) value and compared with a 32-bit value, so the char value
> seems extended, which resulted in comparison failure.

Yes, the problem statement was clear.
As I mentioned I read the change incorrectly. Since the review was done on a holiday, I probably wasn't paying enough attention.

> When you say the test seems to work on PowerPC, it is with the fix I believe?

With the change the test works on powerpc.

sacsant (approved):

kalpanashtty:
> > I thought the issue statement was clear...
> > We have a signed 8-bit (char) value and compared with a 32-bit value, so the char value
> > seems extended, which resulted in comparison failure.
> 
> Yes, the problem statement was clear. As I mentioned I read the change incorrectly. Since the review was done on a holiday, I probably wasn't paying enough attention.
> 
> > When you say the test seems to work on PowerPC, it is with the fix I believe?
> 
> With the change the test works on powerpc.

Thank you! @sacsant

PR #2534, opened 2023-01-24
Mis-match of yaml parameter name for numactl test

Naresh-ibm:
The name of a parameter in yaml is "input_device" and in .py file it is "input_file" names were not matching. so changed yaml parameter name to "input_file" which is appropriate.


abdhaleegit (approved):
LGTM

PR #2535, opened 2023-01-25
Defaulting disk and dir yaml parameters to null

AYUSHJAIN951:
Changing disk and dir parameters to be null rather than empty strings as it produces errors by not considering default dir parameter provided. Making /mnt as default dir parameter in following tests rather than none.
And fixing styling issues in following testcases.

Signed-off-by: Ayush Jain <ayush.jain3@amd.com>

### DIFF
--- a/perf/perf_nmem.py
+++ b/perf/perf_nmem.py
@@ -173,10 +173,10 @@ class perfNMEM(Test):
             mix_events = []
             for keys in self.all_events.keys():
                 mix_events.append(self.all_events[keys][0])
-            op = process.system_output("perf stat -e '{%s}' sleep 1"
-                                       % (",".join(mix_events)), shell=True,
-                                       ignore_status=True)
-            er_ln = "The events in group usually have to be from the same PMU"
+            op = process.run("perf stat -e '{%s}' sleep 1"
+                             % (",".join(mix_events)), shell=True,
+                             ignore_status=True)
+            er_ln = "WARNING: grouped events cpus do not match, disabling group:"
             output = op.stdout.decode() + op.stderr.decode()
             # Expecting failure with the string in 'er_ln'
             if er_ln in output:
@@ -210,6 +210,8 @@ class perfNMEM(Test):
             self.log.info("Found cpumasks are = %s" % pmu_cpumask)
         else:
             self.fail("Fail to get cpumask")
+        if int(process.run("numactl -H").stdout.split()[1]) > 1:
+            self.cancel("test not supported for multiple numa nodes")
         # Get the unique value from the cpumask list
         if len(set(pmu_cpumask)) != 1:
             self.fail("cpumask values are not same: %s" % pmu_cpumask)
@@ -229,8 +231,9 @@ class perfNMEM(Test):
             self.fail("Can't offline cpumask cpu %s" % disable_cpu)
         current_cpu = self._get_cpumask(pmu1)
         self.log.info("Current CPU: %s" % current_cpu)
-        self._check_cpumask()
         # After confirming cpu got disabled, enable back
         if current_cpu in online_cpus and disable_cpu != current_cpu:
             if cpu.online(disable_cpu):
+                self.log.info("cpu enabled back %s" % online_cpus)
+            else:
                 self.fail("Can't online cpu %s" % disable_cpu)


narasimhan-v (approved):

narasimhan-v:
@AYUSHJAIN951 please add relevant logs

AYUSHJAIN951:
[log_disk_dir.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/10497274/log_disk_dir.txt)
Please refer to the attached logs
Thanks @narasimhan-v  for review

io/disk/disktest.py
```diff
@@ -64,37 +64,38 @@ def setUp(self):
...
         self.fs_create = False
-        self.raid_needed = self.params.get('raid', default=False)
+        self.raid_needed = self.params.get("raid", default=False)
         self.raid_create = False
-        self.disk = self.params.get('disk', default=None)
-        self.dir = self.params.get('dir', default=None)
-        self.fstype = self.params.get('fs', default='ext4')
-        self.raid_name = '/dev/md/sraid'
+        self.disk = self.params.get("disk", default=None)
+        self.dir = self.params.get("dir", default="/mnt")
```
Naresh-ibm, line 74: @AYUSHJAIN951 I think this may not be the right think to give /mnt as default dir. every one doesn't uses the /mnt for their test and some time it may get messup with OS disk space like for fiotest for creating the temp files on test runs.
So I advise to keep it as "None" only.

io/disk/disktest.py
```diff
@@ -64,37 +64,38 @@ def setUp(self):
...
-            if not smm.check_installed("mdadm") \
-               and not smm.install("mdadm"):
-                self.cancel('mdadm is needed for the test to be run')
+            if not smm.check_installed("mdadm") and not smm.install("mdadm"):
+                self.cancel("mdadm is needed for the test to be run")
 
-        self.vgname = 'avocado_vg'
-        self.lvname = 'avocado_lv'
+        self.vgname = "avocado_vg"
+        self.lvname = "avocado_lv"
```
Naresh-ibm, line 93: @AYUSHJAIN951 is there any specific case where this styling is a must required ?
if not then keeping existing style as it is better than getting into a mesh.

Naresh-ibm (changes requested):

io/disk/disktest.py
PraveenPenguin: yes i agree with @Naresh-ibm in your use case if you are looking to use upstream repo please have yam file and driver your scenario using that

io/disk/disktest.py
AYUSHJAIN951: These changes were recommended by "Python Black".
But as its not so much needed i have removed these changes and pushed again.

io/disk/disktest.py
AYUSHJAIN951: Thanks for these recommendations rather than "/mnt", we can have self.workdir as the self.dir in case when self.disk is provided and self.dir is not and also if both are not provided by the user.

AYUSHJAIN951:
[IO_Disk_dir.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/10544496/IO_Disk_dir.log)
logs

io/disk/ltp_fsstress.py
```diff
@@ -60,6 +60,13 @@ def setUp(self):
         self.p_val = self.params.get('p_val', default='100')
 
         smm = SoftwareManager()
+        #if self.dir is None using using self.workdir as self.dir
+        if self.disk:
+            if not self.dir:
+                self.dir = self.workdir
```
Naresh-ibm, line 66: @AYUSHJAIN951 if the disk is not given then also test will pass here. kind of agree with it.
But if disk is  given but it is not available in OS (could be some typo or having some partition on it etc.) or issue in parsing the yaml parameter to the test (we have seen this in the past) then also test will be running on the local disk and it will pass after above changes.   and the end user will be under impression that test is ran on given disk but not in real. And the whole intention of running the test on some block device will not be fulfill here.

Second reason is test like ltp, fio creates test temp file on the given dir, so one should really aware of disk space available in the given dir.

 Best way to cover this scenario will be using the another yaml file and hard code the /mnt or use extra parameter in the yaml file like `default_dir` and handle the same in the .py code like below,

```
if default_dir:
    self.dir = default_dir
```
hope this will help your scenario. tuning yaml is easiest way than changing the code flow as it will have major impact :-)

Naresh-ibm (changes requested):

io/disk/disktest.py
Naresh-ibm: @AYUSHJAIN951 you have still has the changes here, please check if it is still a mandatory change.

io/disk/ltp_fsstress.py
AYUSHJAIN951: @Naresh-ibm First the case of Disk being misspelled and similar cases ends up with a "Missing Disk in OS" prompt itself with this solution and doesn't run the test ahead.

AYUSHJAIN951: Secondly previously in your comment you referred that "/mnt" can't be used and in your recent comment you are referring to use "/mnt" default its getting quite confusing for me personally to understand, so could you please explain it more in detail.
And let's collaborate together and get this to a better solution.

io/disk/disktest.py
AYUSHJAIN951: @Naresh-ibm  Can you please check again as it's not there now.

io/disk/ltp_fsstress.py
Naresh-ibm: @AYUSHJAIN951 
1. if test ends with "Missing Disk in OS", then your disk is not available in the OS or it has some partition already hence it is not found in json format. please check the logic in the Library for this. if you still did not understand it, let me know I get you the exact logic.
2. I asked you earlier,  not to hard code /mnt in the .py file as a default parameter. Now I am suggesting you to us this parameter in a separate yaml file which will not be affecting the .py file(As existing .py not helping you for your test  scenario). In the new yaml file you can hard code all your dir values whichever you want. OR in other scenarios, where you want to use the current for test case, then for this case to handle,  I suggested to add a new parameter in yaml called, `default_dir` and add the related logic in the .py file as suggest above. this way it will not be effecting anybody and actual test scenario meant for block devices.

PraveenPenguin:
@AYUSHJAIN951 can you please release your patch , @Naresh-ibm  after respin patch please review and we have quick merge

Naresh-ibm:
@AYUSHJAIN951 please check the travis failures and re-base the commits as Praveen mentioned. we can review it and merge after that

AYUSHJAIN951:
There were couple of issues in this branch while rebasing, so closed this PR and opened a new for same https://github.com/avocado-framework-tests/avocado-misc-tests/pull/2560

PR #2551, opened 2023-02-08
disktest.py:add support for test run by device id

abdhaleegit:
In contineous test setups where device names undertest are unpredicatable and this code will help run the test by accepting unique device id which are persistant across os installs.

### DIFF
--- a/io/disk/disktest.py
+++ b/io/disk/disktest.py
@@ -73,6 +73,10 @@ class Disktest(Test):
         self.raid_needed = self.params.get('raid', default=False)
         self.raid_create = False
         self.disk = self.params.get('disk', default=None)
+        self.sysdisks = disk.get_disks()
+        if 'scsi-' in self.disk or 'nvme-' in self.disk:
+            self.disk = '/dev/disk/by-id/%s' % self.disk
+            self.sysdisks = disk.get_disks_by_id()
         self.dir = self.params.get('dir', default=None)
         self.fstype = self.params.get('fs', default='ext4')
         self.raid_name = '/dev/md/sraid'
@@ -122,7 +126,7 @@ class Disktest(Test):
         dmesg.clear_dmesg()
         self.pre_cleanup()
         if self.disk is not None:
-            if self.disk in disk.get_disks():
+            if self.disk in self.sysdisks:
                 if self.raid_needed:
                     self.create_raid(self.target, self.raid_name)
                     self.raid_create = True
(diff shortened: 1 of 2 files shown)


io/disk/disktest.py
```diff
@@ -122,7 +126,7 @@ def _init_params(self):
         dmesg.clear_dmesg()
         self.pre_cleanup()
         if self.disk is not None:
-            if self.disk in disk.get_disks():
+            if self.disk in self.sysdisks:
```
Naresh-ibm, line 129: just small suggestions,
it would have been good if we could check for exact match here. this will be helpful when we have partition on the disks like below,

/dev/disk/by-id/scsi-360050768108001b3a80000000000010b-part3'
if we pass `scsi-360050768108001b3a80000000000010b` it will match the multiple instance here.
and while trying access it with actual name, then it may not work.

io/disk/disktest.py.data/README.txt
```diff
@@ -8,6 +8,7 @@ each disk) and checking failures after each chunk is written.
 Available parameters
 --------------------
 
-disk       - Disk to be used in test.
+disk       - Provide the test disk name /dev/sda or /dev/mapper/mpatha or 
+             scsi-360050768108000000000283 or nvme-eui.364555305250000003-part1
```
Naresh-ibm, line 12: Add the command here, from where can we get this details,

Naresh-ibm (changes requested):

io/disk/disktest.py
abdhaleegit: @Naresh-ibm is there any problem to use block device instead of its already created partitions ?

abdhaleegit: @Naresh-ibm I tested, basically when you are matching a string to list using 'in' than it is exact match .. so we are good here

io/disk/disktest.py.data/README.txt
abdhaleegit: done

io/disk/disktest.py.data/README.txt
```diff
@@ -8,6 +8,8 @@ each disk) and checking failures after each chunk is written.
 Available parameters
 --------------------
 
-disk       - Disk to be used in test.
+disk       - Provide the test disk name /dev/sda or /dev/mapper/mpatha or 
+             scsi-360050768108000000000283 or nvme-eui.364555305250000003
```
PraveenPenguin, line 12: some how here i also feel we should be always drive with device id

Naresh-ibm (approved):
LGTM

Naresh-ibm:
@abdhaleegit Thanks for the PR.

PraveenPenguin:
@Naresh-ibm ah you already did merge this as still we are discussing :(

PR #2558, opened 2023-02-21
Added condition for user exists

PraveenPenguin:
as earlier test has failure in skip case as it  try to run tearDown user deletion step , this patch address delete user only when it exists


PraveenPenguin:
@disgoel can you please review this PR

disgoel (commented):
@PraveenPenguin userdel changes looks good but groupdel condition is failing on other distros as this is only specific to SUSE.

disgoel (approved):
@PraveenPenguin changes looks good to me

Issue #2557, opened 2023-02-20
Stress-ng Failing


AYUSHJAIN951:
While running Stress-ng on a system test fails with below errors
```
Fetching asset from stress-ng.py:Stressng.test
JOB ID     : 97f57fcbde2710d59d22f55f38fd426a12cac0de
JOB LOG    : /root/avocado/job-results/job-2023-02-20T07.04-97f57fc/job.log
 (1/3) stress-ng.py:Stressng.test;run-exclude-fs_type-ext4-stressors-workers-f640: STARTED
 (1/3) stress-ng.py:Stressng.test;run-exclude-fs_type-ext4-stressors-workers-f640:  ERROR: Command 'losetup -d /dev/loop0' failed.\nstdout: b''\nstderr: b'losetup: /dev/loop0: detach failed: No such device or address\n'\nadditional_info: None (52.67 s)
 (2/3) stress-ng.py:Stressng.test;run-exclude-fs_type-xfs-stressors-workers-411a: STARTED
 (2/3) stress-ng.py:Stressng.test;run-exclude-fs_type-xfs-stressors-workers-411a:  ERROR: Command 'umount /dev/loop0' failed.\nstdout: b''\nstderr: b'umount: /dev/loop0: not mounted.\n'\nadditional_info: None (16.62 s)
 (3/3) stress-ng.py:Stressng.test;run-exclude-fs_type-btrfs-stressors-workers-ccec: STARTED
 (3/3) stress-ng.py:Stressng.test;run-exclude-fs_type-btrfs-stressors-workers-ccec:  ERROR: Command 'umount /dev/loop0' failed.\nstdout: b''\nstderr: b'umount: /dev/loop0: not mounted.\n'\nadditional_info: None (16.79 s)
RESULTS    : PASS 0 | ERROR 3 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB TIME   : 89.77 s

Test summary:
stress-ng.py:Stressng.test;run-exclude-fs_type-btrfs-stressors-workers-ccec: ERROR
stress-ng.py:Stressng.test;run-exclude-fs_type-ext4-stressors-workers-f640: ERROR
stress-ng.py:Stressng.test;run-exclude-fs_type-xfs-stressors-workers-411a: ERROR
```

@disgoel  @PraveenPenguin  @Naresh-ibm

So there are couple of issues with this test currently

[...]

PR #2559, opened 2023-02-22
Remove hard coded value of loop device

PraveenPenguin:
replace hardcoded value to get free loop device as using losetup utility to get free loop device node


PraveenPenguin:
@disgoel @AYUSHJAIN951 can you please check if the above patch will help #2557 to fix

AYUSHJAIN951:
Looks Good to me!!

disgoel (approved):
@PraveenPenguin LGTM

PR #2561, opened 2023-02-23
perf_bench: add condition to check assertion error

disgoel:
Sometimes on some of the systems perf bench test fails with some ‘Assertion’ and the test gets stuck and will not proceed further. Adding a condition for test to fail when hitting assertion error.

Signed-off-by: Disha Goel [disgoel@linux.vnet.ibm.com](mailto:disgoel@linux.vnet.ibm.com)

### DIFF
--- a/perf/perf_bench.py
+++ b/perf/perf_bench.py
@@ -66,10 +66,14 @@ class perf_bench(Test):
                 self.fail("Test Failed : %s in dmesg" % fail_pattern)
 
     def run_cmd(self, cmd):
+        err_ln = "Assertion `!(ret)' failed"
         try:
-            process.run(cmd, ignore_status=False, sudo=True)
+            op = process.run(cmd, ignore_status=False, sudo=True)
         except process.CmdError as details:
             self.fail("Command %s failed: %s" % (cmd, details))
+        output = op.stdout.decode() + op.stderr.decode()
+        if err_ln in output:
+            self.fail("command %s failed with assertion code" % cmd)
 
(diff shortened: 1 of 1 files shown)


perf/perf_bench.py
```diff
@@ -66,10 +66,14 @@ def verify_dmesg(self):
                 self.fail("Test Failed : %s in dmesg" % fail_pattern)
 
     def run_cmd(self, cmd):
+        err_ln = "Assertion `!(ret)' failed"
```
PraveenPenguin, line 69: is this exact string we are seeing in assertion case if not i feel we can use regex pattern for same

disgoel: @PraveenPenguin yes this is the exact string we see

```
# Running  1x3-convergence, "perf bench numa mem -p 1 -t 3 -P 512 -s 100 -zZ0qcm --thp  1"
perf: bench/numa.c:302: bind_to_node: Assertion `!(ret)' failed.
Aborted (core dumped)

# Running main, "perf bench numa numa-mem -p 2 -t 1 -P 1024 -C 0,8 -M 1,0 -s 20 -zZq --thp 1 --no-data_rand_walk"
# NOTE: ignoring bind NODEs starting at NODE#1
# NOTE: 0 tasks mem-bound, 2 tasks unbound
perf: bench/numa.c:334: bind_to_cpumask: Assertion `!(ret)' failed.
```

PraveenPenguin (approved):

PR #2566, opened 2023-02-28
Xfstest group support

disgoel:
added group parameter support in xfstest to run any group specific tests
fixed cleanup in teardown
update all yaml files with group and test_range parameter

Signed-off-by: Disha Goel [disgoel@linux.vnet.ibm.com](mailto:disgoel@linux.vnet.ibm.com)

with both group and test_range as null
JOB ID     : a8f8df6816cf620a827159a931b4e7e2bc0a3175
JOB LOG    : /home/avocado-fvt-wrapper/results/job-2023-02-28T02.32-a8f8df6/job.log
 (1/1) xfstests.py:Xfstests.test;run-fs_type-fs_ext4_4k-gen_exclude-share_exclude-group-loop_type-disk-test_range-bd68: STARTED
 (1/1) xfstests.py:Xfstests.test;run-fs_type-fs_ext4_4k-gen_exclude-share_exclude-group-loop_type-disk-test_range-bd68: CANCEL: incorrect yaml parameter, group and test range can                          not be run/none at same time (0.61 s)
JOB HTML   : /home/avocado-fvt-wrapper/results/job-2023-02-28T02.32-a8f8df6/results.html
JOB TIME   : 13.12 s

with group parameter
JOB ID     : 7ada616a8184c943c30dc5801c32e20a29f8324a
JOB LOG    : /home/avocado-fvt-wrapper/results/job-2023-02-28T02.34-7ada616/job.log
 (1/1) xfstests.py:Xfstests.test;run-fs_type-fs_ext4_4k-gen_exclude-share_exclude-loop_type-disk-test_range-83c9: STARTED
 (1/1) xfstests.py:Xfstests.test;run-fs_type-fs_ext4_4k-gen_exclude-share_exclude-loop_type-disk-test_range-83c9: PASS (23.13 s)
JOB HTML   : /home/avocado-fvt-wrapper/results/job-2023-02-28T02.34-7ada616/results.html
JOB TIME   : 34.69 s

[job.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/10848237/job.log)

### DIFF
--- a/fs/xfstests.py
+++ b/fs/xfstests.py
@@ -223,6 +223,7 @@ class Xfstests(Test):
         self.skip_dangerous = self.params.get('skip_dangerous', default=True)
         self.group = self.params.get('group', default='auto')
         self.test_range = self.params.get('test_range', default=None)
+        self.base_disk = self.params.get('disk', default=None)
         self.scratch_mnt = self.params.get(
             'scratch_mnt', default='/mnt/scratch')
         self.test_mnt = self.params.get('test_mnt', default='/mnt/test')
@@ -321,7 +322,6 @@ class Xfstests(Test):
                           ignore_status=True):
             self.cancel('Unknown filesystem %s' % self.fs_to_test)
         mount = True
-        self.devices = []
         self.log_devices = []
         shutil.copyfile(self.get_data('local.config'),
                         os.path.join(self.teststmpdir, 'local.config'))
@@ -332,9 +332,8 @@ class Xfstests(Test):
         self.log_scratch = self.params.get('log_scratch', default='')
 
         if self.dev_type == 'loop':
-            base_disk = self.params.get('disk', default=None)
             loop_size = self.params.get('loop_size', default='7GiB')
-            if not base_disk:
+            if not self.base_disk:
                 # Using root for file creation by default
                 check = (int(loop_size.split('GiB')[0]) * 2) + 1
                 if disk.freespace('/') / 1073741824 > check:
@@ -342,7 +341,7 @@ class Xfstests(Test):
                     mount = False
                 else:
                     self.cancel('Need %s GB to create loop devices' % check)
-            self._create_loop_device(base_disk, loop_size, mount)
+            self._create_loop_device(loop_size, mount)
         elif self.dev_type == 'nvdimm':
             self.setup_nvdimm()
         else:
@@ -498,7 +497,7 @@ class Xfstests(Test):
             for dev in self.devices:
                 process.system('losetup -d %s' % dev, shell=True,
                                sudo=True, ignore_status=True)
-            if self.disk_mnt:
+            if self.part:
                 self.part.unmount()
         elif self.dev_type == 'nvdimm':
             if hasattr(self, 'region'):
@@ -513,10 +512,10 @@ class Xfstests(Test):
                             process.system('losetup -d %s' % dev, shell=True,
                                            sudo=True, ignore_status=True)
 
-    def _create_loop_device(self, base_disk, loop_size, mount=True):
+    def _create_loop_device(self, loop_size, mount=True):
         if mount:
             self.part = partition.Partition(
-                base_disk, mountpoint=self.disk_mnt)
+                self.base_disk, mountpoint=self.disk_mnt)
             self.part.mount()
         # Creating two loop devices
         for i in range(2):
--- a/fs/xfstests.py
+++ b/fs/xfstests.py
@@ -82,7 +82,7 @@ class Xfstests(Test):
                     mount = False
                 else:
                     self.cancel('Need %s GB to create loop devices' % check)
-                self._create_loop_device(None, '2038M', mount)
+                self._create_loop_device('2038M', mount)
                 self.log_test = self.devices.pop()
                 self.log_scratch = self.devices.pop()
             namespaces = self.plib.run_ndctl_list('-N -r %s' % self.region)
@@ -221,6 +221,7 @@ class Xfstests(Test):
                 self.cancel("Fail to install %s required for this test." %
                             package)
         self.skip_dangerous = self.params.get('skip_dangerous', default=True)
+        self.group = self.params.get('group', default='auto')
         self.test_range = self.params.get('test_range', default=None)
         self.scratch_mnt = self.params.get(
             'scratch_mnt', default='/mnt/scratch')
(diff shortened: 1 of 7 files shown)


fs/xfstests.py
```diff
@@ -82,7 +82,7 @@ def setup_nvdimm(self):
                     mount = False
                 else:
                     self.cancel('Need %s GB to create loop devices' % check)
-                self._create_loop_device(None, '2038M', mount)
+                self._create_loop_device('2038M', mount)
```
abdhaleegit, line 85: @disgoel hardcoding disk size incode, please see if you can pass it from yaml... as available disk space per system varies.. so user can test with smaller or bigger loop device

fs/xfstests.py
```diff
@@ -221,14 +221,22 @@ def setUp(self):
...
         self.scratch_mnt = self.params.get(
             'scratch_mnt', default='/mnt/scratch')
         self.test_mnt = self.params.get('test_mnt', default='/mnt/test')
         self.disk_mnt = self.params.get('disk_mnt', default='/mnt/loop_device')
         self.fs_to_test = self.params.get('fs', default='ext4')
         self.run_type = self.params.get('run_type', default='distro')
 
+        self.devices = []
+        self.part = None
+        if (self.group and self.test_range) or (self.group is None and self.test_range is None):
```
abdhaleegit, line 236: Regarding `(self.group is None and self.test_range is None):`

None is a bolean, it is sufficient to just check the param like `if self.group`  or `if not self.group`  and no need to explicitly use `is None` 

>>> group = None
>>> if group:
...     print(True)
... 
>>> if not group:
...     print(True)
... 
True
>>>

fs/xfstests.py.data/btrfs-subpage.yaml
```diff
@@ -2,7 +2,8 @@ skip_dangerous: True
 scratch_mnt: '/mnt/scratch'
 test_mnt: '/mnt/test'
 disk_mnt: '/mnt/loop-device'
-test_range: '54-55'
+group: 'auto'
+test_range: null
```
abdhaleegit, line 6: for all below yamls.. if there is no value.. keep it empty.  thats the standard we follow .. no need to provide null
in code when params.get  keep default value empty

abdhaleegit (changes requested):

fs/xfstests.py.data/btrfs-subpage.yaml
disgoel: updated all the yaml files

fs/xfstests.py
disgoel: @abdhaleegit 
looks like legacy region ndctl is not supported on powerpc, this is supported on x86.
In this condition we are using loop device as a log device and xfs restrict max log size to 2136997888, which is 10M less than 2GB. So we need to set loop_size as 2038M here.

fs/xfstests.py
abdhaleegit: any reason to have multiple same checks  ?
if (self.group and self.test_range) or (self.group is None and self.test_range is None):  

will just if (self.group and self.test_range):  not sufficient ?

abdhaleegit (commented):
Rest all looks good

fs/xfstests.py
disgoel: @abdhaleegit we need to have both the checks:
If both parameters are set test can not be run as its not supported together.
Also if both the parameters are none test will not run because we dont have any default value.

abdhaleegit: @disgoel so in that case please make sure the group default value is not None, meaning

in code keep default value to 'auto' for group so the test should not cancel in case yaml is not used and also only this single check is enough in this case
if (self.group and self.test_range):

so with this
1. Always group has default auto value irrespective of yaml is provided or just .py file is run
2. (self.group is None and self.test_range is None) is not needed as at any time one param is set
3. If someone wants to run test_range.. make sure they provide yaml with empty group and test_range value, please document this in Readme

disgoel: @abdhaleegit if we keep group default value in .py with this the test_range parameter will never run because both parameters will be set and test is going to be cancelled.

abdhaleegit: @disgoel thats the 3rd case I mentioned in previous comment, so in case of test_range test .. user must provide yaml with group as empty and a value in test_range.

basically make sure the .py work for 1 default value always ... though it will cancel fail if both values are None .. so instead make it run for default

abdhaleegit: @PraveenPenguin your views ?

disgoel: @abdhaleegit if we keep default group, when we pass null or empty group in yaml still its going to take default value.
`self.group = self.params.get('group', default='auto')`

```
group:
# Uncomment and edit test_range for running specific tests
test_range: '1-5'

PARAMS (key=group, path=*, default=auto) => 'auto'
```

```
group: null
# Uncomment and edit test_range for running specific tests
test_range: '1-5'

PARAMS (key=group, path=*, default=auto) => 'auto'
```

disgoel: one more way we can implement is we can have empty group and test_range in yaml, user can edit for specific use cases. In .py we can pass default as none in starting and later we can check if group is none it should take default value as auto. 
Tested with this changes works fine for both group and test_range scenario.

`self.test_range = self.params.get('test_range', default=None)
`
```
            if not self.group:
                self.group = 'auto'
            cmd = './check %s -g %s' % (args, self.group)
```
@abdhaleegit @PraveenPenguin your thoughts with this approach ?

abdhaleegit: > one more way we can implement is we can have empty group and test_range in yaml, user can edit for specific use cases. In .py we can pass default as none in starting and later we can check if group is none it should take default value as auto. Tested with this changes works fine for both group and test_range scenario.
> 
> `self.test_range = self.params.get('test_range', default=None) `
> 
> ```
>             if not self.group:
>                 self.group = 'auto'

so this would add one extra check.. but the same can be achieved by
self.test_range = self.params.get('test_range', default=auto)

disgoel: > No, yaml input value takes precedence over py default values
> Also do not use null, instead use None with out qoutes

I have set group default value as auto in py and None in yaml. Its taking None as a group.

```
group: None
PARAMS (key=group, path=*, default=auto) => 'None'
PARAMS (key=test_range, path=*, default=None) => '1-5'
Running './check  -E /var/tmp/avocado_lt6n6xnw/exclude -g None'
```

PraveenPenguin: I agree with @abdhaleegit view, the test should run without ymal and ymal has high precedence always , Rest looks good to me

disgoel: @PraveenPenguin @abdhaleegit have addressed the changes, please review. Also updated the README file with latest changes.

PraveenPenguin (approved):
LGTM

abdhaleegit (approved):
LGTM

PR #2560, opened 2023-02-23
Defaulting disk and dir yaml parameters to null

AYUSHJAIN951:
Changing disk and dir parameters to be null rather than empty strings as it produces errors by not considering default dir parameter provided. Making /mnt as default dir parameter in following tests rather than none.
And fixing styling issues in following testcases.

Signed-off-by: Ayush Jain [ayush.jain3@amd.com](mailto:ayush.jain3@amd.com)

PR #2574, opened 2023-03-07
Fixing self.disk and self.dir parameters

AYUSHJAIN951:
Defaulting self.dir to be self.workdir rather than /mnt
And Cancelling test when self.disk not provided
or missing.


AYUSHJAIN951:
@Naresh-ibm @narasimhan-v @PraveenPenguin  Please Review

io/disk/fiotest.py
```diff
@@ -73,6 +73,18 @@ def setUp(self):
         self.raid_create = False
         self.devdax_file = None
 
+        # Checking if self.disk input provided or not, if not Cancelling test
+        if self.disk:
+            # Checking if self.disk is a valid disk in OS
+            if self.disk in disk.get_disks():
+                #if self.dir is None using using self.workdir as self.dir
+                if self.dir is None:
+                    self.dir = self.workdir
```
Naresh-ibm, line 82: @AYUSHJAIN951 why you want to change dir only when the disk is given.
why cannot we take it out of this loop.
we already has a parameter in yaml , i,e dir. where the user will be providing the dir.
if you want to use the self.worldir, then use it in the following manner,

```
if not self.dir:
    self.dir = self.worldir
```
Please re-check on this.

io/disk/fiotest.py
```diff
@@ -138,33 +150,27 @@ def setUp(self):
...
-                self.lv_disk = self.target
-                self.target = self.create_lv(self.target)
-                self.lv_create = True
-
-            if fstype:
-                self.create_fs(self.target, self.dir, fstype,
-                               fs_args, mnt_args)
-                self.fs_create = True
-        else:
-            self.cancel("Missing disk %s in OS" % self.disk)
```
Naresh-ibm, line 167: @AYUSHJAIN951 fiotest is also used for a pmem testing where they have the different ways of using it.
And I don't think the existing way of handling has any problem.

what I feel is you want to extend these test cases to use on OS dir?
if that is the case, please come up with a proper way of handling it instead of just moving the thing up and down.
let me know if you need any help here.

Naresh-ibm (changes requested):

io/disk/fiotest.py
AYUSHJAIN951:
```
JOB ID     : 0bfe15e4ed99a1f27d43237a525f79b18fbc9b78
JOB LOG    : /root/tests/results/job-2023-03-07T05.17-0bfe15e/job.log
 (1/1) fiotest.py:FioTest.test;run-dir-disk-fs-ext4-lv-lv-raid-raid-5880: STARTED
 (1/1) fiotest.py:FioTest.test;run-dir-disk-fs-ext4-lv-lv-raid-raid-5880: ERROR: 'NoneType' object has no attribute 'split' (1.12 s)
RESULTS    : PASS 0 | ERROR 1 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /root/tests/results/job-2023-03-07T05.17-0bfe15e/results.html
JOB TIME   : 2.61 s
```
In the current version of fiotest.py if disk is not provided it fails with this error on line 147 
        self.sraid = softwareraid.SoftwareRaid(self.raid_name, '0',
                                               self.disk.split(), '1.2')
which is needed to be handled thats why pushing disk checks above that

io/disk/fiotest.py
AYUSHJAIN951: Because as you quoted in https://github.com/avocado-framework-tests/avocado-misc-tests/pull/2560#:~:text=Note%20%3A%20These%20test%20cases%20are%20mainly%20intended%20for%20a%20block%20devices%20and%20may%20not%20be%20useful%20to%20run%20on%20only%20directory.
Note : These test cases are mainly intended for a block devices and may not be useful to run on only directory.

So if self.disk is not provided there is no point of changing self.dir anyways.

io/disk/fiotest.py
Naresh-ibm: I could see it printing the Missing disk message on not providing the disk in yaml.

io/disk/fiotest.py
Naresh-ibm: @AYUSHJAIN951 I will re-iterate the same again here,

The test scripts were mainly for block devices as of now. if you want to extend the same script to run on any OS installed disk, following problems might occur,

1. since we are cleaning the disk before starting the test and after starting the test, your OS may get corrupt when you use dir from the OS.
2. few test cases consumes lot of space on the disk/dir and OS disk may not be capable of providing it all the time.

Your always welcome to add any new scenarios to these script, but make sure that your changes should not break the current flow or functional verification or should not lead to other problems.

take the example of the same fiotest here, we have pmem test available on this.
they have handled it using yaml and scripts in .py which is not disturbing other functionality verification here.

your always welcome to add the scenarios in that way.

 coming to #2560 , you were trying run the test on self.workdir and the test was cancelling when the disk is not given.
obviously when you want to run the test on self.workdir, user will not provide the disk value and test will cancel without running on either them, disk/dir.
That was a invalid scenario, and you might need to tweak it properly.

one more case,
if an user wants to run the test on some disk and he provides the disk in yaml, but due to some parsing issue, input parameters are not parsed and test ran on some temp dir and test pass.
then this is also a false PASS, as he wants to run the test on disk which was not happened but results say it is passed

So, as a summery, take care of all such cases and add any number of scenarios.

hope you got my view.

AYUSHJAIN951:
@Naresh-ibm  Can you review again with the latest changes.

AYUSHJAIN951:
Ping @Naresh-ibm

AYUSHJAIN951:
>

Naresh-ibm (changes requested):

io/disk/fiotest.py
```diff
@@ -138,7 +138,16 @@ def setUp(self):
...
+        # Checking if self.disk provided or not, if not Cancelling test
+        if self.disk:
+            # Checking if self.disk is a valid disk in OS
+            if self.disk not in disk.get_disks():
+                self.cancel("Missing disk %s in OS" % self.disk)
+        else:
+            self.cancel("Please Provide valid disk name")
+
+        #if self.dir is None using using self.workdir as self.dir
+        if not self.dir:
```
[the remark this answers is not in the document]
AYUSHJAIN951: sure

io/disk/fiotest.py
```diff
@@ -138,7 +138,13 @@ def setUp(self):
...
                         self.cancel("PMEM engines not built with fio")
 
-        if not self.disk:
+        if self.disk:
+            if self.disk not in disk.get_disks():
+                self.cancel("Missing disk %s in OS" % self.disk)
+        else:
+            self.cancel("Please Provide valid disk name")
+
+        if not self.dir:
```
Naresh-ibm, line 147: @AYUSHJAIN951 tried to understand the different yaml files we have for this test case and discuss with the end user of `fio-pmem.yaml,` they want to run the test on pmem chunk and may not be passing the disk all the time.

even I could see there is no disk parameter in `fio-perf.yaml` and they have hard coded the /mnt in the this yaml which indicates they will run this test on /mnt and without disk.

So if we check only disk here and cancel the test its not going to help other users.
so lets have a common yaml parameter across this yaml files.
parameter can be like, 

`run_on_disk : True`

So for test to run on any san/sas or on any raw disk we set it as `True` else `False`

Add above line as comment at this parameter in yaml.
So for fio.yaml, it will be `True` and for other yamls it will be `False`. hard code this in yaml accordingly.

Now in .py
```
if run_on_disk:
    if not self.disk:
         self.cancel ("provide the disk to run test")
if  not self.dir:
     self.dir = self.workdir
```

@narasimhan-v let me know your thoughts too on this to handle

Naresh-ibm: one more thing,
creating lvs/raids will be done only when the test is intended to run raw disk. and lvs and raids not applicable in other case.
So have this check in such a way that creation of lvs/raids will happen only when disk is provided and in other case it should not be ran.

let me know if you need any help here.

Naresh-ibm (changes requested):

io/disk/fiotest.py
narasimhan-v: @Naresh-ibm are we unnecessarily complicating the test case to handle all users of it ? I can see that every change is required to handle all types of inputs.
Can we have a separate test case for pmem and simplify the test rather than adding parameters which is confusing to anyone new to contribute ?

narasimhan-v: Right now I can see the test case is having numerous parameters manipulated, like self.dir, self.workdir, etc. This is not intuitive and makes it difficult for anyone to send a quick fix. I strongly think we should just clone the test and have it separately for pmem and perf, since I do not see any contributions from those users for a long time now.

io/disk/fiotest.py
```diff
@@ -59,7 +59,7 @@ def setUp(self):
         default_url = "https://brick.kernel.dk/snaps/fio-git-latest.tar.gz"
         url = self.params.get('fio_tool_url', default=default_url)
         self.disk = self.params.get('disk', default=None)
-        self.dir = self.params.get('dir', default='/mnt')
+        self.dir = self.params.get('dir', default=None)
```
PraveenPenguin, line 62: sorry  for the late review, this looks good as it can be YAML driven but what is the harm in having `/mnt` as default as we always have way override this via YAML ,  @AYUSHJAIN951 if you are using other set for yaml i suggest push those yaml in upstream

io/disk/fiotest.py
PraveenPenguin: change of this part of the code seems promising, @AYUSHJAIN951 please take care pmem.yaml may we need have align that @Naresh-ibm can you also please help here

io/disk/fiotest.py
PraveenPenguin: How does this sound have YAML file part of upstream and in clone it will work, new YAML aligned with the user's needs, then we should be having a generic solution. and avoid to   change py file,

io/disk/fiotest.py
Naresh-ibm: @PraveenPenguin we have agreed to not have any default value for this parameter as it has other side effects when comes to block devices or raw disk. so I feel having `None` is fine

Naresh-ibm:
@narasimhan-v I agree on your view, that each components should have separate test, but I feel it is not that clumsy as of now here.
more over we don't know when the other users comes back to use it. handling with extra hard coded yaml parameter will solve this issue instead of splitting the test as of now.
I am not aware of exact functionality on nvdia type of disk here in order to split and make different test.
So lets re-think on this. even Praveen's comments conveys the same message.
We are still not understanding what exact functionalities you guys are looking with this changes. As of now what I understand is @AYUSHJAIN951 wants to run the test on slef.workdir and tweak the code accordingly. And this will be achieved without disturbing the .py file and adding the extra yaml parameter as I suggested.

narasimhan-v:
> @narasimhan-v I agree on your view, that each components should have separate test, but I feel it is not that clumsy as of now here.

A top level look at this PR and its history gives me a view that we are struggling to achieve running test on disk and having many parameters manipulated which makes it more difficult to achieve this.

> more over we don't know when the other users comes back to use it. handling with extra hard coded yaml parameter will solve this issue instead of splitting the test as of now.

Exactly what I have stated above. Without going into details, I see the comments are difficult to understand than the code changes itself, IMO. I think we all need to re-evaluate our goal in this PR and try to see what works best for all stakeholders. Different yaml is fine, but if code changes are needed, we need to come to a common ground that is beneficial for all of us.


> As of now what I understand is @AYUSHJAIN951 wants to run the test on slef.workdir and tweak the code accordingly. And this will be achieved without disturbing the .py file and adding the extra yaml parameter as I suggested.

@AYUSHJAIN951 Need your comments here.

AYUSHJAIN951:
> @narasimhan-v I agree on your view, that each components should have separate test, but I feel it is not that clumsy as of now here. more over we don't know when the other users comes back to use it. handling with extra hard coded yaml parameter will solve this issue instead of splitting the test as of now. I am not aware of exact functionality on nvdia type of disk here in order to split and make different test. So lets re-think on this. even Praveen's comments conveys the same message. We are still not understanding what exact functionalities you guys are looking with this changes. As of now what I understand is @AYUSHJAIN951 wants to run the test on slef.workdir and tweak the code accordingly. And this will be achieved without disturbing the .py file and adding the extra yaml parameter as I suggested.

So the need of this PR was to have a 
1.Default directory other than '/mnt' to run tests on which can be easily achieved by self.workdir the way i have proposed in this PR
2. To avoid **ERROR: 'NoneType' object has no attribute 'split'**
3.There is no proper structure in this  testcase and also in other io/disk tests to verify a disk is provided by user/yaml or not, and if provided is it a valid disk or not.

```
         if self.disk:
            if self.disk not in disk.get_disks():
                self.cancel("Missing disk %s in OS" % self.disk)
         else:
            self.cancel("Please Provide valid disk name")
```
This solution works well for the disk check  for fio.yaml, fio-pmem.yaml 


And if we can have similar solution in all other io/disk testcases as well this gives us a better flow in all tests to understand and manipulate with the parameters as well when needed.
So having a change in py file gives a better edge in longer term rather than yaml changes now and hitting similar range issues next again.

PraveenPenguin:
> > @narasimhan-v I agree on your view, that each components should have separate test, but I feel it is not that clumsy as of now here. more over we don't know when the other users comes back to use it. handling with extra hard coded yaml parameter will solve this issue instead of splitting the test as of now. I am not aware of exact functionality on nvdia type of disk here in order to split and make different test. So lets re-think on this. even Praveen's comments conveys the same message. We are still not understanding what exact functionalities you guys are looking with this changes. As of now what I understand is @AYUSHJAIN951 wants to run the test on slef.workdir and tweak the code accordingly. And this will be achieved without disturbing the .py file and adding the extra yaml parameter as I suggested.
> 
> So the need of this PR was to have a 1.Default directory other than '/mnt' to run tests on which can be easily achieved by self.workdir the way i have proposed in this PR 2. To avoid **ERROR: 'NoneType' object has no attribute 'split'** 3.There is no proper structure in this testcase and also in other io/disk tests to verify a disk is provided by user/yaml or not, and if provided is it a valid disk or not.
> 
> ```
>          if self.disk:
>             if self.disk not in disk.get_disks():
>                 self.cancel("Missing disk %s in OS" % self.disk)
>          else:
>             self.cancel("Please Provide valid disk name")
> ```
> 
> This solution works well for the disk check for fio.yaml, fio-pmem.yaml
> 
> And if we can have similar solution in all other io/disk testcases as well this gives us a better flow in all tests to understand and manipulate with the parameters as well when needed. So having a change in py file gives a better edge in longer term rather than yaml changes now and hitting similar range issues next again.

@AYUSHJAIN951 sorry for the late response seems promising, if the solution works for the major case we are good to go in 

@Naresh-ibm what do you say?

Naresh-ibm:
lets us merge this for now, if any others gets effected(pmem or nvdimm) we will change the script accordingly at later stage as we see , we don't have any inputs to this script  from past so many days in PMEM and Nvdimm areas and similar functionalities are also available in other scripts.

@PraveenPenguin @narasimhan-v Please ACK, will merge this

PraveenPenguin:
> lets us merge this for now, if any others gets effected(pmem or nvdimm) we will change the script accordingly at later stage as we see , we don't have any inputs to this script from past so many days in PMEM and Nvdimm areas and similar functionalities are also available in other scripts.
> 
> @PraveenPenguin @narasimhan-v Please ACK, will merge this

thanks @Naresh-ibm

PR #2589, opened 2023-03-29
Change class and definition names to unique name

Krishan-Gopal-saraswat:
Changing class and definition names of the scripts as per the test case name.


PraveenPenguin (approved):
LGTM

PR #2590, opened 2023-03-30
EEH for block devies are updated. Added validation to check multipath devices [wait for #5632]

maramsmurthy:
EEH for block devies are updated. Added validation to check multipath devices

after EEH.

EEH for block devies are updated. Added validation to check multipath devices after EEH.

### DIFF
--- a/io/pci/EEH.py
+++ b/io/pci/EEH.py
@@ -26,6 +26,8 @@ from avocado.utils import pci
 from avocado.utils import genio
 from avocado.utils import distro
 from avocado.utils import dmesg
+from avocado.utils import multipath
+from avocado.utils import genericutility
 from avocado.utils.software_manager.manager import SoftwareManager
 
 EEH_HIT = 0
@@ -44,7 +46,9 @@ class EEHRecoveryFailed(Exception):
         self.log = log
 
     def __str__(self):
-        return "%s %s recovery failed: %s" % (self.msg, self.dev, self.log)
+        return f"{self.msg} {self.dev} recovery failed: {self.log}" % (
+                self.msg, self.dev, self.log
+            )
 
 
 class EEH(Test):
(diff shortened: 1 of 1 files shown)


maramsmurthy:
It can be merged only after https://github.com/avocado-framework/avocado/pull/5632

abdhaleegit:
Run logs please

maramsmurthy:
Please find attached job.log file
[job.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/11135003/job.log)

abdhaleegit (commented):
@maramsmurthy will this handle to not try max eeh tries and remove adapter ?

maramsmurthy:
> @maramsmurthy will this handle to not try max eeh tries and remove adapter ? the last discussion we had was to try max value -1 so the adapter will not be removed for other tests

Hi Abdul, Yes handled it here.

Naresh-ibm:
@maramsmurthy as discussed and agreed, please sent the new PR for validation and re-push this PR with only part of fixes which resolve the other issue basic issues

Naresh-ibm (changes requested):
as suggested above please split the PR into two part and re-send this PR

maramsmurthy:
Updated to use exiting utils instead of new one.

io/pci/EEH.py
```diff
@@ -121,31 +129,41 @@ def test_eeh_basic_pe(self):
...
+                                        f" EEH"
+                                )
                                 break
                             else:
                                 self.log.info("PE recovered successfully")
+                        time.sleep(10)
+                        if pci.get_pci_class_name(self.pci_device) == "fc_host":
+                            after_eeh_path_status = multipath.get_multipath_details()
+                            get_diff_bef_aft = genericutility.recursive_compare_dict(
+                                    before_eeh_path_status, after_eeh_path_status
```
Naresh-ibm, line 150: @maramsmurthy some how I am not convinced with dictionary way of validation, because the dictionary holds many different parameters which may change after driver reload on EEH.
if any of the parameter also gets changed then the test will fail.

So my view is just focus on only what needs to be validated than validating everything which is even not necessary.

Naresh-ibm (changes requested):

io/pci/EEH.py
```diff
@@ -75,15 +79,15 @@ def setUp(self):
         self.mem_addr = pci.get_memory_address(self.pci_device)
         self.mask = pci.get_mask(self.pci_device)
         if self.is_baremetal():
-            cmd = "echo %d > /sys/kernel/debug/powerpc/eeh_max_freezes"\
-                % self.max_freeze
+            cmd = f"echo > {self.max_freeze} > /sys/kernel/debug/powerpc/eeh_max_freezes"
```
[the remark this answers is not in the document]
maramsmurthy: ok, Sure Will change it

io/pci/EEH.py
maramsmurthy: @Naresh-ibm You are correct. I verified with successfull EEH recovery case by doing same comparison. Only two attributes are getting impacted as per your "switch_grp" and "path_faults". I will be excluding these attributes from comparison.

maramsmurthy: @Naresh-ibm  Addressed the problem as suggested

maramsmurthy: Please find attached log file for reference
[job_dict_check.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/11207825/job_dict_check.log)

PR #2603, opened 2023-04-12
ras_ppcutils.py: Validating lparstat functionality

shirishaganta1:
Added a new testcase to validate below two
1.laprstat -E output %busy and %idle should not be < 0 and > 100 
2.Normalized %busy + %idle = percentage under frequency

### DIFF
--- a/ras/ras_ppcutils.py
+++ b/ras/ras_ppcutils.py
@@ -14,8 +14,10 @@
 # Copyright: 2016 IBM
 # Author: Pavithra <pavrampu@linux.vnet.ibm.com>
 # Author: Sachin Sant <sachinp@linux.ibm.com>
+# Author: Shirisha Ganta <shirisha.ganta1@ibm.com>
 
 import os
+import re
 from avocado import Test
 from avocado.utils import process, distro, build, archive
 from avocado import skipIf, skipUnless
@@ -410,6 +412,9 @@ class RASToolsPpcutils(Test):
         """
         Test case to validate lparstat functionality. lparstat is a tool
         to display logical partition related information and statistics.
+        And also validates laprstat -E output %busy and %idle
+        should not be < 0 or > 100
+        Normalized %busy + %idle should be equal to percentage under frequency
         """
         self.log.info("===============Executing lparstat tool test===="
                       "===========")
@@ -428,6 +433,31 @@ class RASToolsPpcutils(Test):
                 self.log.info("%s command passed" % cmd)
                 self.fail("lparstat: Expected failure, %s command exeucted \
                           successfully." % cmd)
+        output = process.system_output("lparstat -E 1 1").decode("utf-8")
+        for line in output.splitlines():
+            if 'GHz' in line:
+                # Define the regular expression pattern
+                pattern = (r'(\d+\.\d+)\s+(\d+\.\d+)\s+\d+\.\d+GHz\[\s*(\d+)%\]\s+'
+                           r'(\d+\.\d+)\s+(\d+\.\d+)')
+                # Find all matches in the input string
+                matches = re.findall(pattern, line)
+                for data in matches:
+                    actual_busy = float(data[0])
+                    actual_idle = float(data[1])
+                    normal_idle = float(data[3])
+                    normal_busy = float(data[4])
+                    normal = normal_idle + normal_busy
+                    freq_percentile = float(data[2])
+        if (actual_busy > 0) and (actual_idle < 100):
+            self.log.info("Busy and idle actual values are correct")
+        else:
+            self.fail("Busy and idle actual values are incorrect")
+        if normal == freq_percentile:
+            self.log.info("Normalised busy plus idle value match with \
+                          Frequency percentage")
+        else:
+            self.fail("Normalised busy plus idle value does not match \
+                        with Frequency percentage")
 
     @skipIf(IS_POWER_NV or IS_KVM_GUEST,
             "This test is not supported on KVM guest or PowerNV platform")


ras/ras_ppcutils.py
```diff
@@ -447,3 +449,40 @@ def test_lparnumascore(self):
...
+        should not be < 0 or > 100
+        Normalized %busy + %idle should be equal to percentage under frequency
+        """
+        output = process.system_output("lparstat -E 1 1").decode("utf-8")
+        input_string = ''
+        for line in output.splitlines():
+            if 'GHz' in line:
+                input_string = line
+                # Define the regular expression pattern
+                pattern = r'(\d+\.\d+)\s+(\d+\.\d+)\s+\d+\.\d+GHz\[\s*(\d+)%\]\s+(\d+\.\d+)\s+(\d+\.\d+)'
```
Naresh-ibm, line 467: @shirishaganta1 looks like the length is >79. please take care of it.

Naresh-ibm (changes requested):

ras/ras_ppcutils.py
```diff
@@ -447,3 +449,40 @@ def test_lparnumascore(self):
...
+                    normal_idle = float(data[3])
+                    normal_busy = float(data[4])
+                    normal = normal_idle + normal_busy
+                    freq_percentile = float(data[2])
+        if (actual_busy > 0) and (actual_idle < 100):
+            self.log.info(
+                "Actual busy and idle values are displaying  correctly")
+        else:
+            self.cancel(
+                "busy value and idle values are not  displaying correctly")
```
abdhaleegit, line 482: should fail instead of cancel ?

ras/ras_ppcutils.py
```diff
@@ -447,3 +449,40 @@ def test_lparnumascore(self):
...
+        if normal == freq_percentile:
+            self.log.info("normal busy and idle are matching with the \
+                           percentage under frequency")
+        else:
+            self.cancel("normal busy and value percentage are not matching \
+                        with the frequency percentile")
```
abdhaleegit, line 488: here too should fail instead of cancel ?

abdhaleegit (changes requested):
we can always avoid one condition in if /else  and only log if fail .. and ignore log.info if pass

ras/ras_ppcutils.py
shirishaganta1: sure will take care of that

ras/ras_ppcutils.py
shirishaganta1: yeah, for the above two cases I have changed it to fail if the condition is not met

sacsant (commented):
Why do we need a separate test_ routine? Can we add this code to existing test_lparstat() routine ?

ras/ras_ppcutils.py
shirishaganta1: > Why do we need a separate test_ routine? Can we add this code to existing test_lparstat() routine ?

yes we can do that..

Naresh-ibm:
@shirishaganta1 please re-push taking care of Travis and above suggested changes

shirishaganta1:
Addressed all the changes suggested above and re-pushed it

Naresh-ibm:
@shirishaganta1 are we still pursuing this  in ras_ppcutils.py?
Or
as suggested by Sachin, will add this code to existing test_lparstat() routine ?

PraveenPenguin (approved):
LGTM @Naresh-ibm it already moved in lpar stats

ras/ras_ppcutils.py
```diff
@@ -428,6 +433,35 @@ def test_lparstat(self):
...
+                    actual_idle = float(data[1])
+                    normal_idle = float(data[3])
+                    normal_busy = float(data[4])
+                    normal = normal_idle + normal_busy
+                    freq_percentile = float(data[2])
+        if (actual_busy > 0) and (actual_idle < 100):
+            self.log.info(
+                "Actual busy and idle values are displaying  correctly")
+        else:
+            self.fail("busy value and idle values are not \
```
sacsant, line 457: Busy and idle actual values are incorrect.

ras/ras_ppcutils.py
```diff
@@ -428,6 +433,35 @@ def test_lparstat(self):
...
+                       displaying correctly")
+        if normal == freq_percentile:
+            self.log.info("normal busy and idle are matching with the \
```
sacsant, line 460: Normalised busy plus idle value matches with Frequency (percentage)

ras/ras_ppcutils.py
```diff
@@ -428,6 +433,35 @@ def test_lparstat(self):
...
+                           percentage under frequency")
+        else:
+            self.fail("normal busy and value percentage are not matching \
```
sacsant, line 463: Normalised busy plus idle value does not match with Frequency (percentage)

shirishaganta1: @sacsant As Suggested above changed the sentences and re-pushed it again

sacsant (commented):
Thanks. Looks good to me.

ras/ras_ppcutils.py
```diff
@@ -14,8 +14,10 @@
 # Copyright: 2016 IBM
 # Author: Pavithra <pavrampu@linux.vnet.ibm.com>
 # Author: Sachin Sant <sachinp@linux.ibm.com>
+# Author: Shirisha Ganta <shirisha.ganta1@ibm.com>
```
abdhaleegit, line 17: A common practice is to use vnet id for Author and also for sign-off,  ignore if the id is vnet id

ras/ras_ppcutils.py
```diff
@@ -428,6 +433,33 @@ def test_lparstat(self):
                 self.log.info("%s command passed" % cmd)
                 self.fail("lparstat: Expected failure, %s command exeucted \
                           successfully." % cmd)
+        output = process.system_output("lparstat -E 1 1").decode("utf-8")
+        input_string = ''
+        for line in output.splitlines():
+            if 'GHz' in line:
+                input_string = line
```
abdhaleegit, line 440: cannot I use line directly, any need to assign to a new param ?

abdhaleegit (approved):
Rest all LGTM

ras/ras_ppcutils.py
shirishaganta1: from next Patch onwards will do that as currently I need to do few more steps to get the vnet id..

ras/ras_ppcutils.py
shirishaganta1: yes, we can use the line directly .. Changed the code and re-pushed it again

Naresh-ibm (approved):
@shirishaganta1 Thanks for all changes and PR
@abdhaleegit @PraveenPenguin @sacsant Thanks for reviews and suggestions.
LGTM

PR #2607, opened 2023-04-12
disk_info: support for any device names for input disk

abdhaleegit:
This commit adds support for block device names which persistent accross installs, dlpar, reboot etc.

Block devices names like /dev/sdb are not same across install in some	distros, so disk_info.py currently supports only the the /dev/sdX as	input device, now with the new code changes a user can provide any of sdX, mpathX, dm-0, /dev/mapper/mpathX ,/dev/disk/by-id/scsi-xxx, /dev/disk/by-path/fc-xxx, /dev/dmX for the	below tests

### DIFF
--- a/io/disk/disk_info.py
+++ b/io/disk/disk_info.py
@@ -28,6 +28,7 @@ from avocado.utils import process
 from avocado.utils import genio
 from avocado.utils import distro
 from avocado.utils import multipath
+from avocado.utils import disk
 from avocado.utils.partition import Partition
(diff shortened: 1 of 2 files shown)


abdhaleegit:
[disk_info-devbyid-logs.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/11212976/disk_info-devbyid-logs.txt)

narasimhan-v (commented):
Code changes look good to me.
Will wait for @AYUSHJAIN951 to validate as well

PR #2612, opened 2023-04-19
Fixing self.disk and self.dir parameter

AYUSHJAIN951:
Defaulting self.dir to self.workdir instead of /mnt. 
Cancelling test when self.disk is missing.
Initialising self.err in the beginning.

[log_io_disk_changes.txt](https://github.com/avocado-framework-tests/avocado-misc-tests/files/11268373/log_io_disk_changes.txt)

### DIFF
--- a/io/disk/bonnie.py
+++ b/io/disk/bonnie.py
@@ -53,6 +53,7 @@ class Bonnie(Test):
         Source:
         http://www.coker.com.au/bonnie++/experimental/bonnie++-1.03e.tgz
         """
+        self.err_mesg = []
         self.fstype = self.params.get('fs', default='')
         self.fs_create = False
         lv_needed = self.params.get('lv', default=False)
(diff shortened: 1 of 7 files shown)


AYUSHJAIN951:
With PR https://github.com/avocado-framework-tests/avocado-misc-tests/pull/2574 we understood why having Disk and dir parameter handling needs to be changed, so implementing similar changes in other io/disk files.

Naresh-ibm:
@AYUSHJAIN951 couple of things I want to understand here,

1. what is the exact issue your facing with the existing flow here? The flow looks perfectly fine for me other than getting the absolute_disk_path after checking the disk presence.
But this is also handled in current script but one step later.
2. why did not we pointed out these requirement in our recent review? it would have reduced both your effort and the reviewers effort and time.

The don't think we have complete blockage or critical alarms with the existing script. I don't know if I failing to understand your requirement and perception here.  But I feel your just using the different style of coding.

AYUSHJAIN951:
> 
1. With the current flow when a disk is not provided by the user, tests Errors out
   With such> ERROR: 'Dbench' object has no attribute 'disk'
   Which needed to be handled as i showed with the last PR https://github.com/avocado-framework-tests/avocado-misc-tests/pull/2574 
2. Defaulting use of self.workdir for dir as '/mnt' is not consistent(Your recommendation).
3. **Need for these changes :-** Getting similar code flow in all tests is needed as having different code structure in all tests makes hard to understand, and rework over that in future. 
    So implementing similar flow in most testcases.

narasimhan-v (approved):

Naresh-ibm (approved):
@AYUSHJAIN951 Thanks for PR
@narasimhan-v Thanks for reviews
LGTM

PR #2633, opened 2023-05-11
Improve the smt workload sh script

Geet2:
1. ppc64_cpu --smt=8 and ppc64_cpu --smt=on mean the same thing. So there is no point running these commands one after another.

2. added ppc64_cpu --smt=on followed by ppc64_cpu --smt=off so that we have an instance where maximum no. of threads in a core (7) go offline at once.


PraveenPenguin:
@SamirMulani can you please review this PR

SamirMulani:
> @SamirMulani can you please review this PR

@PraveenPenguin LGTM

PR #2643, opened 2023-05-18
xfstests: Implement many missing features and few cleanups

riteshharjani:
- Added support for collecting results after the run
- Added support for "args" param which can inturn give all options support by xfstests i.e. -i/-I for iteration, -e for exclude -R for xunit result.xml output.
- Added support for multiple fstests config e.g. 1k, 4k, 64k adv, with quick and auto group tests. 
- Added support for btrfs scratch_dev_pool 
- Cleaned up not so useful arguments like "group", "test_range", skip_dangerous
- Removed group file support as it is not supported by xfstests anymore
- In the end added a stressng dir mount point which I noticed while running stress-ng-fs.yaml.


fs/xfstests.py
```diff
@@ -444,36 +441,30 @@ def setUp(self):
...
+        srcdir = f"{self.teststmpdir}/results"
+
+        job_dir = os.path.dirname(os.path.dirname(self.logdir))
+        self.job_id = os.path.basename(job_dir)
+
+        self.log.debug(" Job ID: %s, logdir: %s, srcdir: %s, outputdir: %s: " %
+                       (self.job_id, self.logdir, srcdir, self.outputdir))
+
+        new_outputdir = os.path.join(self.outputdir, os.path.basename(srcdir))
+        if (os.path.exists(self.outputdir)):
```
disgoel, line 465: this condition fails when src results dir is not present. eg- running btrfs on rhel
we should check if srcdir is present or not before copying
`if os.path.exists(srcdir):`

fs/xfstests.py.data/xfstests.yaml
```diff
@@ -1,41 +1,41 @@
...
     type: 'loop'
-    loop_size: '7GiB'
+    loop_size: '5GiB'
     # Option to provide disk for loop device creation,
     # Uses '/' by default for file creation
     disk: "null"
 fs_type: !mux
     fs_ext4_4k:
         fs: 'ext4'
+        args: '-e ext4/048 -R xunit generic/001'
```
disgoel, line 14: default args value needs to be changed to quick/auto group at all places here

fs/xfstests.py.data/xfstests_disk.yaml
```diff
@@ -1,18 +1,26 @@
-skip_dangerous: True
 scratch_mnt: '/mnt/scratch'
 test_mnt: '/mnt/test'
-group:
-# edit test_range for running specific tests
-test_range:
 fs_type: !mux
     fs_ext4:
         fs: 'ext4'
+        args: '-e ext4/048 -R xunit generic/001'
```
disgoel, line 6: same here

disgoel (changes requested):

fs/xfstests.py
```diff
@@ -512,89 +503,35 @@ def tearDown(self):
...
-                            sudo=True)
-            else:
-                process.run('fallocate -o 0 -l %s %s/file-%s.img' %
-                            (loop_size, self.disk_mnt, i), shell=True,
-                            sudo=True)
+
+        # remove any previous losetup images & mounts
+        process.system('umount %s %s' % (self.scratch_mnt, self.test_mnt),
+                       sudo=True, ignore_status=True)
+        process.run('losetup -D')
```
AYUSHJAIN951, line 526: why are you removing all loop devices in system attached which are out of context from this test.
as `losetup -D` ->`Detach all associated loop devices.`

AYUSHJAIN951 (changes requested):

fs/xfstests.py
AYUSHJAIN951: better will be to check what next loop devices are available and only be using those.

fs/xfstests.py
disgoel: when we run btrfs on rhel, it should cancel the test and run teardown. But in teardown since results dir is not present it throws error.
```
 (5/6) xfstests.py:Xfstests.test;run-fs_type-fs_btrfs_4k-loop_type-b9dc: ERROR: [Errno 2] No such file or directory: '/var/tmp/avocado_em29zjl5/results' (0.75 s)
 (6/6) xfstests.py:Xfstests.test;run-fs_type-fs_btrfs_64k-loop_type-ab81: ERROR: [Errno 2] No such file or directory: '/var/tmp/avocado_em29zjl5/results' (0.75 s)
```

fs/xfstests.py
riteshharjani: Sometimes when the test fails... it never gets to run the tearDown.. 
Hence I wanted to keep losetup -D.  I mostly see this problem while development, so anyways let me drop this patch.. 
I will debug the tearDown problem later.

fs/xfstests.py
```diff
@@ -323,19 +328,22 @@ def setUp(self):
             self.cancel('Unknown filesystem %s' % self.fs_to_test)
         mount = True
         self.log_devices = []
+
+        # For btrfs we need minimum of 5 loop devices for SCRATCH_DEV_POOL
+        self.num_loops = 2
```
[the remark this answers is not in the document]
riteshharjani: ok.

fs/xfstests.py
riteshharjani: ok got it.. So this happens during error scenario.. I will fix it in the next rev.

riteshharjani:
@AYUSHJAIN951 @disgoel @PraveenPenguin 
I have address the review comments and have opened another v2 pull request here [1]. 
Hence closing this one.

[1]: https://github.com/avocado-framework-tests/avocado-misc-tests/pull/2644

PR #2659, opened 2023-06-09
Adding new subtest to create n number of namespaces

Naresh-ibm:
current test cases doesn't has any test that can create with given count so adding same to validate.

### DIFF
--- a/io/disk/ssd/nvmetest.py
+++ b/io/disk/ssd/nvmetest.py
@@ -26,6 +26,7 @@ from avocado.utils import process
 from avocado.utils import archive
 from avocado.utils import build
 from avocado.utils import download
+from avocado.utils import nvme
 from avocado.utils.software_manager.manager import SoftwareManager
 
 
@@ -44,6 +45,7 @@ class NVMeTest(Test):
         """
         self.device = self.params.get('device', default='nvme0')
         self.device = "/dev/%s" % self.device
+        self.ns_count = self.params.get('namespace_count', default=1)
         cmd = 'ls %s' % self.device
         if process.system(cmd, ignore_status=True):
             self.cancel("%s does not exist" % self.device)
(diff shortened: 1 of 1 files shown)


io/disk/ssd/nvmetest.py
```diff
@@ -44,6 +45,7 @@ def setUp(self):
         """
         self.device = self.params.get('device', default='nvme0')
         self.device = "/dev/%s" % self.device
+        self.ns_count = self.params.get('namespace_count', default=1)
```
PraveenPenguin, line 48: can  this be moved in test itself and having variable scop local

Naresh-ibm: @PraveenPenguin we will be needing this in other test too in future, as complete test structure will be going to change going forward.
So I feel having in setup will help better.

io/disk/ssd/nvmetest.py
```diff
@@ -431,3 +433,10 @@ def testsubsystemreset(self):
         cmd = '%s subsystem-reset %s' % (self.binary, self.device)
         if process.system(cmd, ignore_status=True, shell=True):
             self.fail("Subsystem reset failed")
+
+    def test_create_namespaces(self):
+        """
+        creates the specified number of namespaces on nvme drive
+        """
+        device = self.device.split("/")[-1]
+        nvme.create_n_namespaces(device, self.ns_count)
```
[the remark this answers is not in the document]
Naresh-ibm: @abdhaleegit 
1. delete part is taken care in utils
2. util will error out if creation of any namespace fails

so no extra validation is required here

io/disk/ssd/nvmetest.py
PraveenPenguin: no, still I feel whenever requirement we should ACT not on a futuristic patch  :) , current one will be based on today's requirement

io/disk/ssd/nvmetest.py
abdhaleegit, line 442: @Naresh-ibm please  rename the create_n_namespace to proper one which is in utils

abdhaleegit:
@Naresh-ibm looks like merge conflicts

abdhaleegit (changes requested):
@Naresh-ibm we need one more test to delete the namesapces from cfg

please add test_delete_namspaces()

PR #2683, opened 2023-07-28
Per data and vmstat comparison with different stress-ng CPU workload.

manvanthar:
Per data and vmstat comparison with different stress-ng CPU workload.

Test runs different CPU workload and tries to capture the perf data and vmstat with multiple iterations.  Average is taken to compare if the test PASS/FAIL.


manvanthar:
[](url)
[job.log](https://github.com/avocado-framework-tests/avocado-misc-tests/files/12197662/job.log)

manvanthar:
Raised new PR

PR #2684, opened 2023-07-31
Perf data and vmstat comparison with different stress-ng CPU workload

manvanthar:
Perf data and vmstat comparison with different stress-ng CPU workload

Test runs different CPU workload and tries to capture the perf data and vmstat with multiple iterations. Average is taken to compare if the test PASS/FAIL.


manvanthar:
travisci broken . resending PR

PR #2697, opened 2023-08-31
bonding: Add support for mac base input interfaces

abdhaleegit:
In a contineous test environement setups where device names are not persistent accross os, having a unique key like mac addr as yaml input which does not change across reboots, dlpar or os install with different linux flavours, also the legacy interface name still works.


io/net/bonding.py.data/README.txt
```diff
@@ -31,5 +31,3 @@ Requirements:
 -----------------------
 1. install netifaces using pip
 command: pip install netifaces
-2. Generate sshkey for your test partner to run the test uninterrupted.(Have a passwordless ssh between the peers)
-3. Make sure IPs are set for interfaces to be used, via configuration file. ifup / ifdown should set the IPs back.
```
PraveenPenguin, line 35: why remove this info are we doing this in the test?

io/net/bonding.py.data/README.txt
```diff
@@ -16,7 +16,7 @@ Note:
 -----------------------------
 Inputs Needed To Run Tests:
 ------------------------------
-bond_interfaces --> Interfaces in the Host machine requird for Bonding
+bond_interfaces --> Interfaces names or mac address space separated in the Host machine requird for Bonding
```
PraveenPenguin, line 19: can you populate example string as well

io/net/bonding.py
```diff
@@ -579,4 +585,5 @@ def error_check(self):
             self.fail("Tests failed. Details:\n%s" % "\n".join(self.err))
 
     def tearDown(self):
-        self.session.quit()
+        if self.host_interfaces:
+            self.session.quit()
```
PraveenPenguin, line 589: why this change is this was bug earlier , if yes push this in another commit

PraveenPenguin (changes requested):

Naresh-ibm:
@abdhaleegit are we pursuing this?

io/net/bonding.py.data/README.txt
abdhaleegit: both the requirements are already automated into script.. so they are not needed

io/net/bonding.py.data/README.txt
abdhaleegit: sure

io/net/bonding.py
abdhaleegit: this is to safely exit  teardown.. if interfaces are not available.. or wrong user input

abdhaleegit:
> @abdhaleegit are we pursuing this?

yes will pursue it... I will update with latest code

PraveenPenguin:
> > @abdhaleegit are we pursuing this?
> 
> yes will pursue it... I will update with latest code

Gentle reminder here ,  if not pursuing in this PR , please close

PR #2712, opened 2023-09-25
interfaces: use common name accross all tests

abdhaleegit:
To simplify and avoid extra parameters, keep a common input parameter name across all network test i.e rename host_interface to just interfaces


abdhaleegit (commented):
LGTM

PR #2725, opened 2023-10-20
fix parsing of metric events

disgoel:
With current logic it was not able to run events which contains lowercase char.
For example: General, Instruction_Misses, Memory_BW, No_group, Others, Translation, dL1_Reloads

Fix this logic by using --raw-dump to collect metric/metricgroup events which also eliminates using if condition.


abdhaleegit (approved):
LGTM

PR #2726, opened 2023-10-30
Correct software_manager import

nasastry:
Correct software_manager import in fscrypt and fsverity tests.


Naresh-ibm (approved):
@nasastry Thanks for this fix.
LGTM

PR #2732, opened 2023-11-15
io/iommu: Add test to validate paging level in IOMMU v2 mode

dhsrivas:
IOMMU in v2 mode should use the same level of paging as host page table. For e.g. if the host uses 5-level of paging and IOMMU is using v2 page table then IOMMU should also use 5-level paging for v2 page table.

### DIFF
--- a/io/iommu/amd/iommu_v2pgmode_test.py
+++ b/io/iommu/amd/iommu_v2pgmode_test.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2023 Advanced Micro Devices, Inc.
+# Author: Dheeraj Kumar Srivastava <dheerajkumar.srivastava@amd.com>
+
+"""
+Validate 5 level page table support in v2 iommu page table mode
+"""
+
+from avocado import Test
+from avocado import skipUnless
+from avocado.utils import cpu, process, linux_modules
+
+
+def check_kernelconf_5lvl():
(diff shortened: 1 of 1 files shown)


narasimhan-v (approved):

io/iommu/amd/iommu_v2pgmode_test.py
```diff
@@ -0,0 +1,104 @@
...
+
+def get_va_bits():
+    '''
+    Check for VA address bit size in /proc/cpuinfo
+    '''
+    cpu_info = genio.read_file("/proc/cpuinfo")
+    for line in cpu_info.splitlines():
+        if "address sizes" in line:
+            return line.split()[-3].strip()
+    return ""
```
PraveenPenguin, line 34: can we push this as a utility  in avocado

io/iommu/amd/iommu_v2pgmode_test.py
```diff
@@ -0,0 +1,104 @@
...
+
+def check_dmesg(string):
+    """
+    Check dmesg for 'string' in dmesg
+    """
+    cmd = f'dmesg | grep -i "{string}"'
+    output = process.run(cmd, ignore_status=True, shell=True).stdout_text
+    if output != "":
+        return True
+    return False
```
PraveenPenguin, line 56: dmsg utility won't help? can you please check this

PraveenPenguin (changes requested):

io/iommu/amd/iommu_v2pgmode_test.py
dhsrivas: Hi,
I should use this 
'''
def collect_errors_dmesg(patterns):
    """Check patterns in dmesg.

    :param patterns : list variable to search in dmesg
    :returns: error log in form of list
    :rtype: list of str
    """
'''
but didn't due to the naming of the function. Its good if we can make it something like "collect_dmesg_by_patterns"

io/iommu/amd/iommu_v2pgmode_test.py
dhsrivas: Sure i will send patch for this util to avocado.

dhsrivas: Done. I have updated the patch. Please review.

dhsrivas:
@PraveenPenguin Hi, Is there anything to be done?

PR #2744, opened 2023-12-21
perf/pgbench: Introduce pgbench benchmark

gautammenghani:
Introduce the pgbench benchmark to enable performance benchmarking of postgres databases. This will be useful in identifying performance regressions quickly.

> [root@localhost perf]# avocado run pgbench.py -m pgbench.py.data/pgbench.yaml
JOB ID     : c0fcdc9b3a30b93249b9dc6cdb07a681375cf394                                                                         
JOB LOG    : /root/avocado/job-results/job-2023-12-21T06.56-c0fcdc9/job.log                                                   
(1/1) pgbench.py:Pgbench.test;run-transaction_count-f854: STARTED                                                                      
(1/1) pgbench.py:Pgbench.test;run-transaction_count-f854: PASS (82.17 s)                                                               
JOB TIME   : 84.30 s

### DIFF
--- a/perf/pgbench.py
+++ b/perf/pgbench.py
@@ -35,21 +35,20 @@ class Pgbench(Test):
     already_installed = True
 
     def run_cmd(self, cmd, ignore_failure):
-        command = process.SubProcess(cmd)
-        command.run()
+        command = process.run(cmd, ignore_status=True)
 
-        if command.result.exit_status and ignore_failure:
-            self.log.info(f"Command '{cmd}' returned '{command.result.exit_status}' "
-                          f"with the text '{command.result.stderr_text}'")
-            return [command.result.exit_status, command.result.stderr_text]
-        elif command.result.exit_status:
-            self.cancel(f"Command '{cmd}' returned '{command.result.exit_status}' "
-                        f"with the text '{command.result.stderr_text}'")
+        if command.exit_status and ignore_failure:
+            self.log.info(f"Command '{cmd}' returned '{command.exit_status}' "
+                          f"with the text '{command.stderr_text}'")
+            return [command.exit_status, command.stderr_text]
+        elif command.exit_status:
+            self.cancel(f"Command '{cmd}' returned '{command.exit_status}' "
+                        f"with the text '{command.stderr_text}'")
         else:
-            return [command.result.exit_status, command.result.stdout_text]
+            return [command.exit_status, command.stdout_text]
 
     def get_pkg_manager(self):
-        for pkg_mgr in ['apt', 'dnf', 'yum']:
+        for pkg_mgr in ['apt', 'dnf', 'yum', 'zypper']:
             if self.sm.check_installed(pkg_mgr):
                 return pkg_mgr
         return None
--- a/perf/pgbench.py
+++ b/perf/pgbench.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2023 IBM
+# Author: Gautam Menghani <gautam@linux.ibm.com>
+
+import os
+
+from avocado import Test
+from avocado.utils import process, cpu
+from avocado.utils.service import ServiceManager
+from avocado.utils.software_manager.manager import SoftwareManager
+
+
+class Pgbench(Test):
+
+    """
+    This module will run the pgbench benchmark. Pgbench is a benchmark for
+    measuring the throughput achieved by executing database transactions on
+    PostgresSQL database.
+    """
+    sm = SoftwareManager()
+    scaling_factor = worker_threads = db_clients = protocol = \
+        benchmark_duration = transaction_count = 0
+    already_installed = True
+
+    def run_cmd(self, cmd, ignore_failure):
+        command = process.SubProcess(cmd)
+        command.run()
+
+        if command.result.exit_status and ignore_failure:
+            self.log.info(f"Command '{cmd}' returned '{command.result.exit_status}' "
+                          f"with the text '{command.result.stderr_text}'")
+            return [command.result.exit_status, command.result.stderr_text]
+        elif command.result.exit_status:
+            self.cancel(f"Command '{cmd}' returned '{command.result.exit_status}' "
+                        f"with the text '{command.result.stderr_text}'")
+        else:
+            return [command.result.exit_status, command.result.stdout_text]
+
+    def get_pkg_manager(self):
+        for pkg_mgr in ['apt', 'dnf', 'yum']:
+            if self.sm.check_installed(pkg_mgr):
+                return pkg_mgr
+        return None
+
+    def install_and_initialize_postgres(self, pkg_mgr):
+        packages = ["postgresql", "postgresql-contrib"]
+        for package in packages:
+            self.sm.install(package)
+
+        # rhel needs an extra pkg and initialization
+        if pkg_mgr in ['dnf', 'yum']:
+            self.sm.install('postgresql-server')
+            output = self.run_cmd("postgresql-setup --initdb", True)
+            if output[0] and "not empty" not in output[1]:
+                self.cancel(f"Failed to initialize postgres - {output[1]}")
+
+    def setUp(self):
+        '''
+        Install Pgbench
+        '''
+        # check root privileges ($USER should be able to run commands
+        # as another user without requiring a password)
+        if os.geteuid() != 0:
+            self.cancel("This script requires root privileges, "
+                        "Please try again with 'sudo' or as 'root'")
+
+        # check if pgbench is installed
+        if (self.sm.check_installed("pgbench")):
+            self.log.info("Pgbench is already installed")
+        else:
+            pkg_mgr = self.get_pkg_manager()
+            if pkg_mgr:
+                self.install_and_initialize_postgres(pkg_mgr)
+                self.already_installed = False
+            else:
+                self.cancel("Unsupported Linux distribution")
+
+        # restart postgres service
+        ManagerService = ServiceManager()
+        ManagerService.restart("postgresql")
+
+        # add current user to db
+        output = self.run_cmd(f"sudo -u postgres createuser --superuser {os.getlogin()}", True)
+        if output[0] and "already exists" not in output[1]:
+            self.cancel(f"Could not create role {os.getlogin()}")
+
+        # Setup the params
+        self.scaling_factor = self.params.get("scaling_factor", default=100)
+        self.worker_threads = self.params.get("worker_threads", default=cpu.online_count())
+        self.db_clients = self.params.get("db_clients", default=cpu.online_count())
+        self.protocol = self.params.get("protocol", default="prepared")
+        self.benchmark_duration = self.params.get("benchmark_duration", default=0)
+        self.transaction_count = self.params.get("transaction_count", default=0)
+
+        # validation: pgbench cannot accept both transaction count and duration
+        if self.benchmark_duration > 0 and self.transaction_count > 0:
+            self.log.warn("Pgbench cannot accept both max transaction count and duration of the test, proceeding with specified time limit")
+        elif self.benchmark_duration == 0 and self.transaction_count == 0:
+            self.log.warn("Transaction count / duration of the test not specified, defaulting to time limit of 120 secs")
+            self.benchmark_duration = 120
+
+        # create the db
+        self.run_cmd("createdb pgbench", False)
+
+        # setup tables
+        self.run_cmd(f"pgbench -i -s {self.scaling_factor} -n pgbench", False)
+
+    def test(self):
+
+        # run the benchmark
+        if self.benchmark_duration > 0:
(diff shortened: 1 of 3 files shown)


perf/pgbench.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # validation: pgbench cannot accept both transaction count and duration
+        if benchmark_duration > 0 and transaction_count > 0:
+            self.log.warn("Pgbench cannot accept both max transaction count and duration of the test, proceeding with specified time limit")
+        elif benchmark_duration == 0 and transaction_count == 0:
+            self.log.warn("Transaction count / duration of the test not specified, defaulting to time limit of 120 secs")
+            benchmark_duration = 120
+
+        # create the db
+        self.run_cmd("createdb pgbench", False)
```
vaibhav92, line 111: I think the steps to create and initalize the pgbench db are better done in setUp() rather than run_cmd()

perf/pgbench.py
```diff
@@ -0,0 +1,127 @@
...
+        # Setup the params
+        scaling_factor = self.params.get("scaling_factor", default=100)
+        worker_threads = self.params.get("worker_threads", default=cpu.online_count())
+        db_clients = self.params.get("db_clients", default=cpu.online_count())
+        protocol = self.params.get("protocol", default="prepared")
+        benchmark_duration = self.params.get("benchmark_duration", default=0)
+        transaction_count = self.params.get("transaction_count", default=0)
+
+        # validation: pgbench cannot accept both transaction count and duration
+        if benchmark_duration > 0 and transaction_count > 0:
```
vaibhav92, line 104: These validations too can be part of setUp()

perf/pgbench.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def install_packages(self, pkg_mgr):
+        packages = ["postgresql", "postgresql-contrib"]
+        for package in packages:
+            self.sm.install(package)
+
+        # rhel needs an extra pkg
+        if pkg_mgr in ['dnf', 'yum']:
+            dnf_pkgs = ["postgresql-server"]
+            for package in dnf_pkgs:
```
vaibhav92, line 61: dnf_pkgs has just one element. do you really need this loop

perf/pgbench.py
```diff
@@ -0,0 +1,127 @@
...
+    This module will run the pgbench benchmark. Pgbench is a benchmark for
+    measuring the throughput achieved by executing database transactions on
+    PostgresSQL database.
+    """
+    sm = SoftwareManager()
+
+    def run_cmd(self, cmd, ignore_failure):
+        command = process.SubProcess(cmd)
+        command.run()
+        if command.result.exit_status:
```
vaibhav92, line 37: this if else ladder can be simplified

if ignore_failure:
   return [command.result.exit_status]
elif command.result.exit_status:
   self.cancel()
else:
   return [exit_status..]

perf/pgbench.py
```diff
@@ -0,0 +1,127 @@
...
+            self.log.warn("Pgbench cannot accept both max transaction count and duration of the test, proceeding with specified time limit")
+        elif benchmark_duration == 0 and transaction_count == 0:
+            self.log.warn("Transaction count / duration of the test not specified, defaulting to time limit of 120 secs")
+            benchmark_duration = 120
+
+        # create the db
+        self.run_cmd("createdb pgbench", False)
+
+        # setup tables
+        self.run_cmd(f"pgbench -i -s {scaling_factor} -n pgbench", False)
```
vaibhav92, line 114: You shouldnt ignore errors for these command runs

perf/pgbench.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # run the benchmark
+        if benchmark_duration > 0:
+            result = self.run_cmd(f"pgbench --protocol={protocol} --jobs={worker_threads} --scale={scaling_factor} --client={db_clients}  -n --time={benchmark_duration} -r pgbench", False)
+        else:
+            result = self.run_cmd(f"pgbench --protocol={protocol} --jobs={worker_threads} --scale={scaling_factor} --client={db_clients}  -n --transactions={transaction_count} -r pgbench", False)
+        self.log.info("===== Pgbench benchmark results ======")
+        if not result[0]:
+            self.log.info(result[1])
```
vaibhav92, line 123: you should fail the test if the pgbench command run returns an error

perf/pgbench.py
gautammenghani: yes I am specifying `ignore_failure` as False. So if there is an error , run_cmd() will just cancel the test. 
we can't proceed from this point if there is an error in this command

perf/pgbench.py
gautammenghani: yes similar to previous comment, I am specifying `ignore_failure` as False. So if there is an error , run_cmd() will just cancel the test. we can't proceed from this point if there is an error in this command

gautammenghani:
Thanks for the comments, I've made the changes
Please review and let me know if anything else is needed

perf/pgbench.py
```diff
@@ -0,0 +1,128 @@
...
+        for package in packages:
+            self.sm.install(package)
+
+        # rhel needs an extra pkg
+        if pkg_mgr in ['dnf', 'yum']:
+            self.sm.install('postgresql-server')
+
+            # restart postgres service
+            ManagerService = ServiceManager()
+            ManagerService.restart("postgresql")
```
vaibhav92, line 67: Is the indentation of these 2 statements intentional ? Do you only want to restart the postgres of you installed it via dnf/yum.

perf/pgbench.py
```diff
@@ -0,0 +1,128 @@
...
+        return None
+
+    def install_packages(self, pkg_mgr):
+        packages = ["postgresql", "postgresql-contrib"]
+        for package in packages:
+            self.sm.install(package)
+
+        # rhel needs an extra pkg
+        if pkg_mgr in ['dnf', 'yum']:
+            self.sm.install('postgresql-server')
```
vaibhav92, line 63: Do you want to exit the loop if one of the command succeeds ?

perf/pgbench.py
```diff
@@ -0,0 +1,128 @@
...
+        Install Pgbench
+        '''
+        # check root privileges ($USER should be able to run commands
+        # as another user without requiring a password)
+        if os.geteuid() != 0:
+            self.cancel("This script requires root privileges, "
+                        "Please try again with 'sudo' or as 'root'")
+
+        # check if pgbench is installed
+        if (self.sm.check_installed("pgbench")):
```
vaibhav92, line 80: shouldnt you restart pgbench or ensure that its running if the pgbench package is already installed

perf/pgbench.py
```diff
@@ -0,0 +1,128 @@
...
+            result = self.run_cmd(f"pgbench --protocol={self.protocol} --jobs={self.worker_threads} --scale={self.scaling_factor} --client={self.db_clients}  -n --time={self.benchmark_duration} -r pgbench", False)
+        else:
+            result = self.run_cmd(f"pgbench --protocol={self.protocol} --jobs={self.worker_threads} --scale={self.scaling_factor} --client={self.db_clients}  -n --transactions={self.transaction_count} -r pgbench", False)
+        self.log.info("===== Pgbench benchmark results ======")
+        if not result[0]:
+            self.log.info(result[1])
+
+    def tearDown(self):
+        # destory the db
+        self.run_cmd("dropdb pgbench", True)
```
vaibhav92, line 128: Low Priority: I prefer tests to be idempotent, So _ideally_ you should also remove the packages you installed during tearDown()

perf/pgbench.py
vaibhav92: Acked.. please ignore the review comment

perf/pgbench.py
gautammenghani: Which loop are you referring to? we are not iterating over the pkg mgrs here

perf/pgbench.py
gautammenghani: Yes this was intentional, but I've generalized this code to restart the service regardless of whether pgbench was already there or not

gautammenghani:
Thanks for the comments, I've made the necessary changes

perf/pgbench.py
vaibhav92: Argh, vacation brain did 's/if/for' .. please ignore this comment

vaibhav92 (approved):
In its current state this PR LGTM

abdhaleegit (approved):
@gautammenghani Looks good to me, can you confirm once if this works fine for SUSE distro

vaibhav92 (approved):

perf/pgbench.py
```diff
@@ -0,0 +1,143 @@
...
+        # check if pgbench is installed
+        if (self.sm.check_installed("pgbench")):
+            self.log.info("Pgbench is already installed")
+        else:
+            pkg_mgr = self.get_pkg_manager()
+            if pkg_mgr:
+                self.install_and_initialize_postgres(pkg_mgr)
+                self.already_installed = False
+            else:
+                self.cancel("Unsupported Linux distribution")
```
PraveenPenguin, line 88: instead of the package manager, we should be having checks based on the distro name

perf/pgbench.py
```diff
@@ -0,0 +1,143 @@
...
+    PostgresSQL database.
+    """
+    sm = SoftwareManager()
+    scaling_factor = worker_threads = db_clients = protocol = \
+        benchmark_duration = transaction_count = 0
+    already_installed = True
+
+    def run_cmd(self, cmd, ignore_failure):
+        command = process.SubProcess(cmd)
+        command.run()
```
PraveenPenguin, line 39: why not process.run () ?

perf/pgbench.py
```diff
@@ -0,0 +1,143 @@
...
+                self.already_installed = False
+            else:
+                self.cancel("Unsupported Linux distribution")
+
+        # restart postgres service
+        ManagerService = ServiceManager()
+        ManagerService.restart("postgresql")
+
+        # add current user to db
+        output = self.run_cmd(f"sudo -u postgres createuser --superuser {os.getlogin()}", True)
```
PraveenPenguin, line 95: process run has an argument as run as `sudo=True` that will help to run any command on root privileged user

perf/pgbench.py
```diff
@@ -0,0 +1,143 @@
...
+
+        # reset things if we set them up
+        if not self.already_installed:
+            packages = ["postgresql", "postgresql-contrib"]
+            for package in packages:
+                self.sm.remove(package)
+
+            pkg_mgr = self.get_pkg_manager()
+            if pkg_mgr in ['dnf', 'yum']:
+                self.sm.remove('postgresql-server')
```
PraveenPenguin, line 143: here also please check based on distro name

perf/pgbench.py.data/README
```diff
@@ -0,0 +1,11 @@
...
+
+Inputs in yaml file
+--------------------
+1. scaling_factor : Factor by which we multiply the number of rows generated.
+2. transaction_count: Number of transactions each client runs.
+3. benchmark_duration: Run the benchmark for specified number of seconds.
+4. worker_threads: Number of worker threads spawned by pgbench.
+5. db_clients: Number of clients / concurrent database connections created.
+6. protocol : Protocol to use for submitting queries to the server (simple/extended/prepared).
+
```
PraveenPenguin, line 11: please remove this whitespace

PraveenPenguin (changes requested):

perf/pgbench.py
gautammenghani: we used package manger so that we don't have to limit the script to a few distros like debian/ubuntu. If the user is using a debian derivative like ubuntu/pop os/linux mint, we don't have to hardcode all those distros as the package manager will be the same; so we cover more ground if we detect package manager

gautammenghani: as per abdul's comment above, I am adding support for suse as well

PraveenPenguin: sure , some how not liking taking decision based on package manger

perf/pgbench.py
gautammenghani: Please not that this command has to be run as postgres user. So if we just do `sudo=True` it won't work

perf/pgbench.py
gautammenghani: ok can this break in any scenario? Because if we use distro names and just specify "debian,ubuntu,rhel,fedora,suse",  a user running this on a debian derivative will get "unsupported distro" even if the benchmark can run just fine.

gautammenghani:
@abdhaleegit This script now works on suse as well

```
localhost:~/avocado-misc-tests/perf # cat /etc/os-release
NAME="openSUSE Tumbleweed"
# VERSION="20231219"
ID="opensuse-tumbleweed"
ID_LIKE="opensuse suse"
VERSION_ID="20231219"
PRETTY_NAME="openSUSE Tumbleweed"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:tumbleweed:20231219"
BUG_REPORT_URL="https://bugzilla.opensuse.org"
SUPPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org"
DOCUMENTATION_URL="https://en.opensuse.org/Portal:Tumbleweed"
LOGO="distributor-logo-Tumbleweed"
localhost:~/avocado-misc-tests/perf #
localhost:~/avocado-misc-tests/perf # avocado run pgbench.py -m pgbench.py.data/pgbench.yaml
JOB ID     : dfdc7bedc6d30b1b9b54c90c258b4d2e2815d67a
JOB LOG    : /root/avocado/job-results/job-2024-01-10T07.03-dfdc7be/job.log
(1/1) pgbench.py:Pgbench.test;run-transaction_count-f854: STARTED
(1/1) pgbench.py:Pgbench.test;run-transaction_count-f854:  PASS (26.21 s)
JOB TIME   : 27.16 s
```

perf/pgbench.py
gautammenghani: or is it that we want to implicitly imply that we have tested this to work on these 4-5 distros? Please advise on this

PraveenPenguin: @gautammenghani ideally where we tested we should have supported if any user is looking supported in the next distro we can welcome a patch

PraveenPenguin (approved):

PR #2745, opened 2023-12-27
Drop executable bits and shebangs

clebergnu:
The files containing Avocado instrumented tests have no reason for having executable bits, or having a Python shebang, as they will only be run under the "avocado run" or similar Avocado Job API environment.

Let's remove all executabl ebits and shebangs, to simplify boilerplate code and influence new tests files to not use them.


cpu/dwh.py
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 #
```
PraveenPenguin, line 1: can you please remove `#` as well and where it applicable

PraveenPenguin:
@clebergnu gentle ping

PR #2774, opened 2024-03-15
kernel/will-it-scale: Add scalability test

sacsant:
This patch adds a test to validate scalability aspect of kernel. Will It Scale
takes a testcase and runs n parallel copies to see if the testcase will scale. 
This test is based on https://github.com/antonblanchard/will-it-scale

The test includes yaml based input for each supported testcase.

### DIFF
--- a/kernel/will-it-scale.py
+++ b/kernel/will-it-scale.py
@@ -1,0 +1,95 @@
+#!/usr/bin/env python
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2024 IBM
+# Author: Sachin Sant <sachinp@linux.ibm.com>
+#
+
+
+import os
+import shutil
+from sys import version_info
+
+from avocado import Test
+from avocado import skipIf
+from avocado.utils import process, build, memory, archive
+from avocado.utils.software_manager.manager import SoftwareManager
+
+SINGLE_NODE = len(memory.numa_nodes_with_memory()) < 2
+VERSION_CHK = version_info[0] < 4 and version_info[1] < 7
+
+
+class WillItScaleTest(Test):
+    """
+    Will It Scale takes a testcase and runs n parallel copies to see if the
+    testcase will scale.
+    Source - https://github.com/antonblanchard/will-it-scale
+
+    :avocado: tags=kernel,ppc64le
+    """
+
+    @skipIf(SINGLE_NODE, "Test requires atleast two numa nodes")
+    @skipIf(VERSION_CHK, "Test requires Python 3.7+")
+    def setUp(self):
+        """
+        To execute test using git copy
+          make
+          ./runalltests
+
+        To generate graphical results
+          ./postprocess.py
+        """
+        smm = SoftwareManager()
+        for package in ['gcc', 'make', 'hwloc-devel']:
+            if not smm.check_installed(package) and not smm.install(package):
+                self.cancel('%s is needed for the test to be run' % package)
+
+        self.postprocess = self.params.get('postprocess', default=True)
+        self.testcase = self.params.get('name', default='brk1')
+        url = self.params.get(
+                'willit_url', default='https://github.com/antonblanchard/'
+                'will-it-scale/archive/refs/heads/master.zip')
+        tarball = self.fetch_asset('willit.zip', locations=[url],
+                                   expire='7d')
+        archive.extract(tarball, self.workdir)
+        self.sourcedir = os.path.join(self.workdir, 'will-it-scale-master')
+        os.chdir(self.sourcedir)
+        if build.make(self.sourcedir):
+            self.fail('make failed, please check debug logs')
+
+    def test_scaleitall(self):
+        """
+        Invoke and execute test(s)
+        """
+        os.chdir(self.sourcedir)
+        self.log.info("Starting test...")
+
+        # Identify the test to be executed
+        if self.testcase in 'All':
+            cmd = './runalltests'
+        else:
+            cmd = './runtest.py %s > %s.csv' % (self.testcase, self.testcase)
+
+        # Execute the test(s)
+        if process.system(cmd, shell=True, sudo=True, ignore_status=True) != 0:
+            self.fail('Please check the logs for failure')
+        if self.testcase not in 'All':
+            shutil.copy(f"{self.testcase}.csv", self.logdir)
+
+        # Generate graphical results if postprocessing is enabled
+        if self.postprocess:
+            cmd = './postprocess.py'
+            if process.system(cmd, shell=True, ignore_status=True) != 0:
+                self.warn('Post processing failed, graph may not be generated')
+        if self.testcase not in 'All':
+            shutil.copy(f"{self.testcase}.html", self.logdir)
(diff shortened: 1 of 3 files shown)


PraveenPenguin:
@sacsant can see scope to merge all yaml in single YAML as it will reduce maintenance effort

abdhaleegit:
> @sacsant can see scope to merge all yaml in single YAML as it will reduce maintenance effort

single yaml with all test mux variants

sacsant:
Depending on resources available on system, each test takes anywhere between 40min to a couple of hours. I don't think we should run the test against all option available as it will take maybe more than 24 hours.

I can include a couple of sample YAML files and remove the remaining one. The README file has information about all possible tests. A user who wants to run a specific test can always create their own yaml's.  

Would that work?

sacsant:
I have consolidated the YAML's into a single one using mux and have added a warning about long execution time.

sacsant:
Test result 
 (1/1) kernel/will-it-scale.py:WillItScaleTest.test_scaleitall;run-testcase-mmap1-f366: STARTED
 (1/1) kernel/will-it-scale.py:WillItScaleTest.test_scaleitall;run-testcase-mmap1-f366: PASS (1590.29 s)

kernel/will-it-scale.py.data/will-it-scale.yaml
```diff
@@ -0,0 +1,188 @@
...
+#testcase: !mux
+#    All:
+#        name: All
+#        postprocess: True
+# Each testcase can take upto an hour or so depending on amount
+# of resources available on the system under test
+testcase: !mux
+    brk1:
+        name: brk1
+        postprocess: True
```
abdhaleegit, line 11: remove  postprocess: True  in each variant and keep it above testcase: !mux  makes it available for each variant

abdhaleegit (changes requested):

kernel/will-it-scale.py
```diff
@@ -0,0 +1,100 @@
...
+                                   expire='7d')
+        archive.extract(tarball, self.workdir)
+        self.sourcedir = os.path.join(self.workdir, 'will-it-scale-master')
+        os.chdir(self.sourcedir)
+        if build.make(self.sourcedir):
+            self.fail('make failed, please check debug logs')
+
+    @skipIf(SINGLE_NODE, "Test requires atleast two numa nodes")
+    @skipIf(VERSION_CHK, "Test requires Python 3.7+")
+    def test_scaleitall(self):
```
PraveenPenguin, line 71: as having single test move skip statement in setup itself else unnecessary setUp will call

kernel/will-it-scale.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        To execute test using git copy
+          make
+          ./runalltests
+
+        To generate graphical results
+          ./postprocess.py
+        """
+        smm = SoftwareManager()
+        pkgs = ['gcc', 'make', 'hwloc-devel']
```
PraveenPenguin, line 51: is `hwloc-devel` package available on all distro , as i am locate this only in RHEL9.*/centos9 onward ? Please double-check and consider test case in other distro if test not supported or we can have other alternative of this package ,

kernel/will-it-scale.py
```diff
@@ -0,0 +1,100 @@
...
+        self.sourcedir = os.path.join(self.workdir, 'will-it-scale-master')
+        os.chdir(self.sourcedir)
+        if build.make(self.sourcedir):
+            self.fail('make failed, please check debug logs')
+
+    @skipIf(SINGLE_NODE, "Test requires atleast two numa nodes")
+    @skipIf(VERSION_CHK, "Test requires Python 3.7+")
+    def test_scaleitall(self):
+        """
+        Invoke and execute all tests
```
PraveenPenguin, line 73: Can it will reword as `execute test(s)` as it running test based on input

kernel/will-it-scale.py
```diff
@@ -0,0 +1,100 @@
...
+
+        # Identify the test to be executed
+        if self.testcase in 'All':
+            cmd = './runalltests'
+        else:
+            cmd = './runtest.py %s > %s.csv' % (self.testcase, self.testcase)
+
+        # Execute the test(s)
+        ret = process.system(cmd, shell=True, sudo=True, ignore_status=True)
+        if ret != 0:
```
PraveenPenguin, line 86: Please consider like 

`if process.system(cmd, shell=True, sudo=True, ignore_status=True) !=0 :`

as it avoid to have new variable

kernel/will-it-scale.py
```diff
@@ -0,0 +1,100 @@
...
+            self.fail('Please check the logs for failure')
+        if self.testcase not in 'All':
+            self.filetocopy = self.testcase + ".csv"
```
PraveenPenguin, line 89: Can we use string concatenation ?

kernel/will-it-scale.py
```diff
@@ -0,0 +1,100 @@
...
+            shutil.copy(self.filetocopy, self.logdir)
+
+        # Generate graphical results if postprocessing is enabled
+        if self.postprocess:
+            cmd = './postprocess.py'
+            ret = process.system(cmd, shell=True, ignore_status=True)
+            if ret != 0:
```
PraveenPenguin, line 96: here also please 

`if process.system(cmd, shell=True, ignore_status=True) != 0:`

kernel/will-it-scale.py.data/README
```diff
@@ -0,0 +1,79 @@
...
+ - tlb_flush3
+ - unix1
+ - unlink1
+ - unlink2
+ - write1
+ - writeseek1
+ - writeseek2
+ - writeseek3
+
+Please refer to the github project documentation on how to add new test cases.
```
PraveenPenguin, line 70: is this statement relevant to the current repository ? It should  not like 


Please refer to the GitHub project documentation to pick other test cases

kernel/will-it-scale.py.data/will-it-scale.yaml
```diff
@@ -0,0 +1,129 @@
+# To run all tests use following stanza instead
+#testcase: !mux
+#    All:
+#        name: All
+#        postprocess: True
```
PraveenPenguin, line 5: I feel these comment statement can be removed as user already knowing how to use ? May this can move as one of combination

kernel/will-it-scale.py
```diff
@@ -0,0 +1,100 @@
...
+        To execute test using git copy
+          make
+          ./runalltests
+
+        To generate graphical results
+          ./postprocess.py
+        """
+        smm = SoftwareManager()
+        pkgs = ['gcc', 'make', 'hwloc-devel']
+        for package in pkgs:
```
PraveenPenguin, line 52: just have like 

`for package in ['gcc', 'make', 'hwloc-devel']: `

PraveenPenguin (changes requested):

kernel/will-it-scale.py
sacsant: Have moved the skipif statement to setup()

kernel/will-it-scale.py.data/will-it-scale.yaml
sacsant: Yes, have removed it from yaml and added it to README.

sacsant:
Test run after review comments have been incorporated.

 (1/1) kernel/will-it-scale.py:WillItScaleTest.test_scaleitall: STARTED
 (1/1) kernel/will-it-scale.py:WillItScaleTest.test_scaleitall: PASS (795.09 s)

abdhaleegit (approved):
LGTM

kernel/will-it-scale.py
PraveenPenguin: @sacsant thanks for addressing review comments, can you please look in this comment as well

sacsant: Ah, thanks for the reminder. I missed this one. I checked RHEL & Ubuntu and the package is available. SLES does not seem to have this package. I can add a check for distro versions but was wondering since the test gets cancelled anyway if the package is not available, should we add this extra code?

PraveenPenguin (approved):

PR #2796, opened 2024-04-15
Addes test case to capture ebizzy workload in perf top tool

Krishan-Saraswat:
To run this test, we need to follow this flow
start_ebizzy_workload -> run workload test -> stop_ebizzy_workload.

$avocado run --max-parallel-tasks=1 ebizzy_workload.py:test_start_ebizzy_workload
$ avocado run --max-parallel-tasks=1 perf_top.py:test_workload_output
$avocado run --max-parallel-tasks=1 ebizzy_workload.py:test_stop_ebizzy_workload


PraveenPenguin (approved):

sacsant:
If installation of any packages required by this test fails, this teardown function may error out since temp_file variable   is defined after package install code.  Ideally there should a check added in teardown to check if this symbol/variable is available. Alternately the definition of this variable can be moved to the top of setup()

 def tearDown(self):
        if os.path.isfile(self.temp_file):
            process.system('rm -f %s' % self.temp_file)

PR #2818, opened 2024-05-09
Using new parameter name in yaml and source script file

PavamanSubramaniyam:
To avoid conflict between peer ipaddr parameter being used in different network and ROCE tests with different values assigned


io/net/sriov_device_test.py
```diff
@@ -77,7 +77,8 @@ def setUp(self):
 
         if self.params.get('netmasks'):
             self.prefix = self.netmask_to_cidr(self.netmask[0])
-        self.peer_ip = self.params.get('peer_ip', default="").split(' ')
+        self.peer_ipaddr = self.params.get(
+            'peer_ipaddr', default="").split(' ')
```
Naresh-ibm, line 81: @PavamanSubramaniyam I think changing only `self.params.get(
            'peer_ipaddr', default="").split(' ')` is fine.
do not change the assigned variable name. just change the value in yaml and the one in the `self.params.get('peer_ip', default="").split(' ')`

let me know if you need any help here

Naresh-ibm (changes requested):

PavamanSubramaniyam:
I have resent the patch after incorporating the review comments.

abdhaleegit:
@PavamanSubramaniyam @Naresh-ibm normally we do not allow to change the existing parameter names as it reflects in CI/CR  runs .. so can you elaborate the need of parameter change here

btw, a rule we follow is for single peer ip the param is `peer_ip` and if there are multiple peer ips the param is `peer_ips`

PavamanSubramaniyam:
The current code in sriov_device_test.py script and the corresponding sriov_device_test.yaml, sriov_roce_device_test.yaml files, we are using peer_ip parameter, even though we can pass multiple ip's to this parameter.
So, we need to change this parameter to peer_ips for passing multiple ip's in the script and also in the .yaml files respectively.
The functionalities provided in this sriov_device_test.py script are handling multiple IP's provided in the parameter files.

But the other network test scripts which only need single IP for execution, will also be referring to the same peer_ip parameter provided in the input txt file which causes failure.

abdhaleegit:
> @PavamanSubramaniyam @Naresh-ibm normally we do not allow to change the existing parameter names as it reflects in CI/CR runs .. so can you elaborate the need of parameter change here
> 
> btw, a rule we follow is for single peer ip the param is `peer_ip` and if there are multiple peer ips the param is `peer_ips`

As discussed, will follow the above format. Please resend the code

abdhaleegit:
@PavamanSubramaniyam Please remove the commit

PavamanSubramaniyam:
I have removed the commit and resent the changes as new commit

abdhaleegit (approved):
LGTM

Naresh-ibm (commented):
LGTM

Naresh-ibm (approved):
@PavamanSubramaniyam Thanks for PR

PR #2841, opened 2024-06-12
Fixing issue : as test error out when test cancelled due to package n…

PraveenPenguin:
…ot installed

as self.thp page_cache object is not available in case of package condition not meet


Naresh-ibm (approved):
LGTM

PR #2847, opened 2024-06-27
fix patch apply issue

PraveenPenguin:
this patch address patch apply failure issue in couple of test as recent avocado has regression and community want that it should be fixed in test case


PraveenPenguin:
Merging as hot patch

PR #2849, opened 2024-06-28
as only yaml file is suffcient other can be remove as it increase mai…

PraveenPenguin:
…ntainace overhead

as only yaml file is suffcient other can be remove as it increase maintainace overhead


PraveenPenguin:
@SamirMulani please review this

SamirMulani:
LGTM

PR #2857, opened 2024-07-12
Rename cpustress.py to cpuhotplug.py

abdhaleegit:
Changed the named of test from cpustress.py to cpuhotplug.py to follow naming as per the actual test it performs. i.e cpu add or remove


cpu/cpuhotplug.py
```diff
@@ -39,7 +39,7 @@ def collect_dmesg(object):
     object.whiteboard = process.system_output("dmesg").decode('utf-8')
 
 
-class cpustresstest(Test):
+class cpuhotplug(Test):
```
PraveenPenguin, line 42: @abdhaleegit one request as we are renaming the class name. can we use this opportunity to have the class name as camelCase

abdhaleegit: done

PR #2880, opened 2024-09-03
Added support to run cpu folding workload for specific time

SamirMulani:
Here to run the CPU fold workload for a specific duration added `runtime` parameter using .yaml


SamirMulani:
Latest commit: https://github.com/avocado-framework-tests/avocado-misc-tests/pull/2881

PR #2884, opened 2024-09-04
Fixed logical issue in kselftest for if/else block.

SamirMulani:
Refactor the code logic to execute the kselftest cases based on the selected component correctly. 
`Ex: bpf, cpufrequecy .. etc`

### DIFF
--- a/kernel/kselftest.py
+++ b/kernel/kselftest.py
@@ -177,7 +177,7 @@ class kselftest(Test):
         kself_args = self.params.get("kself_args", default='')
         if self.comp == "bpf":
             self.bpf()
-        if self.comp == "cpufreq":
+        elif self.comp == "cpufreq":
             self.cpufreq()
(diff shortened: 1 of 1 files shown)


kernel/kselftest.py
```diff
@@ -177,7 +177,7 @@ def test(self):
         kself_args = self.params.get("kself_args", default='')
         if self.comp == "bpf":
             self.bpf()
-        if self.comp == "cpufreq":
+        elif self.comp == "cpufreq":
```
PraveenPenguin, line 180: this will break another part of the code like `result` is an undefined variable, please take care of that well, easy way is to return value from `cpufreq` function and store in result variable

PraveenPenguin (changes requested):

Naresh-ibm:
@SamirMulani any update here?

PR #2909, opened 2024-10-22
Patch to fix test failure on lpar with 1GB hugepages.

Pavithra1602:
Adding check to verify count value is non-zero.

[Bug 209952] RHEL10 [ P10 ] [ 6.11.0-0.rc5.22.el10.ppc64le ] [ FW1110.00 ]: transparent_hugepages.py test is failing on RHEL10

### DIFF
--- a/memory/transparent_hugepages.py
+++ b/memory/transparent_hugepages.py
@@ -91,6 +91,8 @@ class Thp(Test):
         # Start Stresssing the  System
         self.log.info('Stress testing using dd command')
 
+        if not self.count:
+            self.fail("Please pass valid value for mem_size in yaml file")
         for iterator in range(self.count):
             stress_cmd = 'dd if=/dev/zero of=%s/%d bs=%dM count=1'\
                          % (self.mem_path, iterator, self.block_size)


memory/transparent_hugepages.py
```diff
@@ -91,6 +91,8 @@ def test(self):
         # Start Stresssing the  System
         self.log.info('Stress testing using dd command')
 
+        if not self.count:
+            self.fail("Please pass valid value for mem_size in yaml file")
```
Naresh-ibm, line 95: @Pavithra1602 where are we checking if `self.count >= 1G`
some time if `self.count < 1G` with some float value, then it should cancel the test right?

And please use `self.cancel` instead of `self.fail`

Naresh-ibm: @Pavithra1602 any update?

Pavithra1602: @Naresh-ibm actually when 1gb hugepage is configured count value is coming as 0 when less than 2gb is configured in yaml file for mem_size
so added check to see if it is non zero

Naresh-ibm: > @Pavithra1602 where are we checking if `self.count >= 1G` some time if `self.count < 1G` with some float value, then it should cancel the test right?
> 
> And please use `self.cancel` instead of `self.fail`

@Pavithra1602 please address this

Pavithra1602:
Raised PR https://github.com/avocado-framework-tests/avocado-misc-tests/pull/2917

PR #2914, opened 2024-11-06
bonnie: fix url link for test suite

abdhaleegit:
The old url link do not work, as it moved from http to https and also update to latest version


PraveenPenguin (commented):
even can you please add as yaml support

Naresh-ibm:
> > even can you please add as yaml support
> 
> we can, but normally the link changes once a year.. does it require to pass from yaml ?

I feel adding through yaml is always a good part, people can just change it as yaml parameter instead of sending the PR to change it again. we reduce this dependency if it is through yaml

PraveenPenguin (approved):

PR #2920, opened 2025-01-20
Skip malloc1 and malloc2 due to indefinite runtime

SamirMulani:
While running the will-it-scale test suite, it was observed that certain tests are getting stuck indefinitely, causing the entire test suite to be blocked. To avoid this, we are temporarily removing the following tests from the will-it-scale test bucket
1. malloc1
2. malloc2


PraveenPenguin (approved):
LGTM

PR #2934, opened 2025-02-25
perf/perf_sdt_probe: enhance test to run all sdt events

disgoel:
This patch adds logic to test all the SDT events present and later removes them. 
Also increase the sleep time to avoid failure in cases where perf list is not updated after adding probe.
```
# avocado run perf_sdt_probe.py
JOB ID     : 4dc31fdc2089b8cad5dda2de1650e181251a571a
JOB LOG    : /home/avocado-fvt-wrapper/results/job-2025-02-25T13.52-4dc31fd/job.log
 (1/1) perf_sdt_probe.py:PerfSDT.test: STARTED
 (1/1) perf_sdt_probe.py:PerfSDT.test: PASS (247.02 s)
```
[debug.log](https://github.com/user-attachments/files/18962884/debug.log)

### DIFF
--- a/perf/perf_sdt_probe.py
+++ b/perf/perf_sdt_probe.py
@@ -34,14 +34,14 @@ class PerfSDT(Test):
     """
 
     def run_cmd(self, cmd):
-        if process.system(cmd, ignore_status=True, sudo=True, shell=True):
+        if process.system(cmd, sudo=True, shell=True):
             self.is_fail += 1
         return
 
     @staticmethod
     def run_cmd_out(cmd):
         return process.system_output(cmd, shell=True,
-                                     ignore_status=True, sudo=True).decode("utf-8")
+                                     sudo=True).decode("utf-8")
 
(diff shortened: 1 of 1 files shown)


perf/perf_sdt_probe.py
```diff
@@ -34,14 +34,14 @@ class PerfSDT(Test):
     """
 
     def run_cmd(self, cmd):
-        if process.system(cmd, ignore_status=True, sudo=True, shell=True):
+        if process.system(cmd, sudo=True, shell=True):
```
PraveenPenguin, line 37: any reason to `ignore_status=True` to removig this

disgoel: If any perf sdt event fails, the test should fail. If we add `ignore_status=True` it will ignore the failure.

Naresh-ibm (approved):
@disgoel Thanks for PR
LGTM

PR #2941, opened 2025-03-03
Changing the plotform validation to power for IBM power servers

Naresh-ibm:
earlier we were validating "ppc" for power server, Now it has changed to power as from RH10 onwards this has been changed and the utility is also now returning powerpc


PraveenPenguin (approved):
LGTM

PR #2947, opened 2025-03-05
Fixed uses of showing warnings in avocado misc tests.

SamirMulani:
In older versions of Avocado, the self.warn method was used to display warnings, and we adapted this approach when the API supported it. However, in the latest version of Avocado, warnings are now displayed using self.log.warn(), as warning messages have been incorporated into the logging class. We have adopted the same API in multiple test cases where we previously used the older method for displaying warnings.


PraveenPenguin:
@SamirMulani please wrap  both the commits in single commit

Naresh-ibm:
@SamirMulani can you please add runs logs for at. least couple of major test cases,

Naresh-ibm:
@SamirMulani if we are pursuing  this, please close it.

PR #2963, opened 2025-03-22
fix for uperf test failure when test CANCELS

vaishnavibhat:
When required packages and not available, instead of CANCEL, the test was giving ERROR. Patch takes care of setting the right status.


vaishnavibhat:
Before patch :

avocado run uperf_test.py -m uperf_test.py.data/uperf_test.yaml --max-parallel-tasks=1
Fetching asset from uperf_test.py:Uperf.test
JOB ID     : 64eb725f0f746a094a298d256560930d431764a1
JOB LOG    : /root/netperf/avocado-fvt-wrapper/results/job-2025-03-22T16.23-64eb725/job.log
 (1/1) uperf_test.py:Uperf.test;run-mtu-1500-1a6a: STARTED
 (1/1) uperf_test.py:Uperf.test;run-mtu-1500-1a6a: ERROR: 'Uperf' object has no attribute 'obj' (6.12 s)
RESULTS    : PASS 0 | ERROR 1 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /root/netperf/avocado-fvt-wrapper/results/job-2025-03-22T16.23-64eb725/results.html
JOB TIME   : 20.61 s

Test summary:
1-uperf_test.py:Uperf.test;run-mtu-1500-1a6a: ERROR


After patch:

avocado run uperf_test.py -m uperf_test.py.data/uperf_test.yaml --max-parallel-tasks=1
Fetching asset from uperf_test.py:Uperf.test
JOB ID     : 8205df722bdfb199a41d745ae7b587dd5b111c6d
JOB LOG    : /root/netperf/avocado-fvt-wrapper/results/job-2025-03-22T16.32-8205df7/job.log
 (1/1) uperf_test.py:Uperf.test;run-mtu-1500-1a6a: STARTED
 (1/1) uperf_test.py:Uperf.test;run-mtu-1500-1a6a: CANCEL: Unable to install the package lksctp-tools-devel on host machine (6.52 s)
JOB HTML   : /root/netperf/avocado-fvt-wrapper/results/job-2025-03-22T16.32-8205df7/results.html
JOB TIME   : 20.01 s

vaishnavibhat:
[Uploading job.log…]()

io/net/uperf_test.py
```diff
@@ -198,7 +200,7 @@ def tearDown(self):
         """
         Killing Uperf process in peer machine
         """
-        if self.networkinterface:
+        if self.networkinterface and self.uperf_run == '1':
```
Naresh-ibm, line 203: @vaishnavibhat This condition always fails based on your new variable. Since `self.uperf_run` not changing after assigned with 0. are we trying to avoid this block of code from executions?

vaishnavibhat: > @vaishnavibhat This condition always fails based on your new variable. Since `self.uperf_run` not changing after assigned with 0. are we trying to avoid this block of code from executions?

self.uperf_run is set to 1 when UPERF_SERVER_RUN =1 . This happens when the uperf server is running.  The patch takes care of stopping the uperf server only if the UPERF_SERVER_RUN =1 / self.uperf_run is set to 1. 

I might have to take care of the cleanup part in the peer  and need to check if the below would be a better approach. 
```
if self.networkinterface:
        self.uperf_run == '1':
            self.obj.stop()
        cmd = "pkill uperf; rm -rf /tmp/uperf-master"
        output = self.session.cmd(cmd)
```

Naresh-ibm: ya this looks better to me

nasastry: Instead of using values 0 & 1, boolean variable can be used with False/True.

abdhaleegit: @vaishnavibhat Please follow the True or False  .. and reflect same in netperf and iperf...  the reason being... checking any condition if it is enabled/disabled or running/notrunning ... always good to have boolean...

I see we are passing 1 or 0  as integer in yaml... in code we are checking as string self.uperf_run == '1'   single quote ..so to avoid this better to go for boolean

you can send the changes to other test too

abdhaleegit (approved):
LGTM

PR #2966, opened 2025-03-28
Adding "lshw" to package list that needs to be installed before test …

Naresh-ibm:
…runs

In few latest distros, lshw package is not getting installed by default during installation and hence the test cases are failing. So added the package name to the list that needs to be installed before test start and test are running fine now.


PraveenPenguin (approved):
LGTM

PR #2968, opened 2025-04-03
use the right command for firewall service

vaishnavibhat:
use firewalld service for sles16


Naresh-ibm:
same for rest of the files too.

PR #2969, opened 2025-04-08
Extend package list for SUSE

vaishnavibhat:
Extend the package list and update the latest tcpdump package for sles16


Naresh-ibm (approved):
LGTM

PR #2983, opened 2025-04-22
Fixing different python package names

PavamanSubramaniyam:
So handling the different package names available to be installed accordingly


abdhaleegit (approved):
LGTM

abdhaleegit:
Thanks for the patch

PR #3001, opened 2025-05-26
Checking if the pci slot obtained is present or not

PavamanSubramaniyam:
Introducing the generic check to see if the slot directory is present to perform add and remove operation

Since for SRIOV VF Ethernet and ROCE ports there will be no slot directory available and hence pci hotplug cannot be done


PraveenPenguin (approved):
LGTM

PR #3023, opened 2025-07-10
Added support for taking device name as susbys or nqn of nvme disk and added support for shared namespaces.

maramsmurthy:
Added changes/support as follows
1. Device name can be provided as follows in yaml file a. subsystem name b. device nqn name
2. Added paramter in yaml to enable shared namespace functionality, Default value is set to False as it should not impact exisitng functionality.
3. Added function calls for shared namespace creation. Conclusion: Changes works for sing controller nvme dis and multi controller nvme disk

### DIFF
--- a/io/disk/ssd/nvmetest.py
+++ b/io/disk/ssd/nvmetest.py
@@ -46,6 +46,10 @@ class NVMeTest(Test):
         nvme_node = self.params.get('device', default=None)
         if not nvme_node:
             self.cancel("Please provide valid nvme node name")
+        elif "subsys" in nvme_node:
+            nvme_node = nvme.get_controllers_with_subsys(nvme_node)[0]
+        elif nvme_node.startswith("nqn."):
+            nvme_node = nvme.get_controllers_with_nqn(nvme_node)[0]
         self.device = disk.get_absolute_disk_path(nvme_node)
         cmd = 'ls %s' % self.device
         if process.system(cmd, ignore_status=True):
@@ -73,8 +77,10 @@ class NVMeTest(Test):
                     smm.install("nvme-cli"):
                 self.cancel('nvme-cli is needed for the test to be run')
             self.binary = 'nvme'
+
         self.format_size = self.get_block_size()
         self.namespace = self.params.get('namespace', default='1')
+        self.shared = self.params.get("shared_namespaces", default=False)
         self.id_ns = "%sn%s" % (self.device, self.namespace)
         self.firmware_url = self.params.get('firmware_url', default='')
         if 'firmware_upgrade' in str(self.name) and not self.firmware_url:
@@ -345,9 +351,10 @@ class NVMeTest(Test):
         """
         Test to create namespace with full capacity
         """
-        self.delete_all_ns()
-        self.create_full_capacity_ns()
-        self.list_ns()
+        device = self.device.split("/")[-1]
+        nvme.delete_all_ns(device)
+        nvme.create_full_capacity_ns(device,
+                                     shared_ns=self.shared)
 
     def testformatnamespace(self):
         """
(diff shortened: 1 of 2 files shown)


maramsmurthy:
[debug.log](https://github.com/user-attachments/files/21170394/debug.log)
Execution log

abdhaleegit:
Waiting for https://github.com/avocado-framework/avocado/pull/6185

abdhaleegit (commented):
@maramsmurthy can you please paste me the nvme device name types input .. it works for ? like what is the values user should give

abdhaleegit:
input examples
```
Exapmle input as follows can be provided to new definitions as followds
NQN: nqn.1994-11.com.vendor:nvme:modela:2.5-inch:SERIALNUM
Subsystem: nvme-subsysX
controller: nvmeX
```

io/disk/ssd/nvmetest.py
```diff
@@ -46,6 +46,10 @@ def setUp(self):
         nvme_node = self.params.get('device', default=None)
         if not nvme_node:
             self.cancel("Please provide valid nvme node name")
+        elif "subsys" in nvme_node:
```
Naresh-ibm, line 49: @maramsmurthy why subsys name will be there is node_name, it is just kind of /dev/nvmeX right?

maramsmurthy: Hi naresh,
Now we can provide as follows for "device" parameter in yaml file
NQN: nqn.1994-11.com.vendor:nvme:modela:2.5-inch:SERIALNUM
Subsystem: nvme-subsysX
controller: nvmeX
So the nvme_node might have value like "nvme-subsysX"

Naresh-ibm: @maramsmurthy can you please add these in the form command above the disk parameter, so that it is clear to any new user

io/disk/ssd/nvmetest.py
```diff
@@ -345,9 +351,13 @@ def test_create_full_capacity_ns(self):
...
-        self.delete_all_ns()
-        self.create_full_capacity_ns()
-        self.list_ns()
+        device = self.device.split("/")[-1]
+        nvme.delete_all_ns(device)
+        nvme.create_full_capacity_ns(device)
+        if self.shared:
+            nvme.delete_all_ns(device)
+            nvme.create_full_capacity_ns(device,
+                                         shared_ns=self.shared)
```
abdhaleegit, line 360: nvme.delete_all_ns(device) looks duplicate ?

io/disk/ssd/nvmetest.py
```diff
@@ -447,6 +457,13 @@ def test_create_namespaces(self):
         ns_count = self.params.get('namespace_count', default=1)
         device = self.device.split("/")[-1]
         nvme.create_namespaces(device, ns_count)
+        if self.shared:
+            nvme.delete_all_ns(device)
+            self.log.info(self.shared)
+            nvme.create_namespaces(device,
+                                   ns_count,
+                                   shared_ns=self.shared)
+            nvme.delete_all_ns(device)
```
abdhaleegit, line 466: why delete all again after creation

abdhaleegit (changes requested):

io/disk/ssd/nvmetest.py
abdhaleegit: also self.shared can have default value as False

no need to add duplicate function call inside check... just have the flag to avoid redundant function calls

io/disk/ssd/nvmetest.py
maramsmurthy: I followed same approach which was defined earlier in order not the break the execution flow.

io/disk/ssd/nvmetest.py
abdhaleegit: ok covering normal namespace and shared names space on splitter

abdhaleegit: @Naresh-ibm your input here :

what my ask is, no need of if condition here.. and self.shared flag can be passed from yaml.. for normal and splitter... and it can use same  old code with just a flag.... and when splitter ..by covering shared name space creation.. than no need to test normal name space ?

 nvme.delete_all_ns(device)
 nvme.create_full_capacity_ns(device, shared_ns=self.shared)

PR #3024, opened 2025-07-11
Fixing the nmap compilation issue

PavamanSubramaniyam:
Different nmap versions to be dowloaded for sles16 and sles15 service pack releases

Also fixing the python not supporting between instances of str and integer


PavamanSubramaniyam:
I have updated the detail results along with the logs for different distros in the Jira.

abdhaleegit:
@PavamanSubramaniyam from our last discussion we decided to use distro provided nmap package... the reason we added upstream nmap source building is for having nping command on SUSE..

Please go for what's appropriate for SLES16, SLESold and RHEL

PR #3026, opened 2025-07-14
Enhanced existing pgbench test case.

SamirMulani:
This commit enhance the existing pgbench test case. Along with this, the codebase is improved to:

- Store benchmark logs under a dedicated test directory with structured subdirectories
- Enable easier parsing and analysis of results for performance evaluation 
- Introduce a smart log analysis module to extract min, max, and average metric
- Added support to run the benchmark for specific number of iteration


PraveenPenguin (approved):
LGTM

PR #3029, opened 2025-07-21
Add DAWR boundary check test case for unaligned 512-byte window

SACHIN-BAPPALIGE:
- Introduced a new DAWR test case 'Boundary_check.c' to validate unalignedcmemory access around a 512-byte boundary.
- Integrated the test into the existing Avocado test framework in dawr.py using the new function 'test_boundary_check'.
- Updated the build setup to include Boundary_check.c and Makefile for compilation during runtime.
- Ensures accurate detection of watchpoint triggers at misaligned addresses.

This helps validate hardware watchpoint behavior across unaligned memory windows for DAWR on Power10/P11 platforms.


PraveenPenguin (approved):
LGTM

PR #3030, opened 2025-07-23
HTX: Add support for on-demand MDT file creation in htx_test.py tests

SamirMulani:
Previously, HTX created all MDT files by default, even when not required. To optimize resource usage and improve flexibility, the HTX team has introduced support for on-demand MDT file creation.

With this update, MDT files will only be created as needed under the path: `/usr/lpp/htx/mdt/`.

Commands:
cmd 1: ` hcl -set_htx_env HTX_ON_DEMAND_MDT_CREATION 1`: Enables on-demand MDT file creation. cmd 2: `hcl -createmdt -mdt mdt_file_name`: Creates or modifies MDT files that are not created by default.

Note: cmd 2 supports both pre-existing MDT files and those created on-demand. As a part of "htx_test.py" test case added support for on demand htx mdt creation under "/usr/lpp/htx/mdt/" path

### DIFF
--- a/workload/htx_test.py
+++ b/workload/htx_test.py
@@ -92,6 +92,9 @@ class HtxTest(Test):
         else:
             self.cancel(
                 "running time unit is not proper, please pass as 'm' or 'h' ")
+        if not os.path.exists("/usr/lpp/htx/mdt/"):
+            self.log.info("mdt directory is created")
+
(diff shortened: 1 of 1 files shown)


PraveenPenguin (approved):
LGTM

PR #3031, opened 2025-07-24
Set True False value for server run

vaishnavibhat:
Set the values of netperf_run and iperf_run to True/ False instead of 0/ 1


abdhaleegit (approved):
LGTM

PR #3032, opened 2025-07-24
Updated htx code for block devices as per changes to htx tool

maramsmurthy:
Updated code to match changes to the HTX the way it work
1. option "-k" is added to avoid cert checks.
2. htx rpm downloaded to /tmp directory.
3. Install htx by providing absolute path.
4. Added a condition for mdt.hd to be created under mdt folder.

### DIFF
--- a/io/disk/htx_block_devices.py
+++ b/io/disk/htx_block_devices.py
@@ -143,7 +143,7 @@ class HtxTest(Test):
         """
         distro_pattern = f"{self.dist_name}{self.dist_version}"
         temp_string = process.getoutput(
-            f"curl --silent {self.rpm_link}",
+            f"curl --silent -k {self.rpm_link}",
             verbose=False, shell=True, ignore_status=True)
         matching_htx_versions = re.findall(
             r"(?<=\>)htx\w*[-]\d*[-]\w*[.]\w*[.]\w*", str(temp_string))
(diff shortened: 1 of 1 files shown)


io/disk/htx_block_devices.py
```diff
@@ -172,7 +177,9 @@ def test_start(self):
         process.run("htxcmdline -createmdt")
 
         if not os.path.exists(f"/usr/lpp/htx/mdt/{self.mdt_file}"):
-            self.fail(f"MDT file {self.mdt_file} not found")
+            process.run(f"htxcmdline -createmdt -mdt {self.mdt_file}")
+            if not os.path.exists(f"/usr/lpp/htx/mdt/{self.mdt_file}"):
+                self.fail(f"MDT file {self.mdt_file} not found")
```
Naresh-ibm, line 182: make this fail message as failed to create mdt file, as your checking after create command

io/disk/htx_block_devices.py
```diff
@@ -152,8 +152,13 @@ def install_htx_rpm(self):
...
         distro_specific_htx_versions.sort(reverse=True)
         self.latest_htx_rpm = distro_specific_htx_versions[0]
-
-        if process.system(f"rpm -ivh --nodeps {self.rpm_link}{self.latest_htx_rpm} --force",
+        tmp_dir_rpm = "/tmp/" + self.latest_htx_rpm
+        cmd = "curl -k %s/%s -o %s" % (self.rpm_link,
+                                       self.latest_htx_rpm,
+                                       tmp_dir_rpm)
+        if process.system(cmd, shell=True, ignore_status=True):
+            self.cancel(f"Dowload of hrx rpm {self.latest_htx_rpm} from rpm link {self.rpm_link}")
```
Naresh-ibm, line 160: download of htx failed from...

Naresh-ibm (changes requested):

io/disk/htx_block_devices.py
maramsmurthy: Good catch @Naresh-ibm . Updated as suggested.

io/disk/htx_block_devices.py
maramsmurthy: Yeah you are true, Corrected typos and made the message meaningful.

maramsmurthy:
Consolidated debug log captured against both distros RHEL & SUES
[debug_consolidate.log](https://github.com/user-attachments/files/21425626/debug_consolidate.log)

maramsmurthy:
rpm Installation and test is successful against RHEL and SUES across FC disk and NVME disk

PR #3036, opened 2025-07-25
Test case to verify live patching feature.

Pavithra1602:
[logs.txt](https://github.com/user-attachments/files/21433283/logs.txt)
The test case ensures that the live patching feature works as expected by applying and reverting a patch, and verifying the presence of "glibc-livepatch" messages in the output.


PraveenPenguin (approved):
LGTM

PR #3042, opened 2025-08-16
Adding support for htx to execute on different distros

PavamanSubramaniyam:
Making htx test suite to execute on sles host and rhel peer on different network adapter ports


PavamanSubramaniyam:
I have updated the detailed results along with the logs in the Jira.

abdhaleegit (commented):
Is this test for both
host rhel and peer sles
host sles with peer rhel

PR #3047, opened 2025-08-26
memory/vatest: Fix the logic for 2M/1G pages

sacsant:
get_hugepage_mbytes returns a number in megabytes for a hugepage.
This number is used in set_alt_hugepage() function to identify
hugepage size. Possibly due to an oversight check for 2M and 1G
hugepages is swapped.

Fix the logic accordingly.


PraveenPenguin (approved):
LGTM

PR #3056, opened 2025-09-25
nvme sed utils definitions added to nvme tests

maramsmurthy:
All the nvme sed utils definitions defined ealier in avocado framework are reverted due to pexpect dependency.

All those utils definitions are added in the same tests file here

No changes to defined functionality.


Naresh-ibm:
@maramsmurthy are we using the new yaml for this?

maramsmurthy:
> @maramsmurthy are we using the new yaml for this?

Yes, yaml file is merged earlier for sed testcases.

PR #3065, opened 2025-11-14
Fixing the pci lsvpd test failure

PavamanSubramaniyam:
Development is printing the lsvpd details in tuple format

So accordingly updating the handling in the test script

### DIFF
--- a/ras/ras_lsvpd.py
+++ b/ras/ras_lsvpd.py
@@ -313,13 +313,15 @@ class RASToolsLsvpd(Test):
 
                 # Device ID match
                 sys_pci_id_output = pci.get_pci_id_from_sysfs(pci_addr)
-                vpd_dev_id = vpd_output['pci_id'][4:]
+                vpd_dev_id = vpd_output['pci_id'][6:10]
+                vpd_sdev_id = vpd_output['pci_id'][19:23]
                 sysfs_dev_id = sys_pci_id_output[5:-10]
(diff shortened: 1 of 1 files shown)


PavamanSubramaniyam:
I have uploaded the complete result logs file to the Jira.

PavamanSubramaniyam:
I have verified with FC adapter and Network adapters also and got the same results. The RAS development team confirmed that the new lsvpd versions will have the tuple format introduced and has been picked up by the RHEL distros like RHEL9.7, RHEL9.8, RHEL10.1 and RHEL10.2.

Naresh-ibm:
> I have verified with FC adapter and Network adapters also and got the same results. The RAS development team confirmed that the new lsvpd versions will have the tuple format introduced and has been picked up by the RHEL distros like RHEL9.7, RHEL9.8, RHEL10.1 and RHEL10.2.

In that case can we make a condition and append the changes, like
if distro.version < 9.7
   <new code>
 else:
    <old_code>


same for sles too.


let me know if you need any help here.
as people might be still running the older release for testing. So don't want this to break.

Naresh-ibm (approved):
LGTM

Naresh-ibm:
@PavamanSubramaniyam I could see there is some conflict for this PR
can you please resolve it and re-send, I will merge it

abdhaleegit:
@PavamanSubramaniyam rebase the code

PR #3066, opened 2025-11-18
perf_record_options.py:: Added new testcase to run all present and less used options for perf record

Tejas3772:
Implemented a new Avocado testcase to validate perf record options. The testcase runs perf record with both documented and less frequently used/unused options, capturing output and errors for each.


Tejas3772:
Okay, will raise to add after the other PR is merged.

PraveenPenguin:
@Tejas3772 can you please add latest job log

PR #3067, opened 2025-11-19
perf_top.py: Fixes related to pexpect and ebizzy

Tejas3772:
Fixes for pexpect and expect:spawn issues with library and recognizing pexpect in perf_top.py:test_top$.
 Also ebizzy finished before capture, fixed that in perf_top.py:test_workload_output$


sacsant (changes requested):
Please fix the commit message, provide a proper signed-off-by ID. The ID mentioned in the signed-off-by seems to be a machine related id

Tejas3772:
Done

PR #3095, opened 2026-02-25
Adding the check for module to be the same as driver to be tested

PavamanSubramaniyam:
Introducing the check for making sure if the network adapter is physically present in the machine Thereby if the driver of the network interface is the same as the kernel module to be tested

### DIFF
--- a/io/driver/driver_parameter.py
+++ b/io/driver/driver_parameter.py
@@ -50,6 +50,9 @@ class Moduleparameter(Test):
         self.module = self.params.get('module', default=None)
         if not self.module:
             self.cancel("Please provide the Module name")
+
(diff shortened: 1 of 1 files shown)


vaishnavibhat (changes requested):
Hi Pavaman, 
Do you think this is a good candidate to be moved to avocado utils ? 

Thank you.

Naresh-ibm:
> Hi Pavaman, Do you think this is a good candidate to be moved to avocado utils ?
> 
> Thank you.

@vaishnavibhat lets have this in misc-test as of now and later we can think of pushing to utils as it requires more brainstorming around this area

Naresh-ibm (approved):
LGTM

PR #3096, opened 2026-02-25
generic/ltp: Replace runltp with kirk tool

sacsant:
Following LTP commit 6efd3605d which removed runltp, update
avocado-misc-tests to use kirk as the test runner instead.
V4 Changes:
Keep backward compatibility support for runltp to be used with older LTP releases.
V3 Changes:
Kirk uses the LTPROOT environment variable (defaulting to /opt/ltp) to locate LTP, but avocado-misc-tests installs it to a custom location. Modified generic/ltp.py to set the LTPROOT environment variable before running kirk:
V2 Changes:
Updated command line options used in YAML to match with kirk
V1 Changes:
Updated generic/ltp.py and io/disk/ltp_fs.py to download kirk submodule and use kirk instead of runltp
Removed IDcheck.sh workaround (no longer needed with kirk)
Added kirk installation verification
Updated YAML configuration files to use 'kirk' key instead of 'runltp'
Updated script references from 'runltp' to 'kirk' in YAML files
Updated command line parameters to match with options supported by kirk.
Updated log analysis logic to parse kirk generated output

Kirk supports some of command-line options as runltp (-f, -d, -S,
etc.) and is now the official LTP test runner.

Reference: https://github.com/linux-test-project/kirk
Code changes assisted by AI

### DIFF
--- a/generic/ltp.py
+++ b/generic/ltp.py
@@ -37,7 +37,7 @@ class LTP(Test):
 
     """
     LTP (Linux Test Project) testsuite
-    :param args: Extra arguments ("runltp" can use with
+    :param args: Extra arguments ("kirk" can use with
                  "-f $test")
     LTP Network test can run on Single host or Two host
     :param two_host_configuration: must be set to True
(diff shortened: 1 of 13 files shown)


sacsant:
> @sacsant thanks for this work(patch) , can we have backward compatibility for time been like we can pass old release LTP tar file and ./runltp can work

Sure, we can add backward compatibility support. Thinking about it, how do we identify whether runltp support is present or not. runltp script still exists in old and new version of LTP.
Adding a YAML switch use_kirk can help simplify this. If this is not set runltp will be used. We still need to figure out how to handle YAML differences with kirk vs runltp. We can leave that to user to use the right set of yaml with runltp or kirk.
Would this be an acceptable solution?

PraveenPenguin (approved):
LGTM

PR #3099, opened 2026-03-05
ras/ras_ppcutils: Add support for multipath disk devices

Krishan-Saraswat:
Fix disk detection to handle multipath devices (e.g., /dev/mapper/mpatha). The test uses disk_name for ofpathname and lsdevinfo commands which require physical disk paths, not device-mapper multipath paths.

Changes:
- Added regex pattern to detect multipath device names (mpatha-part2, mpath0p1)
- Implemented multipath device resolution using 'multipath -ll' command
- Parse multipath output to extract underlying physical disk (sda, sdb, etc.)
- Added fallback to dmsetup if multipath command is unavailable
- Resolve to physical disk path before using in test commands

This fixes the "Couldn't get Disk name" error on systems using multipath storage configurations.

Assisted by AI.

### DIFF
--- a/ras/ras_ppcutils.py
+++ b/ras/ras_ppcutils.py
@@ -68,16 +68,51 @@ class RASToolsPpcutils(Test):
                             package)
         # get the disk name
         self.disk_name = ''
-        output = process.system_output(
-            "df -h", shell=True).decode().splitlines()
-        filtered_lines = [line for line in output if re.search(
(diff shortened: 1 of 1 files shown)


PraveenPenguin:
@Krishan-Saraswat can you please rebase this PR

abdhaleegit:
@Krishan-Saraswat more simple way to fetch underlaying disks  is lsblk

PR #3131, opened 2026-04-07
Added lsslot validation for Spyre

saijanani-c:
Added tests to validate Spyre accelerator
device information displayed in the output
of command "lsslot -c pci".

### DIFF
--- a/ras/ras_ppcutils.py
+++ b/ras/ras_ppcutils.py
@@ -24,6 +24,7 @@ from avocado import Test
 from avocado.utils import process, distro, build, archive, disk
 from avocado import skipIf, skipUnless
 from avocado.utils.software_manager.manager import SoftwareManager
+from avocado.utils import pci
 
 IS_POWER_NV = 'PowerNV' in open('/proc/cpuinfo', 'r').read()
 IS_KVM_GUEST = 'qemu' in open('/proc/cpuinfo', 'r').read()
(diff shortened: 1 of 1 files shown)


ras/ras_ppcutils.py
```diff
@@ -195,6 +205,18 @@ def test_lsslot(self):
             self.run_cmd_out("lsslot -ac pci")
         if not IS_KVM_GUEST:
             self.run_cmd("lsslot -c cpu -b")
+
+        if self.isAccelerator():
+            self.log.info("Accelerator detected, verifying Spyre card details")
+            lsslot_output = self.run_cmd_out("lsslot -c pci")
+            spyre_devices = self.run_cmd_out("lspci | grep -i spyre | awk '{print $1}'").strip().split('\n')
```
abdhaleegit: Please do not use shell pipe, instead handle in python way of parsing .. for better debugging

ras/ras_ppcutils.py
```diff
@@ -195,6 +205,18 @@ def test_lsslot(self):
...
+            if not spyre_devices or spyre_devices == ['']:
+                self.fail("No Spyre devices found in system")
+            for device in spyre_devices:
```
abdhaleegit, line 221: this part of code should come in else block,  becoz this gets executed even if spyre_devices are empty

abdhaleegit (changes requested):

ras/ras_ppcutils.py
saijanani-c: Change has been implemented

ras/ras_ppcutils.py
abdhaleegit: ok, it fails early

abdhaleegit (approved):
LGtm

PR #3132, opened 2026-04-09
kernel/willitscale: Numa check and package dependency fix

sacsant:
will-it-scale test fails to compile code on RHEL with
ERROR: Command 'make' failed.\nstdout: b'cc
The failure is due to missing check for RHEL as a distribution release
name and hence misses to install hwloc-devel package. Add the check so
all the required dependent packages are installed.
    
Additionally the will-it-scale tool does not require multiple NUMA nodes
to function. The tool uses hwloc to detect and distribute tasks across
CPU cores/PUs,not NUMA nodes specifically
The NUMA node check was preventing the test from running on valid
single-node systems where the tool would work correctly. This patch
removes the numa check.
    
### DIFF
--- a/kernel/will-it-scale.py
+++ b/kernel/will-it-scale.py
@@ -23,10 +23,9 @@ from sys import version_info
 
 from avocado import Test
 from avocado import skipIf
-from avocado.utils import process, build, memory, archive, distro
+from avocado.utils import process, build, archive, distro
 from avocado.utils.software_manager.manager import SoftwareManager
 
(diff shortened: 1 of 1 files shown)


sacsant:
Run output

kernel/will-it-scale.py:WillItScaleTest.test_scaleitall: STARTED     
kernel/will-it-scale.py:WillItScaleTest.test_scaleitall: PASS

PraveenPenguin (approved):
LGTM

PR #3136, opened 2026-04-15
perf: Fix and add tearDown() methods for proper cleanup

disgoel:
Fix tearDown() in multiple perf test files to prevent AttributeError and improve cleanup efficiency.

Changes:
- Add hasattr() check before accessing self.temp_file
- Replace process.run('rm -f') with os.remove()
- Replace process.system('rm -f') with os.remove()
- Add missing tearDown() in perf_script_bug.py

Before fix:
  AttributeError: 'PerfBasic' object has no attribute 'temp_file' when setUp() fails before temp_file creation

After fix:
  Tests properly CANCEL instead of ERROR when setUp() fails Temporary files cleaned up efficiently

### DIFF
--- a/perf/perf_basic.py
+++ b/perf/perf_basic.py
@@ -128,5 +128,5 @@ class PerfBasic(Test):
         self.run_cmd("perf bench sched all")
 
     def tearDown(self):
-        if os.path.isfile(self.temp_file):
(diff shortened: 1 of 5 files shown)


PraveenPenguin (approved):
LGTM

abdhaleegit (approved):
LGTM

PR #3173, opened 2026-05-21
Test to verify transperent hugepages under under real-world conditions.

Pavithra1602:
Test Transparent Huge Pages (THP) by running
memory-intensive workloads including kernel compilation, package management, and memory statistics monitoring. This stress test validates THP functionality under real-world conditions.

### DIFF
--- a/memory/thp_stress_test.py
+++ b/memory/thp_stress_test.py
@@ -1,0 +1,291 @@
+#!/usr/bin/env python
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


PraveenPenguin (changes requested):
@Pavithra1602 can this be part of kself test case as i see have common function

Pavithra1602:
> @Pavithra1602 can this be part of kself test case as i see have common function

Hi @PraveenPenguin This test just tries to compile full linux kernel in loop with transpernt hugepages enabled, also this test can be low priority test and we need not run with kselftest in CR and CI

PraveenPenguin:
> > @Pavithra1602 can this be part of kself test case as i see have common function
> 
> Hi @PraveenPenguin This test just tries to compile full linux kernel in loop with transpernt hugepages enabled, also this test can be a low-priority test and we need not run with kselftest in CR and CI

@Pavithra1602 sure... how and when to run that we can control but as code properctive i thiking it can be added as enhancment to kself test case

PR #3177, opened 2026-05-26
Removed upstream part as part of Distro CR run only distro related te…

yeswanth6096:
…sts to get run

Validate CR YAML files to track distro is used as the target instead of upstream so Removed upstream part as part of Distro CR run only distro related tests to get run 

As part of Trace CR bucket only distro testcases only to run ----> https://github.ibm.com/ltctest/avocado-fvt-wrapper/blob/master/config/tests/host/trace_sanity.cfg

### DIFF
--- a/kernel/kselftest.py.data/trace_tests.yaml
+++ b/kernel/kselftest.py.data/trace_tests.yaml
@@ -14,6 +14,3 @@ component: !mux
 run_type: !mux
     distro:
         type: 'distro'
-    upstream:
-        type: 'upstream'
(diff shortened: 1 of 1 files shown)


yeswanth6096:
As part of Trace CR bucket only distro testcases only to run  ---->  https://github.ibm.com/ltctest/avocado-fvt-wrapper/blob/master/config/tests/host/trace_sanity.cfg

PraveenPenguin (approved):

yeswanth6096:
Thanks

yeswanth6096:
./

PR #3178, opened 2026-05-26
cpu: Add comprehensive CFS scheduler tunables test suite

SamirMulani:
Introduces a new test suite for validating Linux CFS scheduler tunables with focus on the top 4 most impactful and verifiable parameters:

1. base_slice_ns - Controls base time slice for CFS scheduler
2. sched_migration_cost_ns - Controls task migration cost threshold
3. sched_nr_migrate - Controls number of tasks to migrate at once
4. sched_schedstats - Enables scheduler statistics for verification

Test Coverage:
- test_01_baseline: Validates default tunable behavior
- test_02_low_latency: Tests reduced latency configuration
- test_03_high_throughput: Tests increased throughput configuration
- test_04_migration_behavior: Tests migration patterns

Features:
- Uses perf stat for accurate scheduler metrics collection
- Runs multiple iterations (10 by default) for reliable measurements
- Calculates averages and standard deviations to reduce noise
- Supports multiple workload types (cpu, context-switch, fork, mixed)
- Auto-calculates stress workers (8x CPU count) for 100% system stress
- Validates tunable impact through before/after comparisons
- Automatically restores original tunable values after tests

The test suite provides comprehensive validation of scheduler behavior under different configurations, enabling performance tuning and regression detection.

### DIFF
--- a/cpu/scheduler_tunables_test.py
+++ b/cpu/scheduler_tunables_test.py
@@ -1,0 +1,1017 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright: 2026 IBM
+# Author: Samir <samir@linux.ibm.com>
+
+"""
+Test suite for Linux CFS scheduler tunables testing.
+FOCUSED ON TOP 4 VERIFIABLE TUNABLES ONLY.
+
+This test validates the 4 most important CFS scheduler tunables:
+1. base_slice_ns - Controls the base time slice for CFS scheduler
+2. sched_migration_cost_ns - Controls task migration cost threshold
+3. sched_nr_migrate - Controls number of tasks to migrate at once
+4. sched_schedstats - Enables scheduler statistics (required for verification)
+"""
+
+import os
+import time
+from avocado import Test
+from avocado.utils import process, cpu, genio
+from avocado.utils.software_manager.manager import SoftwareManager
+
+
+class SchedulerTunablesTest(Test):
+    """
+    Test Linux scheduler tunables - TOP 4 VERIFIABLE TUNABLES.
+    Validates scheduler behavior with different tunable configurations.
+
+    :avocado: tags=cpu,scheduler,tunables,privileged
+    """
+
+    NS_TO_MS = 1_000_000
+    MIN_BASE_SLICE_NS = 750000
+    MIN_MIGRATION_COST_NS = 50000
+
+    def setUp(self):
+        """
+        Setup test environment and verify scheduler support.
+        """
+        self.sm = SoftwareManager()
+
+        required_packages = ['stress-ng', 'perf']
+        for package in required_packages:
+            if not self.sm.check_installed(package):
+                if not self.sm.install(package):
+                    self.cancel(f"Failed to install {package}")
+
+        self.total_cpus = cpu.online_count()
+        self.log.info("Total online CPUs: %d", self.total_cpus)
+
+        self.test_duration = int(self.params.get('test_duration', default=10))
+
+        configured_workers = self.params.get('stress_workers', default=0)
+        if not configured_workers:
+            # Cap at reasonable limit to prevent resource exhaustion on small
+            # LPARs
+            # Use 4x CPUs for systems with <= 8 CPUs, 8x for larger systems
+            multiplier = 4 if self.total_cpus <= 8 else 8
+            self.stress_workers = self.total_cpus * multiplier
+            self.log.info(
+                "Auto-calculating stress workers: %d (%dx %d CPUs)",
+                self.stress_workers, multiplier, self.total_cpus)
+        else:
+            self.stress_workers = int(configured_workers)
+            self.log.info("Using configured stress workers: %d",
+                          self.stress_workers)
+
+        self.original_tunables = {}
+
+        if not os.path.ismount('/sys/kernel/debug'):
+            self.log.warning("debugfs not mounted at /sys/kernel/debug")
+            self.log.info("Attempting to mount debugfs...")
+            try:
+                result = process.run(
+                    'mount -t debugfs none /sys/kernel/debug',
+                    shell=True, sudo=True, ignore_status=True)
+                if result.exit_status != 0:
+                    self.log.warning(
+                        "Failed to mount debugfs. Some tunables under "
+                        "/sys/kernel/debug/sched/ may not be available. "
+                        "Tests will use procfs tunables where possible.")
+                else:
+                    self.log.info("Successfully mounted debugfs")
+            except Exception as e:
+                self.log.warning(
+                    "Could not mount debugfs: %s. Some tunables under "
+                    "/sys/kernel/debug/sched/ may not be available. "
+                    "Tests will use procfs tunables where possible.",
+                    str(e))
+        else:
+            self.log.info("debugfs is mounted at /sys/kernel/debug")
+
+        self.log.info("\n" + "=" * 70)
+        self.log.info("TEST CONFIGURATION - TOP 4 TUNABLES")
+        self.log.info("=" * 70)
+        self.log.info("  Duration: %d seconds", self.test_duration)
+        self.log.info("  Workers: %d", self.stress_workers)
+        self.log.info("  CPUs: %d", self.total_cpus)
+        self.log.info("=" * 70)
+
+        self._check_and_save_tunables()
+
+    def _get_tunable_path(self, tunable_name):
+        """
+        Get the correct path for a tunable.
+        Returns the first existing path, or None if not found.
+
+        Note: These paths are standard across major Linux distributions
+        (RHEL, SLES, Ubuntu) for kernel 5.x and above.
+
+        Kernel version note for base_slice_ns:
+          - kernel >= 6.6: /sys/kernel/debug/sched/base_slice_ns
+          - kernel  < 6.6: /sys/kernel/debug/sched/latency_ns is the
+            equivalent tunable (controls the CFS scheduling period/latency).
+            /proc/sys/kernel/sched_latency_ns also exists on some distros.
+        """
+        tunable_paths = {
+            'base_slice_ns': [
+                '/sys/kernel/debug/sched/base_slice_ns',
+                '/sys/kernel/debug/sched/latency_ns',
+                '/proc/sys/kernel/sched_latency_ns',
+            ],
+            'sched_migration_cost_ns': [
+                '/proc/sys/kernel/sched_migration_cost_ns',
+                '/sys/kernel/debug/sched/migration_cost_ns',
+            ],
+            'sched_nr_migrate': [
+                '/proc/sys/kernel/sched_nr_migrate',
+                '/sys/kernel/debug/sched/nr_migrate',
+            ],
+            'sched_schedstats': [
+                '/proc/sys/kernel/sched_schedstats',
+            ],
+        }
+
+        paths = tunable_paths.get(tunable_name, [])
+        for path in paths:
+            if os.path.exists(path):
+                return path
+        return None
+
+    def _check_and_save_tunables(self):
+        """
+        Check availability and save original values of top 4 tunables.
+
+        For base_slice_ns: on kernels < 6.6 the equivalent tunable is
+        latency_ns (same sysctl path family). _get_tunable_path() resolves
+        this transparently, so we log an alias note when the fallback is used.
+        """
+        self.log.info("\n--- Checking Top 4 Scheduler Tunables ---")
+
+        # Map of canonical name -> fallback display name for logging
+        base_slice_fallbacks = {
+            '/sys/kernel/debug/sched/latency_ns': 'latency_ns',
+            '/proc/sys/kernel/sched_latency_ns': 'sched_latency_ns',
+        }
+
+        tunables = [
+            'base_slice_ns',
+            'sched_migration_cost_ns',
+            'sched_nr_migrate',
+            'sched_schedstats',
+        ]
+
+        for tunable_name in tunables:
+            tunable_path = self._get_tunable_path(tunable_name)
+            if tunable_path:
+                try:
+                    value = genio.read_file(tunable_path).strip()
+                    self.original_tunables[tunable_path] = value
+                    alias = base_slice_fallbacks.get(tunable_path)
+                    if alias:
+                        self.log.info(
+                            "  ✓ %s (using %s on this kernel) = %s",
+                            tunable_name, alias, value)
+                    else:
+                        self.log.info(
+                            "  ✓ %s = %s", tunable_name, value)
+                except Exception as e:
+                    self.log.warning(
+                        "  ⚠ Failed to read %s: %s", tunable_name, str(e))
+            else:
+                self.log.info(
+                    "  - %s NOT FOUND (not supported on this kernel, "
+                    "skipping)", tunable_name)
+
+        self.log.info("=" * 70)
+
+    def _set_tunable(self, tunable_name, value):
+        """
+        Set a scheduler tunable value.
+        """
+        tunable_path = self._get_tunable_path(tunable_name)
+        if not tunable_path:
+            self.fail(f"Tunable {tunable_name} not found")
+
+        try:
+            genio.write_file(tunable_path, str(value))
+            self.log.info("Set %s = %s", tunable_name, value)
+        except Exception as e:
+            self.fail(f"Failed to set {tunable_name}: {str(e)}")
+
+    def _get_tunable(self, tunable_name):
+        """
+        Get current value of a scheduler tunable.
+        """
+        tunable_path = self._get_tunable_path(tunable_name)
+        if not tunable_path:
+            return None
+
+        try:
+            return genio.read_file(tunable_path).strip()
+        except Exception as e:
+            self.log.error("Failed to read %s: %s", tunable_name, str(e))
+            return None
+
+    def _get_tunable_display_name(self, tunable_name):
+        """
+        Return a display name for a tunable that reflects the actual path
+        in use on this kernel. Useful for log messages when a fallback path
+        (e.g. latency_ns instead of base_slice_ns) is being used.
+        """
+        fallback_labels = {
+            '/sys/kernel/debug/sched/latency_ns': 'base_slice_ns(latency_ns)',
+            '/proc/sys/kernel/sched_latency_ns':
+            'base_slice_ns(sched_latency_ns)',
+        }
+        path = self._get_tunable_path(tunable_name)
+        if path and tunable_name == 'base_slice_ns':
+            return fallback_labels.get(path, tunable_name)
+        return tunable_name
+
+    def _log_tunable_change(self, name, old_val, new_val, unit='ns'):
+        """
+        Log tunable value change with appropriate unit conversion.
+
+        Args:
+            name: Name of the tunable
+            old_val: Original value
+            new_val: New value
+            unit: Unit type ('ns' for nanoseconds, 'tasks' for task count)
+        """
+        if unit == 'ns':
+            self.log.info(
+                '  %s: %d -> %d (%.2f ms -> %.2f ms)',
+                name, old_val, new_val,
+                old_val / self.NS_TO_MS, new_val / self.NS_TO_MS)
+        else:
+            self.log.info('  %s: %d -> %d %s', name, old_val, new_val, unit)
+
+    def _compare_and_log_metrics(self, baseline_metrics, test_metrics,
+                                 metric_name, display_name):
+        """
+        Compare and log metric changes between baseline and test.
+
+        Args:
+            baseline_metrics: Dictionary of baseline metrics
+            test_metrics: Dictionary of test metrics
+            metric_name: Key name of the metric to compare
+            display_name: Human-readable name for logging
+
+        Returns:
+            Percentage change if baseline > 0, None otherwise
+        """
+        baseline_val = baseline_metrics.get(metric_name, 0)
+        test_val = test_metrics.get(metric_name, 0)
+
+        self.log.info('\n%s:', display_name)
+        self.log.info('  Baseline: %d', baseline_val)
+        self.log.info('  Test: %d', test_val)
+
+        if baseline_val > 0:
+            change_pct = ((test_val - baseline_val) / baseline_val) * 100
+            self.log.info('  Change: %+.1f%%', change_pct)
+            return change_pct
+        return None
+
+    def _restore_tunables(self):
+        """
+        Restore original tunable values.
+        """
+        self.log.info("\n--- Restoring Original Tunables ---")
+        for tunable_path, value in self.original_tunables.items():
+            try:
+                genio.write_file(tunable_path, value)
+                self.log.info("  Restored %s = %s", tunable_path, value)
+            except Exception as e:
+                self.log.warning("  Failed to restore %s: %s",
+                                 tunable_path, str(e))
+
+    def _build_stress_command(self, workload_type, duration):
+        """
+        Build stress-ng command based on workload type.
+
+        Args:
+            workload_type: Type of workload (cpu, context-switch, fork, mixed)
+            duration: Duration in seconds
+
+        Returns:
+            Tuple of (stress_command, log_message)
+        """
+        if workload_type == 'cpu':
+            return (
+                f"stress-ng --cpu {self.stress_workers} --timeout {duration}s",
+                "Running CPU-bound workload (tests base_slice_ns)"
+            )
+        elif workload_type == 'context-switch':
+            workers = min(self.stress_workers, self.total_cpus * 4)
+            return (
+                f"stress-ng --switch {workers} --timeout {duration}s",
+                "Running context-switch workload (tests migration_cost_ns)"
+            )
+        elif workload_type == 'fork':
+            workers = min(self.stress_workers // 2, 64)
+            return (
+                f"stress-ng --fork {workers} --timeout {duration}s",
+                "Running fork workload (tests sched_nr_migrate)"
+            )
+        elif workload_type == 'mixed':
+            cpu_workers = self.stress_workers // 2
+            io_workers = self.stress_workers // 4
+            return (
+                f"stress-ng --cpu {cpu_workers} --io {io_workers} "
+                f"--timeout {duration}s",
+                "Running mixed CPU+I/O workload"
+            )
+        else:
+            return (
+                f"stress-ng --cpu {self.stress_workers} --timeout {duration}s",
+                f"Running default workload"
+            )
+
+    def _parse_perf_rate(self, rate_str):
+        """
+        Parse a rate value from perf stat output.
+
+        Args:
+            rate_str: Numeric rate string as output by perf (e.g., "423.717",
+                      "1.5"). Note: The SI unit suffix (e.g., "M/sec") is a
+                      separate token in perf output and is NOT handled here.
+                      This function returns the raw numeric value as-is.
+
+        Returns:
+            Float: the numeric value parsed from rate_str, or 0.0 on error.
+        """
+        try:
+            return float(rate_str)
+        except (ValueError, AttributeError):
+            return 0.0
+
+    def _parse_perf_output(self, perf_output):
+        """
+        Parse perf stat output and extract metrics.
+
+        Args:
+            perf_output: Raw perf stat output string
+
+        Returns:
+            Dictionary of parsed metrics
+        """
+        metrics = {
+            'context_switches': 0,
+            'cs_per_second': 0.0,
+            'cpu_migrations': 0,
+            'migrations_per_second': 0.0,
+            'page_faults': 0,
+            'task_clock_ms': 0.0,
+            'cpus_utilized': 0.0,
+            'branch_misses': 0,
+            'branch_miss_rate': 0.0,
+            'branches': 0,
+            'cpu_cycles': 0,
+            'instructions': 0,
+            'insn_per_cycle': 0.0,
+            'time_elapsed': 0.0
+        }
+
+        for line in perf_output.split('\n'):
+            line = line.strip()
+
+            if 'context-switches' in line:
+                parts = line.split()
+                metrics['context_switches'] = int(parts[0].replace(',', ''))
+                for i, part in enumerate(parts):
+                    if '/sec' in part and i > 0:
+                        rate_str = parts[i-1]
+                        metrics['cs_per_second'] = self._parse_perf_rate(
+                            rate_str)
+
+            elif 'cpu-migrations' in line:
+                parts = line.split()
+                metrics['cpu_migrations'] = int(parts[0].replace(',', ''))
+                for i, part in enumerate(parts):
+                    if '/sec' in part and i > 0:
+                        rate_str = parts[i-1]
+                        metrics['migrations_per_second'] = \
+                            self._parse_perf_rate(rate_str)
+
+            elif 'page-faults' in line:
+                parts = line.split()
+                metrics['page_faults'] = int(parts[0].replace(',', ''))
+
+            elif 'task-clock' in line:
+                parts = line.split()
+                metrics['task_clock_ms'] = float(parts[0].replace(',', ''))
+                for i, part in enumerate(parts):
+                    if 'CPUs' in part and i > 0:
+                        metrics['cpus_utilized'] = float(parts[i-1])
+
+            elif 'branch' in line:
+                # Parse branch-related metrics using position-based approach
+                # Perf format: "<count> <event-name> # <description>"
+                # Example 1: "603,989,146  branches  #  423.717 M/sec"
+                # Example 2: "13,838,135   branch-misses #
+                # 2.29% of all  branches"
+                parts = line.split()
+                if len(parts) >= 2 and parts[0].replace(',', '').isdigit():
+                    count = int(parts[0].replace(',', ''))
+                    # Second column is the actual event name
+                    event_name = parts[1]
+
+                    # Check the actual event name (second column),
+                    # not the description
+                    if event_name == 'branch-misses':
+                        metrics['branch_misses'] = count
+                        # Extract percentage if present in description
+                        for i, part in enumerate(parts):
+                            if '%' in part:
+                                metrics['branch_miss_rate'] = float(
+                                    part.replace('%', ''))
+                    elif event_name == 'branches':
+                        metrics['branches'] = count
+
+            elif 'cpu-cycles' in line:
+                parts = line.split()
+                metrics['cpu_cycles'] = int(parts[0].replace(',', ''))
+
+            elif 'instructions' in line:
+                parts = line.split()
+                metrics['instructions'] = int(parts[0].replace(',', ''))
+                for i, part in enumerate(parts):
+                    if 'instructions' in part and i > 1:
+                        try:
+                            metrics['insn_per_cycle'] = float(parts[i-1])
+                        except (ValueError, IndexError):
+                            pass
+
+            elif 'seconds time elapsed' in line:
+                parts = line.split()
+                metrics['time_elapsed'] = float(parts[0])
+
+        return metrics
+
+    def _log_metrics(self, metrics):
+        """
+        Log scheduler metrics in a formatted way.
+
+        Args:
+            metrics: Dictionary of metrics to log
+        """
+        self.log.info("\n" + "=" * 70)
+        self.log.info("SCHEDULER METRICS (via perf stat)")
+        self.log.info("=" * 70)
+        self.log.info(
+            "Context switches: %d (%.1f cs/sec)",
+            metrics['context_switches'], metrics['cs_per_second'])
+        self.log.info(
+            "CPU migrations: %d (%.1f migrations/sec)",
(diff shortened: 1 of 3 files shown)


cpu/scheduler_tunables_test.py.data/scheduler_tunables_test.yaml
```diff
@@ -0,0 +1,130 @@
...
+
+# Test duration in seconds
+test_duration: 10
+
+# Number of stress workers (default: 2x CPU count for overcommit)
+# Set to 0 or omit to use automatic calculation
+stress_workers: 0
+
+# Individual test configurations
+tests:
```
sacsant, line 12: This content is similar to what is mentioned in README.md file. Are all these variables read from py file.
Do you really need description and expected_behavior?

cpu/scheduler_tunables_test.py.data/scheduler_tunables_test.yaml
```diff
@@ -0,0 +1,130 @@
...
+  baseline:
+    enabled: true
+    description: "Baseline test with default system tunables"
+    workload_type: "mixed"
+    expected_behavior: "Establishes reference metrics"
+    
+  low_latency:
+    enabled: true
+    description: "Low latency configuration - reduced time slices"
```
sacsant: Similar comment

cpu/scheduler_tunables_test.py.data/scheduler_tunables_test.yaml
```diff
@@ -0,0 +1,130 @@
...
+    workload_type: "cpu"
+    tunables:
+      base_slice_ns_factor: 0.25  # 25% of baseline
+      migration_cost_ns_factor: 0.10  # 10% of baseline
+    expected_behavior: "More frequent preemption, better responsiveness"
```
sacsant: and here

cpu/scheduler_tunables_test.py.data/scheduler_tunables_test.yaml
```diff
@@ -0,0 +1,130 @@
...
+  migration_behavior:
+    enabled: true
+    description: "Migration behavior test - aggressive task migration"
+    workload_type: "fork"
+    tunables:
+      nr_migrate_factor: 2.0  # 2x baseline, max 128
+    expected_behavior: "Better load balancing with more tasks migrated"
+
+# Tunable definitions and paths
+tunables:
```
sacsant: I don't see a reason why this entire section should be part of YAML. Do you expect user to set these values? If so where is the corresponding code in py file that reads these variables?

cpu/scheduler_tunables_test.py.data/scheduler_tunables_test.yaml
```diff
@@ -0,0 +1,130 @@
...
+    description: "Enable scheduler statistics"
+    paths:
+      - "/proc/sys/kernel/sched_schedstats"
+    unit: "boolean"
+    values: "0 (disabled) or 1 (enabled)"
+
+# Workload types and their purpose
+workloads:
+  cpu:
+    description: "Pure CPU-bound workload"
```
sacsant: description and purpose is used several times. Are you expecting the user to set this?

cpu/scheduler_tunables_test.py.data/scheduler_tunables_test.yaml
```diff
@@ -0,0 +1,130 @@
...
+    description: "Fork-heavy workload"
+    stress_ng_stressor: "fork"
+    purpose: "Tests sched_nr_migrate impact on task migration"
+    
+  mixed:
+    description: "Mixed CPU and I/O workload"
+    stress_ng_stressor: "cpu + io"
+    purpose: "Overall scheduler behavior baseline"
+
+# Expected outcomes per test
```
sacsant: Why do you need this section in YAML file?

cpu/scheduler_tunables_test.py.data/scheduler_tunables_test.yaml
```diff
@@ -0,0 +1,130 @@
...
+    - "Context-switch workload shows reduced overhead"
+    - "Fewer migrations due to higher cost threshold"
+    
+  migration_behavior:
+    - "Increased nr_migrate applied successfully"
+    - "Fork workload shows better load distribution"
+    - "More tasks migrated per balancing operation"
+
+# System requirements
+requirements:
```
sacsant: Similarly this section. Is there a need?

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,757 @@
...
+                int(baseline_migration_cost),
+                int(baseline_migration_cost) / 1_000_000)
+
+            self.log.info("\nRunning workload with DEFAULT tunables...")
+            baseline_result = self._run_workload(workload_type='cpu')
+
+            if not baseline_result['success']:
+                self.fail("Baseline workload failed")
+
+            # STEP 2: Change tunables to low latency values
```
sacsant: Comment not required, self explanatory.

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,757 @@
...
+            self.log.info("\n--- STEP 2: Low Latency (Modified Tunables) ---")
+            low_base_slice = max(750000, int(baseline_base_slice) // 4)
```
sacsant: Can you define constants for these(this line and next) values?

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,757 @@
...
+            low_migration_cost = max(50000, int(baseline_migration_cost) // 10)
+
+            self.log.info("Setting low latency tunables:")
+            self.log.info(
+                "  base_slice_ns: %d -> %d (%.2f ms -> %.2f ms)",
+                int(baseline_base_slice), low_base_slice,
+                int(baseline_base_slice) / 1_000_000,
```
sacsant: and here as well including the repeats

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,757 @@
...
+            self.log.info("Running mixed CPU+I/O workload")
+
+        else:
+            stress_cmd = (
+                f"stress-ng --cpu {self.stress_workers} "
+                f"--timeout {duration}s")
+
+        # Run with perf stat - same metrics as your example
+        perf_cmd = (
+            f"perf stat -e context-switches,cpu-migrations,page-faults,"
```
sacsant, line 519: perf tool is used without checking if it is installed. What if perf is not installed?

Include perf as a dependent package in setup() for this test to run

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,757 @@
...
+        self._check_and_save_tunables()
+
+    def _get_tunable_path(self, tunable_name):
+        """
+        Get the correct path for a tunable.
+        Returns the first existing path, or None if not found.
+        """
+        tunable_paths = {
+            'base_slice_ns': [
+                '/sys/kernel/debug/sched/base_slice_ns',
```
sacsant, line 130: Can you add a check in setup() to ensure /sys/kernel/debug is mounted and available?

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,757 @@
...
+        self.log.info("\n--- Restoring Original Tunables ---")
+        for tunable_path, value in self.original_tunables.items():
+            try:
+                genio.write_file(tunable_path, value)
+                self.log.info("  Restored %s = %s", tunable_path, value)
+            except Exception as e:
+                self.log.warning("  Failed to restore %s: %s",
+                                 tunable_path, str(e))
+
+    def _run_workload_single(self, duration=None, workload_type='cpu'):
```
sacsant: This function is relatively long. Can this be split into smaller methods based on the logic. Something on the following lines:

_build_stress_command(workload_type, duration) - to handle stress-ng command
_parse_perf_output(perf_output) - for parsing perf output
_log_metrics(metrics) - logging the metrics.

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,757 @@
...
+            self.stress_workers = self.total_cpus * 8
+            self.log.info(
+                "Auto-calculating stress workers: %d (8x %d CPUs for "
+                "100%% stress)", self.stress_workers, self.total_cpus)
+        else:
+            self.stress_workers = int(yaml_workers)
+            self.log.info("Using configured stress workers: %d",
+                          self.stress_workers)
+
+        # Store original tunable values for restoration
```
sacsant: Obvious comment, not required. There are several such instances through out the code. Please audit and remove it.

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,757 @@
...
+
+        # Get system information
+        self.total_cpus = cpu.online_count()
+        self.log.info("Total online CPUs: %d", self.total_cpus)
+
+        # Get test parameters
+        self.test_duration = int(self.params.get('test_duration', default=10))
+
+        # Handle stress_workers: 0 means auto-calculate for 100% system stress
+        yaml_workers = self.params.get('stress_workers', default=0)
```
sacsant: yaml_worker name reveals WHERE the data comes from (YAML) rather than WHAT it represents. Can you instead use configured_workers?

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,757 @@
...
+        baseline_base_slice = self._get_tunable('base_slice_ns')
+        baseline_migration_cost = self._get_tunable('sched_migration_cost_ns')
+
+        if baseline_base_slice and baseline_migration_cost:
+            # Set high throughput values (3x baseline)
+            high_base_slice = int(baseline_base_slice) * 3
+            high_migration_cost = int(baseline_migration_cost) * 2
+
+            self.log.info("\nSetting high throughput tunables:")
+            self.log.info(
```
sacsant, line 841: This pattern of logging is duplicated across test methods in multiple places. Add a helper function on the following lines

```
def _log_tunable_change(self, name, old_val, new_val, unit='ns'):
    """Log tunable value change with appropriate unit conversion."""
    if unit == 'ns':
        self.log.info(
            '  %s: %d -> %d (%.2f ms -> %.2f ms)',
            name, old_val, new_val,
            old_val / self.NS_TO_MS, new_val / self.NS_TO_MS)
    else:
        self.log.info('  %s: %d -> %d %s', name, old_val, new_val, unit)
```

Then replace duplicated code:
```
# Instead of lines 567-576
self._log_tunable_change('base_slice_ns', 
                         int(baseline_base_slice), low_base_slice)
self._log_tunable_change('migration_cost_ns',
                         int(baseline_migration_cost), low_migration_cost)
```

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,757 @@
...
+                self.log.info("=" * 70)
+
+                baseline_cs = baseline_result['metrics'].get(
+                    'context_switches', 0)
+                test_cs = test_result['metrics'].get('context_switches', 0)
+                baseline_mig = baseline_result['metrics'].get(
+                    'cpu_migrations', 0)
+                test_mig = test_result['metrics'].get('cpu_migrations', 0)
+
+                self.log.info("\nContext Switches:")
```
sacsant: Metric comparison logic is duplicated in `test_02_low_latency()`.

Add a helper method similar to:
```
def _compare_and_log_metrics(self, baseline_metrics, test_metrics, 
                             metric_name, display_name):
    """Compare and log metric changes between baseline and test."""
    baseline_val = baseline_metrics.get(metric_name, 0)
    test_val = test_metrics.get(metric_name, 0)
    
    self.log.info('\n%s:', display_name)
    self.log.info('  Baseline: %d', baseline_val)
    self.log.info('  Test: %d', test_val)
    
    if baseline_val > 0:
        change_pct = ((test_val - baseline_val) / baseline_val) * 100
        self.log.info('  Change: %+.1f%%', change_pct)
        return change_pct
    return None
```

and then use it as:
```
# Replace lines 599-622
cs_change = self._compare_and_log_metrics(
    baseline_result['metrics'], test_result['metrics'],
    'context_switches', 'Context Switches')

if cs_change and cs_change > 0:
    self.log.info('  ✓ EXPECTED: More context switches with smaller time slices')
else:
    self.log.warning('  ⚠ UNEXPECTED: Fewer context switches')

mig_change = self._compare_and_log_metrics(
    baseline_result['metrics'], test_result['metrics'],
    'cpu_migrations', 'CPU Migrations')
```

sacsant (changes requested):

cpu/scheduler_tunables_test.py.data/scheduler_tunables_test.yaml
SamirMulani: Thank you for the review comments. Yes, we don't need this "description and expected_behavior" driving from the .yaml file; we can add it as a part of the .py itself.

cpu/scheduler_tunables_test.py.data/scheduler_tunables_test.yaml
SamirMulani: Entry for description, expected_behavior we should not have from .yaml. Will update the supportive code for the same.
For other parameters, we are setting the default values for workload_type, nr_migrate_factor, etc. in setup itself which is needed for the test case.

cpu/scheduler_tunables_test.py.data/scheduler_tunables_test.yaml
SamirMulani: I think we can remove and make it simpler without the "description and purpose"

cpu/scheduler_tunables_test.py
SamirMulani: Agree

SamirMulani:
```
JOB ID     : 82a2825ff14aa99ee6b2636573fa69c2fdbeec48
JOB LOG    : /root/avocado-fvt-wrapper/results/job-2026-06-03T16.43-82a2825/job.log
 (1/4) scheduler_tunnables.py:SchedulerTunablesTest.test_01_baseline;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED
 (1/4) scheduler_tunnables.py:SchedulerTunablesTest.test_01_baseline;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (198.66 s)
 (2/4) scheduler_tunnables.py:SchedulerTunablesTest.test_02_low_latency;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED
 (2/4) scheduler_tunnables.py:SchedulerTunablesTest.test_02_low_latency;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (204.87 s)
 (3/4) scheduler_tunnables.py:SchedulerTunablesTest.test_03_high_throughput;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED
 (3/4) scheduler_tunnables.py:SchedulerTunablesTest.test_03_high_throughput;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (203.90 s)
 (4/4) scheduler_tunnables.py:SchedulerTunablesTest.test_04_migration_behavior;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED
 (4/4) scheduler_tunnables.py:SchedulerTunablesTest.test_04_migration_behavior;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (102.87 s)
RESULTS    : PASS 4 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /root/avocado-fvt-wrapper/results/job-2026-06-03T16.43-82a2825/results.html
JOB TIME   : 730.00 s
```

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,757 @@
...
+        if not tunable_path:
+            self.log.warning("Tunable %s not found", tunable_name)
+            return False
+
+        try:
+            genio.write_file(tunable_path, str(value))
+            self.log.info("Set %s = %s", tunable_name, value)
+            return True
+        except Exception as e:
+            self.log.error("Failed to set %s: %s", tunable_name, str(e))
```
sacsant: I guess failing the test here would be appropriate.

sacsant (changes requested):
Can you combine all the changes into a single commit so that it is easy to review.

cpu/scheduler_tunables_test.py
SamirMulani: Agree

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,865 @@
...
+                if not self.sm.install(package):
+                    self.cancel(f"Failed to install {package}")
+
+        self.total_cpus = cpu.online_count()
+        self.log.info("Total online CPUs: %d", self.total_cpus)
+
+        self.test_duration = int(self.params.get('test_duration', default=10))
+
+        configured_workers = self.params.get('stress_workers', default=0)
+        if configured_workers == 0 or configured_workers is None:
```
abdhaleegit: why check 2 values .. keep on

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,865 @@
...
+            self.stress_workers = self.total_cpus * 8
```
abdhaleegit: make sure 8  value does not hog ..cpu resources.. on less cpu lpars

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,865 @@
...
+        if not os.path.ismount('/sys/kernel/debug'):
+            self.log.warning("debugfs not mounted at /sys/kernel/debug")
+            self.log.info("Attempting to mount debugfs...")
+            try:
+                result = process.run(
+                    'mount -t debugfs none /sys/kernel/debug',
+                    shell=True, sudo=True, ignore_status=True)
+                if result.exit_status != 0:
+                    self.log.warning(
+                        "Failed to mount debugfs. Some tunables may not be \
```
abdhaleegit: if this is must .. than fail

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,865 @@
...
+            'sched_migration_cost_ns': [
+                '/proc/sys/kernel/sched_migration_cost_ns',
+                '/sys/kernel/debug/sched/migration_cost_ns',
+            ],
+            'sched_nr_migrate': [
+                '/proc/sys/kernel/sched_nr_migrate',
+                '/sys/kernel/debug/sched/nr_migrate',
+            ],
+            'sched_schedstats': [
+                '/proc/sys/kernel/sched_schedstats',
```
abdhaleegit, line 143: hope all these paths are common across distros ?

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,865 @@
...
+        self.log.info("Task clock: %.2f ms (%.1f CPUs utilized)",
+                      metrics['task_clock_ms'], metrics['cpus_utilized'])
+        self.log.info("Instructions per cycle: %.2f",
+                      metrics['insn_per_cycle'])
+        self.log.info("Time elapsed: %.2f seconds",
+                      metrics['time_elapsed'])
+
+        if metrics['context_switches'] > 0:
+            migrations_per_1k_cs = (
+                metrics['cpu_migrations'] * 1000.0 /
```
abdhaleegit, line 492: hope this works for all lpars/resource counts

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,865 @@
...
+        if duration is None:
+            duration = self.test_duration
+
+        self.log.info(
+            "\n--- Running %d iterations for reliable metrics ---", iterations)
+
+        all_runs = []
+        successful_runs = 0
+
+        for i in range(iterations):
```
abdhaleegit, line 573: make sure there wont be OOM killer called for the range of iteration of stress

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,865 @@
...
+                    'cpu_migrations', 'CPU Migrations')
+
+                self.log.info(
+                    "sched_nr_migrate controls how many tasks are "
+                    "migrated per load balancing operation")
+                self.log.info(
+                    "Higher value can improve load distribution but "
+                    "may increase overhead")
+
+            self._restore_tunables()
```
abdhaleegit: already doing this in teardown

abdhaleegit (changes requested):

SamirMulani:
```
JOB ID     : c0e4f332de9d902c4b430fb7700b6fed164fe22b
JOB LOG    : /root/avocado-fvt-wrapper/results/job-2026-07-09T11.15-c0e4f33/job.log
 (1/4) scheduler_tunnables.py:SchedulerTunablesTest.test_01_baseline;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED

 (1/4) scheduler_tunnables.py:SchedulerTunablesTest.test_01_baseline;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (764.02 s)
 (2/4) scheduler_tunnables.py:SchedulerTunablesTest.test_02_low_latency;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED
 (2/4) scheduler_tunnables.py:SchedulerTunablesTest.test_02_low_latency;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (205.59 s)
 (3/4) scheduler_tunnables.py:SchedulerTunablesTest.test_03_high_throughput;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED
 (3/4) scheduler_tunnables.py:SchedulerTunablesTest.test_03_high_throughput;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (204.30 s)
 (4/4) scheduler_tunnables.py:SchedulerTunablesTest.test_04_migration_behavior;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED
 (4/4) scheduler_tunnables.py:SchedulerTunablesTest.test_04_migration_behavior;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (103.04 s)
RESULTS    : PASS 4 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /root/avocado-fvt-wrapper/results/job-2026-07-09T11.15-c0e4f33/results.html
JOB TIME   : 1299.32 s
```

cpu/scheduler_tunables_test.py
SamirMulani: Yes, all the scheduler tunables that are part of the test are generic and the same across the distros

cpu/scheduler_tunables_test.py
SamirMulani: Yes

PraveenPenguin (approved):

PraveenPenguin:
@abdhaleegit quick review please

PraveenPenguin:
@sacsant can you also please review

cpu/scheduler_tunables_test.py.data/README.md
```diff
@@ -0,0 +1,289 @@
...
+    --mux-yaml scheduler_tunables_test.py.data/scheduler_tunables_test.yaml \
+    -- test_duration:20 stress_workers:64
+```
+
+### Test Parameters (YAML Configuration)
+
+- **`test_duration`**: Duration of each workload run (default: 10 seconds)
+- **`stress_workers`**: Number of stress-ng workers (default: 0 = auto-calculate as 8x CPU count)
+
+## Test Cases
```
sacsant: Can we remove the Test cases section form this README. We will need to make sure this documentation is in sync with the actual implementation. In future if the implementation changes, the document has to be updated.

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,881 @@
...
+                metrics['page_faults'] = int(parts[0].replace(',', ''))
+
+            elif 'task-clock' in line:
+                parts = line.split()
+                metrics['task_clock_ms'] = float(parts[0].replace(',', ''))
+                for i, part in enumerate(parts):
+                    if 'CPUs' in part and i > 0:
+                        metrics['cpus_utilized'] = float(parts[i-1])
+
+            elif 'branch-misses' in line and 'branches' not in line:
```
sacsant: Can you double check on this logic? This may be fragile and is tied to perf tool and how verbose it's output is.
 e.g.
```
603,989,146      branches                  #  423.717 M/sec
  13,838,135      branch-misses             #    2.29% of all branches
```

second line in output for branch-misses also contains branches string

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,881 @@
...
+        baseline_schedstats = self._get_tunable('sched_schedstats')
+
+        self.log.info("  base_slice_ns = %s", baseline_base_slice)
+        self.log.info("  sched_migration_cost_ns = %s",
+                      baseline_migration_cost)
+        self.log.info("  sched_nr_migrate = %s", baseline_nr_migrate)
+        self.log.info("  sched_schedstats = %s", baseline_schedstats)
+
+        workload_type = self.params.get(
+            'tests.baseline.workload_type', default='mixed')
```
sacsant: IS using dot notation for nested parameters correct? I don't think avocado allows it.

All nested params.get() calls for per-test configuration use dot notation (e.g. 'tests.low_latency.workload_type'). Avocado's AvocadoParams.get(key, path, default) treats the first argument as a leaf node key name, not a dotted path. The YAML tree path is passed as the second path= argument. As a result, the tests: section of the YAML config file is completely inoperative — every call silently falls back to its hardcoded default.

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,881 @@
...
+        Set a scheduler tunable value.
+        """
+        tunable_path = self._get_tunable_path(tunable_name)
+        if not tunable_path:
+            self.fail(f"Tunable {tunable_name} not found")
+
+        try:
+            genio.write_file(tunable_path, str(value))
+            self.log.info("Set %s = %s", tunable_name, value)
+            return True
```
sacsant: this statement is dead code and not required. None of the callers check for return code from _set_tunable()

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,881 @@
...
+        }
+
+        for line in perf_output.split('\n'):
+            line = line.strip()
+
+            if 'context-switches' in line:
+                parts = line.split()
+                metrics['context_switches'] = int(parts[0].replace(',', ''))
+                for i, part in enumerate(parts):
+                    if 'cs/sec' in part and i > 0:
```
sacsant: Is this correct? I think perf stat uses SI-prefixed notation instead: K/sec, M/sec

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,881 @@
...
+                        metrics['cs_per_second'] = float(parts[i-1])
+
+            elif 'cpu-migrations' in line:
+                parts = line.split()
+                metrics['cpu_migrations'] = int(parts[0].replace(',', ''))
+                for i, part in enumerate(parts):
+                    if 'migrations/sec' in part and i > 0:
```
sacsant: same comment

sacsant (changes requested):

cpu/scheduler_tunables_test.py.data/README.md
SamirMulani: Agreed. I've removed the Test Cases section to avoid documentation drift.

cpu/scheduler_tunables_test.py
SamirMulani: Good catch! The original logic was indeed fragile because it relied on searching for strings in the entire line, including the description text.
I've completely rewritten the parsing to use a **position-based approach** instead of string matching:

cpu/scheduler_tunables_test.py
SamirMulani: You're absolutely right! I was incorrectly using dot notation in the key parameter

cpu/scheduler_tunables_test.py
SamirMulani: Correct. The return statement at line 180 is unreachable because the function either succeeds (and continues) or calls self.fail(), which raises an exception. I've removed the dead `return True` statement.

cpu/scheduler_tunables_test.py
SamirMulani: You're right! Perf stat uses SI prefixes (K/sec, M/sec, G/sec) for rates, not just "/sec". I've updated the parsing logic to:

1. Look for '/sec' in the token (covers K/sec, M/sec, etc.)
2. Extract the numeric value from the previous token
3. Added a helper method `_parse_si_rate()` to handle SI prefix conversion

This makes the parsing more robust and handles various perf output formats correctly.

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,922 @@
...
+
+            self.log.info("\nSetting aggressive migration:")
+            self._log_tunable_change('sched_nr_migrate',
+                                     int(baseline_nr_migrate),
+                                     aggressive_nr_migrate, unit='tasks')
+            self.log.info(
+                "  Higher value = more tasks migrated at once = "
+                "better load balancing")
+
+            self._set_tunable('sched_nr_migrate', aggressive_nr_migrate)
```
sacsant, line 983: Unlike test_2 and test_3, this test changes tunables before running the workload. Is this okay? Is baseline not required?

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,922 @@
...
+                elif cs_change is not None:
+                    self.log.warning('  ⚠ UNEXPECTED: More context switches')
+
+                self._compare_and_log_metrics(
+                    baseline_result['metrics'], test_result['metrics'],
+                    'cpu_migrations', 'CPU Migrations')
+
+                self.log.info("=" * 70)
+
+            self._restore_tunables()
```
sacsant: Same function is called in teardown(). Isn't this duplicate? Same comment for test_2

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,922 @@
...
+                f"--timeout {duration}s",
+                "Running mixed CPU+I/O workload"
+            )
+        else:
+            return (
+                f"stress-ng --cpu {self.stress_workers} --timeout {duration}s",
+                f"Running default workload"
+            )
+
+    def _parse_si_rate(self, rate_str):
```
sacsant: the name is misleading. The function body just calls float(rate_str) and returns the raw number without applying any SI multiplier. May be rename the function to _parse_perf_rate and document that it returns the numeric value as-is from (a comment in the function already states it) perf output:

cpu/scheduler_tunables_test.py
```diff
@@ -0,0 +1,922 @@
...
+            self.log.info("Context switches: %d",
+                          metrics.get('context_switches', 0))
+            self.log.info("CPU migrations: %d",
+                          metrics.get('cpu_migrations', 0))
+
+            if metrics.get('context_switches', 0) > 0:
+                self.log.info(
+                    "✓ Context switches detected (scheduler is working)")
+            else:
+                self.log.warning("⚠ No context switches detected")
```
sacsant: In this block values checked against certain metrics if found undesirable are flagged as warning only. Shouldn't these be marked as failed.

sacsant: The test otherwise does not have any pass/fail criteria

cpu/scheduler_tunables_test.py
SamirMulani: Good catch. Added a baseline run with default sched_nr_migrate before applying the aggressive value, following the same STEP 1 / STEP 2 pattern used in test_02 and test_03. The comparison now uses the local baseline result instead of the fragile hasattr(self, 'baseline_metrics') dependency on test_01.

cpu/scheduler_tunables_test.py
SamirMulani: You're right. Removed the explicit _restore_tunables() calls from both test_02 and test_03. Since tearDown() already calls it unconditionally after every test, the mid-test calls were redundant.

cpu/scheduler_tunables_test.py
SamirMulani: Agreed, the name was misleading. Renamed to _parse_perf_rate() and updated the docstring to clearly state that the SI unit suffix is a separate token in perf output and is not handled here — the raw numeric value is returned as-is. Updated both call sites accordingly.

cpu/scheduler_tunables_test.py
SamirMulani: Agreed. Changed both self.log.warning() calls to self.fail() for the cases where no context switches and no CPU migrations are detected. These indicate the scheduler is not functioning correctly and should be hard failures, not warnings. The test now has clear pass/fail criteria.

sacsant:
@SamirMulani can you post the logs from updated test code run?

SamirMulani:
` avocado run scheduler_tunables_test.py -m scheduler_tunables_test.py.data/scheduler_tunables_test.yaml --max-parallel-tasks=1
JOB ID     : 415b8147df409074e81f83b38a0bff02fecc2503
JOB LOG    : /root/avocado-fvt-wrapper/results/job-2026-07-29T14.38-415b814/job.log
 (1/4) scheduler_tunables_test.py:SchedulerTunablesTest.test_01_baseline;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED
 (1/4) scheduler_tunables_test.py:SchedulerTunablesTest.test_01_baseline;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (101.70 s)
 (2/4) scheduler_tunables_test.py:SchedulerTunablesTest.test_02_low_latency;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED
 (2/4) scheduler_tunables_test.py:SchedulerTunablesTest.test_02_low_latency;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (206.22 s)
 (3/4) scheduler_tunables_test.py:SchedulerTunablesTest.test_03_high_throughput;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED
 (3/4) scheduler_tunables_test.py:SchedulerTunablesTest.test_03_high_throughput;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (203.70 s)
 (4/4) scheduler_tunables_test.py:SchedulerTunablesTest.test_04_migration_behavior;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED
 (4/4) scheduler_tunables_test.py:SchedulerTunablesTest.test_04_migration_behavior;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (203.67 s)
RESULTS    : PASS 4 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /root/avocado-fvt-wrapper/results/job-2026-07-29T14.38-415b814/results.html
JOB TIME   : 740.43 s`

cpu/scheduler_tunables_test.py
sacsant: you can drop the log.error() and keep self.fail()

sacsant (approved):
Rest of the code looks good to me.

SamirMulani:
```
avocado run scheduler_tunnables.py -m scheduler_tunnables.py.data/scheduler_tunnables.yaml --max-parallel-tasks=1
JOB ID     : 9ec720b7ee78807f883384eaea2f60298be734f5
JOB LOG    : /root/avocado-fvt-wrapper/results/job-2026-07-30T13.55-9ec720b/job.log
 (1/4) scheduler_tunnables.py:SchedulerTunablesTest.test_01_baseline;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED
 (1/4) scheduler_tunnables.py:SchedulerTunablesTest.test_01_baseline;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (101.57 s)
 (2/4) scheduler_tunnables.py:SchedulerTunablesTest.test_02_low_latency;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED
 (2/4) scheduler_tunnables.py:SchedulerTunablesTest.test_02_low_latency;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (203.19 s)
 (3/4) scheduler_tunnables.py:SchedulerTunablesTest.test_03_high_throughput;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED
 (3/4) scheduler_tunnables.py:SchedulerTunablesTest.test_03_high_throughput;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (202.49 s)
 (4/4) scheduler_tunnables.py:SchedulerTunablesTest.test_04_migration_behavior;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: STARTED
 (4/4) scheduler_tunnables.py:SchedulerTunablesTest.test_04_migration_behavior;run-tests-baseline-high_throughput-tunables-low_latency-migration_behavior-daa5: PASS (202.49 s)
RESULTS    : PASS 4 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /root/avocado-fvt-wrapper/results/job-2026-07-30T13.55-9ec720b/results.html
JOB TIME   : 730.91 s
```

abdhaleegit (approved):
LGTM

PraveenPenguin:
@sacsant and @abdhaleegit for review .

PR #3087, opened 2026-02-16
[avocado-misc] Perf: Use the built perf binary [depends on https://github.com/avocado-framework/avocado/pull/6272]

Sumitupadhyay1:
- Introduced support for specifying a custom perf binary path in misc tests.
- Updated perf test scripts to call ensure_tool() from distro_packages
  for dependency handling when a custom binary path is provided.
- Added corresponding .data/ YAML files for test configuration.
- Integrated perf tests into the avocado-misc-tests CI pipeline 
  to ensure consistency and reproducibility across environments.

Note:
- This change depends on ensure_tool() introduced in Avocado:
  https://github.com/avocado-framework/avocado/pull/6272

PR #3202, opened 2026-06-18
[avocado-misc][perf]: update perf_bin path in yaml configs

Sumitupadhyay1:
The perf test data yaml files set perf_bin to /tmp/perf, a transient location that does not match where the built perf binary is installed. Point perf_bin to /usr/local/bin/perf so the tests consistently use the locally built/installed perf binary.

Affected tests: perf_basic, perf_duplicate_probe, perf_events_crash_test, perf_events_test, perf_script_bug, perftool_test.

### DIFF
--- a/perf/perf_basic.py.data/perf_basic.yaml
+++ b/perf/perf_basic.py.data/perf_basic.yaml
@@ -1,1 +1,1 @@
-perf_bin: '/tmp/perf'
+perf_bin: '/usr/local/perf'
--- a/perf/perf_duplicate_probe.py.data/perf_duplicate_probe.yaml
+++ b/perf/perf_duplicate_probe.py.data/perf_duplicate_probe.yaml
@@ -1,1 +1,1 @@
-perf_bin: '/tmp/perf'
+perf_bin: '/usr/local/perf'
--- a/perf/perf_events_crash_test.py.data/perf_events_crash_test.yaml
+++ b/perf/perf_events_crash_test.py.data/perf_events_crash_test.yaml
@@ -1,1 +1,1 @@
-perf_bin: '/tmp/perf'
+perf_bin: '/usr/local/perf'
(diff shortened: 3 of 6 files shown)


Sumitupadhyay1:
@nasastry, @PraveenPenguin , @disgoel, could you please review and approve this pull request.

narasimhan-v (approved):

PraveenPenguin:
@Tejas3772 can you please review this PR

perf/perf_basic.py.data/perf_basic.yaml
```diff
@@ -1 +1 @@
-perf_bin: '/tmp/perf'
+perf_bin: '/usr/local/perf'
```
disgoel, line 1: @Sumitupadhyay1 `/usr/local/perf` is not a standard location. System perf is at `/usr/bin/perf` on RHEL, SLES, Ubuntu, Debian etc.

The original PR #3087 stated: *"Introduced support for specifying a custom perf binary path"*. If the intention was to use the system's default perf, why do we need these YAML files at all?

Sumitupadhyay1: Thanks @disgoel. The intent is the opposite of system perf: these run in our CI where perf is built from the same kernel source under test (make -C tools/ perf_install) so it matches that kernel — the distro /usr/bin/perf can mismatch. That's the custom-path case perf_bin (#3087) was added for; left empty, the tests fall back to PATH perf, which we don't want here. Our build installs it at /usr/local/perf (cp /usr/local/bin/perf /usr/local/perf), moved off /tmp since tmpfs is wiped on reboot into the test kernel.

narasimhan-v: @disgoel  We want to have a custom (upstream) perf built and placed in a path, which we are using here.
In case of system perf, we do not need the yaml file at all (this is the only parameter).

So, the parameter is only for custom path. If you want to use system perf, yaml is not needed.

The problem with current path (/tmp/perf) is it is not persistent. So we are providing a persistent path where perf is built for current kernel (/usr/local/perf). The system perf will still exist at /usr/local/bin/perf, and it will not be disturbed in this approach.

disgoel (approved):
LGTM

PR #3206, opened 2026-06-22
will-it-scale: Add step optimization patch for large systems

SamirMulani:
Add automatic application of step_optimization.patch to reduce test execution time on large systems. This patch dynamically adjusts the step value in runtest.py based on CPU count, significantly reducing test duration while maintaining meaningful scalability data.

Key improvements:
- Reduces test time from 24 hours to ~1 hour on 256-core systems
- Maintains detailed testing (step=1) on small systems (≤16 cores)
- Uses progressive step values: 2 (32 cores), 4 (64 cores), 8 (128 cores), 25 (256+ cores)
- Includes graceful fallback if patch application fails
- Applied automatically for all systems during setUp()

The patch modifies will-it-scale's runtest.py to use dynamic step calculation instead of hardcoded step=1, making the test suite practical for large-scale systems while preserving granular data on smaller systems.

Files modified:
- kernel/will-it-scale.py: Added patch application logic

Files added:
- kernel/will-it-scale.py.data/step_optimization.patch

### DIFF
--- a/kernel/will-it-scale.py
+++ b/kernel/will-it-scale.py
@@ -223,7 +223,7 @@ class WillItScaleTest(Test):
         """
         self.distro_rel = distro.detect()
         smm = SoftwareManager()
-        deps = ['gcc', 'make']
+        deps = ['gcc', 'make', 'patch']
         if self.distro_rel.name.lower() in ['fedora', 'redhat', 'rhel']:
             deps.extend(['hwloc-devel'])
         for package in deps:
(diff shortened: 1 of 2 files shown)


kernel/will-it-scale.py
```diff
@@ -243,6 +243,19 @@ def setUp(self):
         archive.extract(tarball, self.workdir)
         self.sourcedir = os.path.join(self.workdir, 'will-it-scale-master')
         os.chdir(self.sourcedir)
+
+        # Apply step optimization patch for large systems
+        # This reduces test time from 24 hours to ~1 hour on 256-core systems
+        step_patch = self.get_data('step_optimization.patch')
+        if os.path.exists(step_patch):
+            self.log.info("Applying step optimization patch for large systems...")
+            patch_cmd = 'patch -p1 < %s' % step_patch
```
sacsant, line 252: Ensure patch utility is installed on the system. Add it to setup() function.

sacsant (changes requested):

sacsant (approved):
LGTM

Naresh-ibm (approved):
LGTM

PR #3229, opened 2026-07-14
io/disk: Add LVM_extend functional validation

Priyanka98-ibm:
This adds LVM extend functional validation to the existing
lvsetup.py, which already handles all LVM-related operations such
as VG creation, LV creation etc.

Changes:
Added new parameter 'lv_extend_size' to accept extend size
Added 'test_lv_extend' which:
 - Creates a Volume Group and Logical Volume
 - Formats with the given filesystem and mounts/unmounts
 - Records LV size before extending
 - Extends the LV using lvextend command
 - Records LV size after extending
 - Verifies size has increased — fails the test if not
 - Mounts and unmounts the LV again to confirm it is healthy after extension

### DIFF
--- a/io/disk/lvsetup.py
+++ b/io/disk/lvsetup.py
@@ -16,7 +16,6 @@
 # Author: Narasimhan V <sim@linux.vnet.ibm.com>
 # Copyright: 2016 Red Hat, Inc.
 # Author: Lukas Doktor <ldoktor@redhat.com>
-# Author: Priyanka Behera <Priyanka.Behera2@ibm.com>
 #
 # Based on the code by:
 #
@@ -226,10 +225,8 @@ class Lvsetup(Test):
         size_before = lv_utils.lv_list()[self.lv_name]['LSize']
         lv_path = '/dev/%s/%s' % (self.vg_name, self.lv_name)
         cmd = 'lvextend -L +%sM %s' % (self.lv_extend_size, lv_path)
-        result = process.run(cmd, ignore_status=True)
-        if result.exit_status != 0:
-            self.fail('lvextend failed for LV %s: %s' % (self.lv_name,
-                                                         result.stderr_text))
+        if process.system(cmd, ignore_status=True):
+            self.fail('lvextend failed for LV %s' % self.lv_name)
         size_after = lv_utils.lv_list()[self.lv_name]['LSize']
 
         if size_after == size_before:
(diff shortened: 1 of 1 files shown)


io/disk/lvsetup.py
```diff
@@ -16,6 +16,7 @@
 # Author: Narasimhan V <sim@linux.vnet.ibm.com>
 # Copyright: 2016 Red Hat, Inc.
 # Author: Lukas Doktor <ldoktor@redhat.com>
+# Author: Priyanka Behera <Priyanka.Behera2@ibm.com>
```
PraveenPenguin: i don't feel it it major work to add as Author credit

io/disk/lvsetup.py
```diff
@@ -193,6 +210,36 @@ def test_vg_reactivate(self):
...
+        self.create_lv()
+        lv_utils.lv_mount(self.vg_name, self.lv_name, self.mount_loc,
+                          create_filesystem=self.fs_name)
+        lv_utils.lv_umount(self.vg_name, self.lv_name)
+
+        size_before = lv_utils.lv_list()[self.lv_name]['LSize']
+        lv_path = '/dev/%s/%s' % (self.vg_name, self.lv_name)
+        cmd = 'lvextend -L +%sM %s' % (self.lv_extend_size, lv_path)
+        result = process.run(cmd, ignore_status=True)
+        if result.exit_status != 0:
```
PraveenPenguin: is result variable is needed can be done like

    if process.system(cmd, ignore_status=True):

PraveenPenguin (changes requested):

io/disk/lvsetup.py
Priyanka98-ibm: Done, addressed both review comments — removed the Author credit and replaced process.run() + exit_status check with process.system().

Naresh-ibm (changes requested):
@Priyanka98-ibm please attach the Run logs from both rhel and sles distro versions and we can merge this

Priyanka98-ibm:
> @Priyanka98-ibm please attach the Run logs from both rhel and sles distro versions and we can merge this

Attached the automation logs for the LVM Extend validation on both RHEL and SLES for reference.
[lvm_extend.tar.gz](https://github.com/user-attachments/files/30122363/lvm_extend.tar.gz)
[sles_lvm_extend.tar.gz](https://github.com/user-attachments/files/30122367/sles_lvm_extend.tar.gz)

Naresh-ibm (approved):
LGTM
@Priyanka98-ibm Thanks for the PR

PraveenPenguin (approved):
LGTM

PR #3238, opened 2026-07-28
cpu/stressng_cpu: Add runtime scheduling-policy change test

SamirMulani:
Add test_runtime_policy_change() to validate that the Linux kernel correctly applies sched_setscheduler(2) on live stress-ng worker threads at run-time.

The test spawns stress-ng CPU worker threads in the background and walks the following policy transition chain on all worker PIDs:

  SCHED_OTHER -> SCHED_IDLE -> SCHED_BATCH -> SCHED_FIFO -> SCHED_RR -> SCHED_OTHER

SCHED_DEADLINE is excluded as it requires sched_setattr(2) with runtime/deadline/period attributes which are not exposed via the Python os.sched_setscheduler() interface.

Each transition is verified using two sources:
  - os.sched_getscheduler(pid) : direct kernel syscall (primary)
  - chrt -p <pid>              : userspace scheduling tool (secondary)

All transition results are collected in a list and reported at the end so a single run shows the complete picture of pass/fail/skip rather than stopping at the first failure.

Also introduce POLICY_META as a single source of truth for policy attributes (name, label, rt_prio) and derive CHRT_POLICY_FLAG from it via a dict comprehension to avoid duplicate policy definitions.

New YAML parameters:
  - policy_change_runtime: seconds stress-ng workers stay alive (default 120)
  - policy_change_workers: number of --cpu worker threads to spawn (default 4)

### DIFF
--- a/cpu/stressng_cpu.py
+++ b/cpu/stressng_cpu.py
@@ -13,11 +13,15 @@
 #
 # Copyright: 2022 IBM
 # Author: Rohan Deshpande <rohan_d@linux.ibm.com>
-# Modified by: Samir A Mulani <samir@linux.vnet.ibm.com>
+# Modified by: Samir M <samir@linux.ibm.com>
 #
 
 import os
+import re
 import errno
+import signal
+import subprocess
+import time
 from avocado import Test
 from avocado.utils import dmesg
 from avocado.utils import archive
@@ -25,13 +29,38 @@ from avocado.utils import build
 from avocado.utils import process
 from avocado.utils.software_manager.manager import SoftwareManager
 
-# Scheduling policy constants
+# Scheduling policy constants (from linux/sched.h / sched(7))
 SCHED_OTHER = 0
 SCHED_FIFO = 1
 SCHED_RR = 2
 SCHED_BATCH = 3
 SCHED_IDLE = 5
 SCHED_DEADLINE = 6
+
+# Human-readable names and required real-time priorities for each policy.
+# SCHED_FIFO / SCHED_RR require priority in [1, 99]; all others use 0.
+POLICY_META = {
+    SCHED_OTHER:   {"name": "SCHED_OTHER", "label": "other", "rt_prio": 0},
+    SCHED_BATCH:   {"name": "SCHED_BATCH", "label": "batch", "rt_prio": 0},
+    SCHED_IDLE:    {"name": "SCHED_IDLE",    "label": "idle", "rt_prio": 0},
+    SCHED_FIFO:    {"name": "SCHED_FIFO",    "label": "fifo", "rt_prio": 1},
+    SCHED_RR:      {"name": "SCHED_RR",      "label": "rr", "rt_prio": 1},
+    SCHED_DEADLINE: {"name": "SCHED_DEADLINE", "label": "deadline",
+                     "rt_prio": 0},
+}
+
+# Ordered transition chain for the runtime policy-change test.
+# Built from the named constants so the list never repeats bare names.
+# SCHED_DEADLINE is intentionally excluded — it requires sched_setattr(2)
+# with runtime/deadline/period attrs not exposed via os.sched_setscheduler().
+POLICY_TRANSITION_CHAIN = [
+    SCHED_OTHER,   # default CFS — starting point
+    SCHED_IDLE,
+    SCHED_BATCH,
+    SCHED_FIFO,
+    SCHED_RR,
+    SCHED_OTHER,   # restore to CFS at the end
+]
 
 
 class Stressngcpu(Test):
@@ -46,42 +75,35 @@ class Stressngcpu(Test):
     def check_policy_support(self, policy):
         """
         Check if a scheduling policy is supported by this kernel.
-        Returns the policy  if supported, False otherwise.
-        """
-        # Mapping: policy → (priority, label)
-        policy_map = {
-            SCHED_OTHER:    (0, "other"),
-            SCHED_BATCH:    (0, "batch"),
-            SCHED_IDLE:     (0, "idle"),
-            SCHED_FIFO:     (1, "fifo"),   # needs >0 priority
-            SCHED_RR:       (1, "rr"),     # needs >0 priority
-            SCHED_DEADLINE: (0, "deadline")  # handled specially
-        }
+        Returns the label string (e.g. "other", "fifo") if supported,
+        False otherwise.
 
-        if policy not in policy_map:
-            print(f"Invalid scheduling class: {policy}")
+        Uses POLICY_META for priority and label so there is a single
+        source of truth for policy attributes across the whole test.
+        """
+        meta = POLICY_META.get(policy)
+        if meta is None:
+            self.log.warning("Invalid scheduling policy constant: %d", policy)
             return False
 
-        priority, label = policy_map[policy]
+        priority = meta["rt_prio"]
+        label = meta["label"]
 
         try:
-            """
-            For deadline, os.sched_param is a dummy
-            (real params not exposed in Python)
-            """
+            # For SCHED_DEADLINE os.sched_param is a dummy; the kernel will
+            # reject it with EINVAL, which is handled below.
             param = os.sched_param(priority)
-
             os.sched_setscheduler(0, policy, param)
             return label
         except OSError as e:
             if e.errno == errno.EPERM:
-                # Permission denied → policy exists but user isn't root
+                # Permission denied — policy exists but caller is not root
                 return label
             elif e.errno == errno.EINVAL:
-                # Policy not supported by kernel
+                # Policy not recognised by this kernel build
                 return False
             else:
-                # Other errors (ESRCH, EFAULT, etc.)
+                # ESRCH, EFAULT, or any other unexpected error
                 return False
 
     def read_line_with_matching_pattern(self, filename, pattern):
@@ -94,6 +116,10 @@ class Stressngcpu(Test):
 
     def setUp(self):
         self.sched_class_payload = []
+        # PIDs of live stress-ng background workers (runtime policy test)
+        self._stressng_proc = None
+        # Determined once in setUp; avoids re-probing chrt for every PID.
+        self._chrt_available = self._probe_chrt_availability()
         smm = SoftwareManager()
         crt_stressors_list = ["bsearch", "context", "cpu", "crypt", "hsearch",
                               "longjmp", "lsearch", "matrix", "qsort", "str",
@@ -106,6 +132,12 @@ class Stressngcpu(Test):
                                    "ColinIanKing/stress-ng/archive/master.zip")
         self.crt_stressors = self.params.get(
             "crt_stressors", default=crt_stressors_list)
+        # Seconds to keep stress-ng alive during the runtime policy-change test
+        self.policy_change_runtime = int(self.params.get(
+            "policy_change_runtime", default=120))
+        # Number of stress-ng worker threads for runtime policy-change test
+        self.policy_change_workers = int(self.params.get(
+            "policy_change_workers", default=os.cpu_count() or 2))
 
         for package in ['gcc', 'make', 'libattr-devel', 'libcap-devel',
                         'libgcrypt-devel', 'zlib-devel', 'libaio-devel']:
@@ -138,6 +170,34 @@ class Stressngcpu(Test):
 
         # Clear the dmesg to capture the delta at the end of the test.
         dmesg.clear_dmesg()
+
+    def _probe_chrt_availability(self):
+        """
+        Check once whether the ``chrt`` utility is present on this system.
+
+        Called from :meth:`setUp` so the result is cached in
+        ``self._chrt_available`` and consulted by
+        :meth:`_verify_policy_for_pids` to skip the secondary check early
+        without spawning a subprocess for every PID of every transition.
+
+        Returns ``True`` if chrt is found and executable, ``False`` otherwise.
+        """
+        try:
+            subprocess.check_output(
+                ["chrt", "--version"],
+                stderr=subprocess.STDOUT,
+            )
+            return True
+        except FileNotFoundError:
+            self.log.warning(
+                "chrt not found on this system — "
+                "secondary policy verification will be skipped for the "
+                "entire test run")
+            return False
+        except subprocess.CalledProcessError:
+            # chrt --version may exit non-zero on some versions; that still
+            # means it is installed.
+            return True
 
     def stress_sched_class(self, cmd):
         """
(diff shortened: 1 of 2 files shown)


cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+        ======================================
+        Purpose
+        -------
+        Validate that the Linux kernel correctly honours scheduling-policy
+        changes applied to *live* stress-ng worker threads via
+        sched_setscheduler(2) at run-time.
+
+        Background
+        ----------
+        By default, processes run under CFS (SCHED_OTHER, policy=0).
```
sacsant, line 511: I will suggest to only retain this content under background. Remaining content is not required and is too verbose

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+        ----------------------
+        - ``policy_change_runtime``  : seconds stress-ng runs (default 120)
+        - ``policy_change_workers``  : number of CPU stressor threads
+                                       (default: os.cpu_count())
+        """
+        self.log.info("=" * 60)
+        self.log.info("Test: Runtime Scheduling-Policy Change")
+        self.log.info("=" * 60)
+
+        # Aggregate results across all transitions and all PIDs.
```
sacsant: Comment not required.

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+        # Each element is a dict with keys:
+        #   transition, pid, step, status, detail
+        all_results = []
+
+        # ----------------------------------------------------------------
+        # Step 1: Spawn stress-ng in the background
+        # ----------------------------------------------------------------
```
sacsant: Comment is unnecessary

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+        self._stressng_proc = self._spawn_stressng_background(
+            num_workers=self.policy_change_workers,
+            duration_secs=self.policy_change_runtime,
+        )
+
+        # ----------------------------------------------------------------
+        # Step 2: Discover worker PIDs
+        # ----------------------------------------------------------------
```
sacsant: Comment is unnecessary

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+
+        if not worker_pids:
+            self._terminate_stressng()
+            self.fail("No stress-ng worker PIDs found — cannot proceed.")
+
+        self.log.info("Workers under test: %s", worker_pids)
+
+        # ----------------------------------------------------------------
+        # Step 3: Walk through the policy transition chain
+        # ----------------------------------------------------------------
```
sacsant: Comment is unnecessary

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+                continue
+
+            supported_transitions.append(transition_label)
+
+            self.log.info("-" * 60)
+            self.log.info(
+                "Transition [%d/%d]: %s",
+                idx + 1, len(POLICY_TRANSITION_CHAIN) - 1, transition_label)
+
+            # -- 3a: Log current (from) policy for each worker PID ------
```
sacsant, line 583: Comment is unnecessary

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+            self.log.info("Current policies before change:")
+            for pid in worker_pids:
+                cur = self._read_proc_sched_policy(pid)
+                cur_name = POLICY_META.get(cur, {}).get("name", str(cur))
+                self.log.info("  PID %d: %s", pid, cur_name)
+
+            # -- 3b: Apply the new policy to all workers -----------------
```
sacsant, line 590: Comment is unnecessary

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+                    "transition": transition_label,
+                    "pid": pid,
+                    "step": "set-policy",
+                    "status": status,
+                    "detail": detail,
+                })
+                self.log.info(
+                    "  PID %d set-policy %s: %s", pid, to_name, status)
+
+            # -- 3c: Verify the policy was actually applied --------------
```
sacsant, line 610: Comment is unnecessary

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+                    self.log.error("  Verification FAIL: %s", detail_msg)
+                    all_results.append({
+                        "transition": transition_label,
+                        "pid": "see-detail",
+                        "step": "verify-policy",
+                        "status": "FAIL",
+                        "detail": detail_msg,
+                    })
+
+            # Brief pause so the kernel has a chance to run workers under
```
sacsant, line 638: Comment is unnecessary

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+            # the new policy before we move on.
+            time.sleep(1)
+
+        # ----------------------------------------------------------------
+        # Step 4: Terminate the background stress-ng process
+        # ----------------------------------------------------------------
```
sacsant: Comment is unnecessary

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+        self._terminate_stressng()
+
+        # ----------------------------------------------------------------
+        # Step 5: Report results
+        # ----------------------------------------------------------------
```
sacsant: Comment is unnecessary

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
         self.log.info("Return code is %s", return_code)
         self.stress_ng_status_check(return_code)
 
+    # ------------------------------------------------------------------
+    # Helpers for the runtime scheduling-policy change test
+    # ------------------------------------------------------------------
```
sacsant, line 264: not required

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+            "stress-ng",
+            "--cpu", str(num_workers),
+            "--timeout", str(duration_secs),
+            "--metrics-brief",
+        ]
+        self.log.info("Spawning background stress-ng: %s", " ".join(cmd))
+        proc = subprocess.Popen(
+            cmd,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
```
sacsant, line 285: ubprocess.Popen is created with stdout=subprocess.PIPE and stderr=subprocess.PIPE but neither stream is ever read. stress-ng writes metrics and status to stderr (--metrics-brief). The OS pipe buffer (typically 64 KB) will fill up, causing stress-ng to block on its next write. Since _terminate_stressng calls proc.wait() without draining the pipes, the wait() will block too — resulting in a test hang.

Either discard output with DEVNULL, or capture it and log after termination:

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+                    stderr=subprocess.DEVNULL,
+                ).decode()
+                for token in raw.split():
+                    pids.add(int(token))
+            except subprocess.CalledProcessError:
+                self.log.warning(
+                    "pgrep found no children of PID %d", parent_pid)
+
+        # Always include the parent itself so we have at least one target
+        pids.add(parent_pid)
```
sacsant: _get_stressng_worker_pids unconditionally adds parent_pid to pids with pids.add(parent_pid) at the end. This means worker_pids can never be empty. The guard:

if not worker_pids:
    self._terminate_stressng()
    self.fail("No stress-ng worker PIDs found — cannot proceed.")

…is dead code and will never trigger. The real case of interest — no children (workers) found — is silently swallowed, and the test proceeds exercising only the parent PID (the stress-ng manager process, not a CPU stressor).

Suggested to check for children separately from the parent.

cpu/stressng_cpu.py
```diff
@@ -17,22 +17,59 @@
...
+    SCHED_IDLE:    {"name": "SCHED_IDLE",    "label": "idle", "rt_prio": 0},
+    SCHED_FIFO:    {"name": "SCHED_FIFO",    "label": "fifo", "rt_prio": 1},
+    SCHED_RR:      {"name": "SCHED_RR",      "label": "rr", "rt_prio": 1},
+    SCHED_DEADLINE: {"name": "SCHED_DEADLINE", "label": "deadline",
+                     "rt_prio": 0},
+}
+
+# Derived from POLICY_META so label strings are never duplicated.
+# chrt -p <pid> uses "--<label>" as the policy flag.
+CHRT_POLICY_FLAG = {
```
sacsant: This is never used anywhere in the code. Is this required?

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+            len(pids), sorted(pids))
+        return sorted(pids)
+
+    def _pid_alive(self, pid):
+        """Return True if *pid* still exists in /proc."""
+        return os.path.exists("/proc/%d" % pid)
+
+    def _read_proc_sched_policy(self, pid):
+        """
+        Read the current scheduling policy integer for *pid* from
```
sacsant: The docstring opens with "Read the current scheduling policy integer for pid from /proc/<pid>/status" and then contradicts itself in the same sentence, describing what actually happens (using os.sched_getscheduler()). The /proc/<pid>/status reference is wrong — that file does not contain the scheduling policy integer — and will confuse anyone reading this code.

Please change the docstring accordingly

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+        Requirements
+        ------------
+        - Must be run as root (CAP_SYS_NICE required for RT policies).
+        - ``stress-ng`` must be installed (built in setUp).
+        - ``chrt`` utility must be present on the system.
+
+        Parameters (from YAML)
+        ----------------------
+        - ``policy_change_runtime``  : seconds stress-ng runs (default 120)
+        - ``policy_change_workers``  : number of CPU stressor threads
```
sacsant, line 524: also this contradicts with default value mentioned in YAML

cpu/stressng_cpu.py
```diff
@@ -229,11 +267,493 @@ def test_cpu(self):
...
+
+        try:
+            raw = subprocess.check_output(
+                ["chrt", "-p", str(pid)],
+                stderr=subprocess.STDOUT,
+            ).decode()
+        except (subprocess.CalledProcessError, FileNotFoundError) as chrt_err:
+            self.log.warning(
+                "chrt verification skipped for PID %d: %s", pid, chrt_err)
+            return True, "chrt-unavailable"
```
sacsant: When chrt is not installed (FileNotFoundError) or fails (CalledProcessError), the method returns (True, "chrt-unavailable"). This means the secondary verification silently succeeds — the test may pass while only the primary os.sched_getscheduler() check actually ran. This is particularly concerning because the PR's stated purpose is dual verification. There is no warning to the user that the secondary verification was skipped for the whole test run (not just one PID).

sacsant (changes requested):

cpu/stressng_cpu.py
SamirMulani: Trimmed the Background section to a single essential paragraph. Removed the transition chain diagram, SCHED_DEADLINE explanation, Verification Method, and Test Flow sections.

cpu/stressng_cpu.py
SamirMulani: Split the combined except (subprocess.CalledProcessError, FileNotFoundError) into separate handlers. FileNotFoundError now returns (False, "chrt-not-installed") and emits a warning indicating that secondary verification is unavailable for the entire test run. CalledProcessError returns (False, "chrt-error") with a per-PID warning. Both cases are now reported as actual failures in all_results instead of being treated as successful secondary verification.

SamirMulani:
```
avocado run stressng_cpu.py -m stressng_cpu.py.data/stressng_cpu.yaml --max-parallel-tasks=1
Fetching asset from stressng_cpu.py:Stressngcpu.test_cpu
Fetching asset from stressng_cpu.py:Stressngcpu.test_runtime_policy_change
JOB ID     : 72f05568be7101d2c3012eb1f2c9daa33e4f4b43
JOB LOG    : /root/avocado-fvt-wrapper/results/job-2026-07-30T12.25-72f0556/job.log
 (1/2) stressng_cpu.py:Stressngcpu.test_cpu;run-test_mode-428f: STARTED

 (1/2) stressng_cpu.py:Stressngcpu.test_cpu;run-test_mode-428f: PASS (672.98 s)
 (2/2) stressng_cpu.py:Stressngcpu.test_runtime_policy_change;run-test_mode-428f: STARTED
 (2/2) stressng_cpu.py:Stressngcpu.test_runtime_policy_change;run-test_mode-428f: PASS (76.54 s)
RESULTS    : PASS 2 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /root/avocado-fvt-wrapper/results/job-2026-07-30T12.25-72f0556/results.html
JOB TIME   : 765.99 s
```

cpu/stressng_cpu.py
```diff
@@ -229,11 +259,461 @@ def test_cpu(self):
...
+            duration_secs=self.policy_change_runtime,
+        )
+
+        worker_pids = self._get_stressng_worker_pids(
+            self._stressng_proc.pid, settle_secs=3)
+
+        child_pids = [p for p in worker_pids
+                      if p != self._stressng_proc.pid]
+        if not child_pids:
+            self._terminate_stressng()
```
sacsant, line 544: _terminate_stressng() sets self._stressng_proc = None as its first action (line 682) before doing anything else. After calling _terminate_stressng() on line 544, the very next line (548) dereferences self._stressng_proc.pid, which will raise AttributeError: 'NoneType' object has no attribute 'pid' and mask the real failure with an uncaught exception instead of the intended self.fail() message.

Current Code:

 line 543-548
if not child_pids:
    self._terminate_stressng()     # sets self._stressng_proc = None !
    self.fail(
        "No stress-ng worker child PIDs found under PID %d — "
        "cannot exercise CPU stressor scheduling policies."
        % self._stressng_proc.pid)   # <-- AttributeError: None.pid

Suggested Fix:

Capture the PID before terminating:

if not child_pids:
    mgr_pid = self._stressng_proc.pid
    self._terminate_stressng()
    self.fail(
        "No stress-ng worker child PIDs found under PID %d — "
        "cannot exercise CPU stressor scheduling policies." % mgr_pid)

cpu/stressng_cpu.py
```diff
@@ -229,11 +259,461 @@ def test_cpu(self):
...
+        return results
+
+    def _verify_policy_for_pids(self, pids, expected_policy):
+        """
+        Verify that every living PID in *pids* is running under
+        *expected_policy*.
+
+        Uses two independent sources:
+          1. ``os.sched_getscheduler(pid)``  — kernel syscall (primary)
+          2. ``chrt -p <pid>``              — userspace tool (secondary)
```
sacsant, line 450: verify_policy_via_chrt() returns (False, "chrt-not-installed") or (False, "chrt-error") when chrt is absent or fails. In _verify_policy_for_pids(), any False from chrt_ok is unconditionally appended to failures[], which then causes self.fail(). This means the entire test will be marked FAILED on any system where chrt is not installed or where it temporarily fails (e.g. PID disappeared between set and verify). The PR description calls chrt a secondary check, but the code treats it as primary and mandatory. The FileNotFoundError warning at line 385 claims "secondary policy verification is unavailable for the entire test run" — but the loop never exits early after that; it will emit this warning for every PID of every transition, flooding the log.

Current Code:
_verify_policy_for_pids, lines 487-495

```
chrt_ok, chrt_actual = self._verify_policy_via_chrt(pid, expected_policy)
if not chrt_ok:
    failures.append(
        "PID %d: chrt mismatch — expected %s got %s"
        % (pid, policy_name, chrt_actual))

```
Suggested Fix:

Distinguish between chrt tool errors (warn only) and actual policy mismatches:

```
chrt_ok, chrt_actual = self._verify_policy_via_chrt(pid, expected_policy)
if not chrt_ok:
    if chrt_actual in ("chrt-not-installed", "chrt-error", "parse-error"):
        # Secondary tool unavailable; primary syscall result is authoritative
        self.log.warning(
            "PID %d: chrt secondary check skipped (%s)",
            pid, chrt_actual)
    else:
        failures.append(
            "PID %d: chrt mismatch — expected %s got %s"
            % (pid, policy_name, chrt_actual))
```

Also, track whether chrt is available once (e.g. a self._chrt_available flag set in setUp) and skip all secondary checks early rather than spawning a new chrt subprocess for every PID of every transition.

cpu/stressng_cpu.py
sacsant: can you check on this?

cpu/stressng_cpu.py
SamirMulani: Agree

SamirMulani:
```
JOB ID     : 461313e126c439cffdc91714ea9aa00ae74504bd
JOB LOG    : /root/avocado-fvt-wrapper/results/job-2026-08-12T16.08-461313e/job.log
 (1/2) stressng_cpu.py:Stressngcpu.test_cpu;run-test_mode-08be: STARTED
 (1/2) stressng_cpu.py:Stressngcpu.test_cpu;run-test_mode-08be:  PASS (1194.31 s)
 (2/2) stressng_cpu.py:Stressngcpu.test_runtime_policy_change;run-test_mode-08be: STARTED
 (2/2) stressng_cpu.py:Stressngcpu.test_runtime_policy_change;run-test_mode-08be:  PASS (134.86 s)
RESULTS    : PASS 2 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 0
JOB HTML   : /root/avocado-fvt-wrapper/results/job-2026-08-12T16.08-461313e/results.html
JOB TIME   : 1352.00 s
```

cpu/stressng_cpu.py
```diff
@@ -229,11 +289,479 @@ def test_cpu(self):
...
+
+        Returns a dict  {pid: True|False}  indicating per-PID success.
+        FIFO and RR require a real-time priority >= 1; all others use 0.
+        """
+        meta = POLICY_META.get(policy, {})
+        rt_prio = meta.get("rt_prio", 0)
+        param = os.sched_param(rt_prio)
+        results = {}
+        for pid in pids:
+            if not self._pid_alive(pid):
```
sacsant, line 455: If all workers exit after enumeration, _set_policy_for_pids() returns an empty result and _verify_policy_for_pids() returns no failures. The test can therefore report:

All runtime policy-change transitions PASSED.

without having successfully changed or verified any worker policy.
This is especially problematic because the test explicitly claims to validate policies on live worker threads.

Suggested fix:

Track whether at least one PID was successfully processed for each transition, and treat a transition with no live workers as a failure or an explicit skip. For example:

```
set_results = self._set_policy_for_pids(worker_pids, to_policy)

if not set_results:
    self.fail(
        "No live stress-ng workers remained for transition %s"
        % transition_label)

```
Alternatively, return explicit per-PID SKIP/FAIL results and ensure the final result summary fails when no worker was exercised.

cpu/stressng_cpu.py
```diff
@@ -229,11 +289,479 @@ def test_cpu(self):
...
+                pid, chrt_err)
+            return False, "chrt-error"
+
+        # Parse the policy name from chrt output
+        match = re.search(
+            r"current scheduling policy:\s+(\S+)", raw, re.IGNORECASE)
+        if not match:
+            self.log.warning(
+                "Could not parse chrt output for PID %d:\n%s", pid, raw)
+            return True, "parse-error"
```
sacsant, line 432: Returning True for an output-parsing failure makes the tuple semantics confusing. The caller currently compensates by checking "parse-error", but returning False would better reflect that the secondary check did not succeed

sacsant (approved):
Otherwise looks good to me.

PR #3240, opened 2026-07-30
workload/Daytrader: Add DayTrader7 workload test for IBM Power

SamirMulani:
DayTrader is a benchmark application that simulates an online stock trading system, originally developed by IBM as the "Trade Performance Benchmark Sample" and donated to the Apache Geronimo community in 2005.

A typical DayTrader user session covers login, portfolio view, stock price lookup, buy/sell transactions, and logout. Apache JMeter acts as the load driver, generating concurrent users against an OpenLiberty (WebSphere) application server backed by an IBM DB2 database. This mimics a real-world transactional workload and is used to measure throughput numbers on IBM Power LPARs.

This test automates the full DayTrader7 setup and workload execution on ppc64le to capture throughput metrics under a realistic online trading simulation. The configurable parameters (users, duration, instances) allow scaling the load to match different LPAR sizes and performance validation scenarios.

Test flow:
- Validates ppc64le architecture and root privileges
- Detects distro at runtime; selects appropriate package list
- Installs prerequisite OS packages via SoftwareManager
- On RHEL, probes enabled repos for Java 17 packages at runtime and installs with warn-and-skip (not cancel-on-failure) logic
- Selects and activates Java >= 17 via update-alternatives; exports JAVA_HOME dynamically for all sub-shells
- Scrapes the install_url directory listing dynamically; no filenames are hardcoded
- Downloads all kit files via fetch_asset, extracts tarball to /root
- Runs INSTALL_*.sh (DB2 + OpenLiberty installation)
- Runs DayTrader7_Run.sh with configurable users, duration, instances
- Reports pass/fail on exit code; warns on error keywords in output
- Cleans up all downloaded and extracted files from /root in tearDown

Files added:
- workload/Daytrader.py
- workload/Daytrader.py.data/daytrader.yaml
- workload/Daytrader.py.data/README.md

### DIFF
--- a/workload/Daytrader.py
+++ b/workload/Daytrader.py
@@ -1,0 +1,558 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+# Copyright: 2024 IBM
+# Author: Samir A Mulani <samir@linux.vnet.ibm.com>
+
+import os
+import re
+import shutil
+import socket
+import time
+
+from avocado import Test
+from avocado.utils import archive, distro, process
+from avocado.utils.software_manager.manager import SoftwareManager
+
+
+_INSTALL_ROOT = "/root"
+
+
+class Daytrader(Test):
+
+    def _run(self, cmd, fail_msg, ignore_status=False):
+        """Run a shell command and fail the test on non-zero exit status.
+
+        Args:
+            cmd (str): Shell command to execute.
+            fail_msg (str): Message passed to self.fail() on error.
+            ignore_status (bool): When True, a non-zero exit status is
+                tolerated and the result is returned without failing.
+
+        Returns:
+            avocado.utils.process.CmdResult: The command result object.
+        """
+        result = process.run(cmd, shell=True, ignore_status=True,
+                             sudo=False, verbose=True)
+        if result.exit_status != 0 and not ignore_status:
+            self.log.error(f"stderr: {result.stderr_text.strip()}")
+            self.fail(fail_msg)
+        return result
+
+    def _list_remote_files(self, base_url):
+        """Retrieve file names listed at *base_url* via an HTTP
+        directory index.
+
+        Parses ``href`` attributes from the HTML returned by *base_url* and
+        returns only plain-file entries (directories, parent links, and query
+        strings are excluded).  The list is used to discover the tarball and
+        install script names dynamically so that hardcoded file names are not
+        required.
+
+        Args:
+            base_url (str): HTTP(S) URL of the directory to list.
+
+        Returns:
+            list[str]: File names found at *base_url*.
+
+        Cancels the test if the URL is unreachable.
+        """
+        result = process.run(
+            f"curl -fsSL {base_url.rstrip('/')}/",
+            shell=True, ignore_status=True, verbose=False)
+        if result.exit_status != 0:
+            self.cancel(
+                f"Cannot reach install URL '{base_url}'. "
+                f"Check network connectivity and URL. rc={result.exit_status}")
+
+        raw = result.stdout_text
+        hrefs = re.findall(r'href="([^"]+)"', raw)
+        files = [h for h in hrefs
+                 if not h.startswith(("?", "/", ".."))
+                 and h != "./"
+                 and not h.endswith("/")]
+        self.log.info(f"Remote files found at {base_url}: {files}")
+        return files
+
+    def _validate_hostname(self):
+        """Validate that the system hostname is a resolvable FQDN.
+
+        Logs a warning when the hostname is not a FQDN.  Fails the test when
+        the FQDN cannot be resolved, because DB2 instance creation requires a
+        resolvable hostname.
+        """
+        hostname = socket.gethostname()
+        self.log.info(f"Current hostname: {hostname}")
+
+        if "." not in hostname:
+            self.log.warning(
+                f"Hostname '{hostname}' is not a FQDN. "
+                "DB2 instance creation may fail. "
+                "Set a FQDN with:  hostname <fqdn>  "
+                "and update /etc/hostname accordingly.")
+        else:
+            try:
+                resolved_ip = socket.gethostbyname(hostname)
+                self.log.info(
+                    f"Hostname '{hostname}' resolves to "
+                    f"{resolved_ip}")
+            except socket.gaierror as exc:
+                self.fail(
+                    f"Hostname '{hostname}' does not resolve: {exc}.  "
+                    "DB2 installation requires a resolvable hostname.")
+
+    def _detect_java_pkgs(self):
+        """Detect which Java packages (>= 17) are available in enabled repos.
+
+        Queries the package manager for all available ``java-*`` packages and
+        selects a JRE and JDK package at major version 17 or higher.  If no
+        qualifying JRE is found the test is cancelled, because DayTrader7
+        cannot run without Java >= 17.
+
+        Returns:
+            list[str]: Package names to install (JRE and optionally JDK).
+        """
+        self.log.info("Probing available Java packages (>= 17) via dnf ...")
+        result = process.run(
+            "dnf list available 2>/dev/null | grep -iE 'java-[0-9]'",
+            shell=True, ignore_status=True, verbose=False)
+
+        available = result.stdout_text.lower()
+        java_pkgs = []
+
+        # Collect all java-<N>-openjdk[-headless] entries where N >= 17.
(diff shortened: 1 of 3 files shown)


workload/Daytrader.py
```diff
@@ -0,0 +1,516 @@
...
+    "gcc-c++",
+    "make",
+    "kernel-devel",
+    "kernel-headers",
+    "libaio-devel",
+    "libstdc++-devel",
+    "pam-devel",
+    "numactl",
+]
+
```
PraveenPenguin: why we need packages has global scope , please have member variable better to define in setup

workload/Daytrader.py
```diff
@@ -0,0 +1,516 @@
...
+    def _run(self, cmd, fail_msg, ignore_status=False):
+        self.log.info("Running: %s", cmd)
+        result = process.run(cmd, shell=True, ignore_status=True,
+                             sudo=False, verbose=True)
+        if result.exit_status != 0 and not ignore_status:
+            self.log.error("Command failed (rc=%d): %s",
+                           result.exit_status, cmd)
+            self.log.error("stdout: %s", result.stdout_text.strip())
+            self.log.error("stderr: %s", result.stderr_text.strip())
```
PraveenPenguin, line 47: to much logging, please consider only stderr if looking to capture other log have proper like 

self.log.info -- for strdout but i feel it is redudent

workload/Daytrader.py
```diff
@@ -0,0 +1,516 @@
...
+                "Check network connectivity and URL. rc=%d" % (
+                    base_url, result.exit_status))
+
+        raw = result.stdout_text
+        hrefs = re.findall(r'href="([^"]+)"', raw)
+        files = [h for h in hrefs
+                 if not h.startswith(("?", "/", ".."))
+                 and h != "./"
+                 and not h.endswith("/")]
+        self.log.info("Remote files found at %s: %s", base_url, files)
```
PraveenPenguin, line 83: is this function really needed , is asset fecth not able to fetch the we should fail there itself

workload/Daytrader.py
```diff
@@ -0,0 +1,516 @@
...
+                "fail.  Set a FQDN with:  hostname <fqdn>  and update "
+                "/etc/hostname accordingly.", hostname)
+        else:
+            try:
+                resolved_ip = socket.gethostbyname(hostname)
+                self.log.info("Hostname '%s' resolves to %s",
+                              hostname, resolved_ip)
+            except socket.gaierror as exc:
+                self.log.warning(
+                    "Hostname '%s' does not resolve: %s.  "
```
PraveenPenguin, line 108: if we have an exception, do we need to fail the test ?

workload/Daytrader.py
```diff
@@ -0,0 +1,516 @@
...
+        if "java-17-openjdk-devel" in available:
+            java_pkgs.append("java-17-openjdk-devel")
+            self.log.info("Detected JDK package: java-17-openjdk-devel")
+        else:
+            self.log.warning(
+                "java-17-openjdk-devel not found in enabled repos.")
+
+        self.log.info("Java packages to install: %s",
+                      java_pkgs if java_pkgs else "(none detected)")
+        return java_pkgs
```
PraveenPenguin, line 159: are you looking only in Java 17?  even checking the Java version is mandatory?  than just check if java installed and have verison grater than 17 than continue else skip/fail test

PraveenPenguin (changes requested):
@SamirMulani i see lot of unnecessary code where it can simplied the use .. Please see we can trim those and focus on functionality testing

workload/Daytrader.py
SamirMulani: Agree

workload/Daytrader.py
SamirMulani: Yes, need to remove some error logging statements here.

workload/Daytrader.py
SamirMulani: Yes, this function is required. It retrieves and populates the file names from the remote location for later use, making the workflow dynamic instead of relying on hardcoded file names. Since the file names at the remote location may change over time, this function eliminates that dependency.

SamirMulani: is asset fecth not able to fetch the we should fail there itself -> Agree

workload/Daytrader.py
```diff
@@ -0,0 +1,516 @@
...
+    "libaio-devel",
+    "libstdc++-devel",
+    "pam-devel",
+    "numactl",
+]
+
+
+class Daytrader(Test):
+
+    def _run(self, cmd, fail_msg, ignore_status=False):
```
Naresh-ibm, line 32: @SamirMulani add the Docstrings

workload/Daytrader.py
```diff
@@ -0,0 +1,516 @@
...
+                             sudo=False, verbose=True)
+        if result.exit_status != 0 and not ignore_status:
+            self.log.error("Command failed (rc=%d): %s",
+                           result.exit_status, cmd)
+            self.log.error("stdout: %s", result.stdout_text.strip())
+            self.log.error("stderr: %s", result.stderr_text.strip())
+            self.fail(fail_msg)
+        return result
+
+    def _list_remote_files(self, base_url):
```
Naresh-ibm, line 51: same here

workload/Daytrader.py
```diff
@@ -0,0 +1,516 @@
...
+        raw = result.stdout_text
+        hrefs = re.findall(r'href="([^"]+)"', raw)
+        files = [h for h in hrefs
+                 if not h.startswith(("?", "/", ".."))
+                 and h != "./"
+                 and not h.endswith("/")]
+        self.log.info("Remote files found at %s: %s", base_url, files)
+        return files
+
+    def _validate_hostname(self):
```
Naresh-ibm, line 86: add docstring

workload/Daytrader.py
```diff
@@ -0,0 +1,516 @@
...
+            try:
+                resolved_ip = socket.gethostbyname(hostname)
+                self.log.info("Hostname '%s' resolves to %s",
+                              hostname, resolved_ip)
+            except socket.gaierror as exc:
+                self.log.warning(
+                    "Hostname '%s' does not resolve: %s.  "
+                    "DB2 installation may fail.", hostname, exc)
+
+    def _detect_java_pkgs(self):
```
Naresh-ibm, line 112: same here

workload/Daytrader.py
```diff
@@ -0,0 +1,516 @@
...
+            self.log.warning(
+                "Workload output contains potential failure keywords: %s. "
+                "Review the full log for details.", found_failures)
+
+        self.log.info(
+            "DayTrader7 workload completed successfully.  "
+            "users=%s, duration=%ss, instances=%s",
+            self.users, self.duration, self.instances)
+
+    def tearDown(self):
```
Naresh-ibm, line 525: add docstring to all the methods please

Naresh-ibm (changes requested):

workload/Daytrader.py
abdhaleegit: use f string

workload/Daytrader.py
```diff
@@ -0,0 +1,544 @@
...
+        Cancels the test if the URL is unreachable.
+        """
+        result = process.run(
+            "curl -fsSL %s" % base_url.rstrip("/") + "/",
+            shell=True, ignore_status=True, verbose=False)
+        if result.exit_status != 0:
+            self.cancel(
+                "Cannot reach install URL '%s'. "
+                "Check network connectivity and URL. rc=%d" % (
+                    base_url, result.exit_status))
```
abdhaleegit: same here use f string

workload/Daytrader.py
abdhaleegit: use fstring

workload/Daytrader.py
```diff
@@ -0,0 +1,544 @@
...
+
+        if "." not in hostname:
+            self.log.warning(
+                "Hostname '%s' is not a FQDN.  DB2 instance creation may "
+                "fail.  Set a FQDN with:  hostname <fqdn>  and update "
+                "/etc/hostname accordingly.", hostname)
+        else:
+            try:
+                resolved_ip = socket.gethostbyname(hostname)
+                self.log.info("Hostname '%s' resolves to %s",
```
abdhaleegit: use fstring

workload/Daytrader.py
```diff
@@ -0,0 +1,544 @@
...
+                                    available)
+        qualifying_jre = [(pkg, int(ver))
+                          for pkg, ver in jre_candidates if int(ver) >= 17]
+        if qualifying_jre:
+            # Prefer the highest available version; prefer headless variant.
+            qualifying_jre.sort(key=lambda x: (x[1], "headless" in x[0]),
+                                reverse=True)
+            best_jre = qualifying_jre[0][0]
+            java_pkgs.append(best_jre)
+            self.log.info("Selected JRE package: %s", best_jre)
```
abdhaleegit: f string

workload/Daytrader.py
```diff
@@ -0,0 +1,544 @@
...
+        _rhel_pkgs = _common_pkgs + [
+            "gcc",
+            "gcc-c++",
+            "make",
+            "kernel-devel",
+            "kernel-headers",
+            "libaio-devel",
+            "libstdc++-devel",
+            "pam-devel",
+            "numactl",
```
abdhaleegit: some more still go to common packages.. loooks same for rhel and sles

PraveenPenguin (approved):
LGTM

PR #3245, opened 2026-08-04
Adding enhancements to EEH SRIOV support

PavamanSubramaniyam:
Introducing a check to verify if secure boot is enabled or not Introducing a new parameter sriov in the yaml file The test_eeh_sriov function will only execute when sriov is yes If sriov is no then test_eeh_basic_pe function is executed Also, introduced the condition to check the number of EEH hits and misses

### DIFF
--- a/io/pci/EEH.py
+++ b/io/pci/EEH.py
@@ -29,6 +29,7 @@ from avocado.utils import distro
 from avocado.utils import dmesg
 from avocado.utils import multipath
 from avocado.utils import data_structures
+from avocado.utils import linux
 from avocado.utils.software_manager.manager import SoftwareManager
 from avocado.utils.network.interfaces import NetworkInterface
 from avocado.utils.network.hosts import LocalHost
@@ -67,6 +68,8 @@ class EEH(Test):
         """
         if 'ppc' not in distro.detect().arch:
             self.cancel("Processor is not ppc64")
+        if linux.is_os_secureboot_enabled():
+            self.cancel("EEH is not supported when Secure boot is enabled.")
         eeh_enable_file = "/sys/kernel/debug/powerpc/eeh_enable"
         if '0x1' not in genio.read_file(eeh_enable_file).strip():
             self.cancel("EEH is not enabled, please enable via FSP")
@@ -81,6 +84,7 @@ class EEH(Test):
             self.cancel("pciutils package is need to test")
         self.mem_addr = pci.get_memory_address(self.pci_device)
         self.mask = pci.get_mask(self.pci_device)
+        self.sriov = self.params.get("sriov", default="no")
         self.ipaddr = self.params.get("host_ip", default=None)
         if self.ipaddr:
             self.peer_ip = self.params.get("peer_ip", default=None)
@@ -143,6 +147,9 @@ class EEH(Test):
         enter_loop = True
         num_of_miss = 0
         num_of_hit = 0
+        if self.sriov == "yes":
+            self.cancel("EEH basic error injection on PE"
+                        "is not supported for SRIOV VF adapter ports")
         while num_of_hit < self.max_freeze:
             for func in self.function:
                 self.log.info("Running error inject on pe %s function %s",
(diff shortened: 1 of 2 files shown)


io/pci/EEH.py
```diff
@@ -219,89 +226,123 @@ def test_eeh_sriov(self):
...
+            bus_address = self.get_bus_address_from_journalctl(bus_id)
+            if not bus_address:
+                self.fail(
+                    f"Failed to get bus address for {self.pci_device} "
+                    f"from journalctl")
+                self.log.info(f"Bus address from journalctl: {bus_address}")
+            enter_loop = True
+            num_of_miss = 0
+            num_of_hit = 0
+            mask = "0xffffffffffc00000"
```
abdhaleegit, line 251: You have hardcoded mask .. the mask changes when the memory region changes i.e.. example 

Region 0: 4MB  MASK: 0xffffffffffc00000
Region 2: 2GB MASK: 0xffffffff80000000
Region 4: 32MB MASK:0xfffffffffe000000

so error is always injected in one region..

abdhaleegit (commented):
rest all looks good

io/pci/EEH.py
PavamanSubramaniyam: This is the value we are using for injecting EEH to SRIOV VF adapter ports after getting this value from development team. So, this value was hardcoded as the mask value by Vaishnavi from her PR which got merged for EEH SRIOV support. I have just used the same code here.

PavamanSubramaniyam:
Execution results logs for the EEH tests with SRIOV ROCE VF ports adapters:

avocado run --max-parallel-tasks=1 EEH.py -m EEH.py.data/EEH.yaml
JOB ID     : e372b2f12bce270ac3c0b9f018ff2526d8432ca1
JOB LOG    : /root/avocado-fvt-wrapper/results/job-2026-08-02T02.33-e372b2f/job.log
 (1/2) EEH.py:EEH.test_eeh_basic_pe;run-fda3: STARTED
 (1/2) EEH.py:EEH.test_eeh_basic_pe;run-fda3: CANCEL: EEH basic error injection on PE is not supported for SRIOV VF adapter ports (3.52 s)
 (2/2) EEH.py:EEH.test_eeh_sriov;run-fda3: STARTED
 (2/2) EEH.py:EEH.test_eeh_sriov;run-fda3: PASS (206.05 s)
RESULTS    : PASS 1 | ERROR 0 | FAIL 0 | SKIP 0 | WARN 0 | INTERRUPT 0 | CANCEL 1
JOB HTML   : /root/avocado-fvt-wrapper/results/job-2026-08-02T02.33-e372b2f/results.html
JOB TIME   : 232.41 s

PavamanSubramaniyam:
Complete and detail result logs are attached in the corresponding Jira.

io/pci/EEH.py
vaishnavibhat: Hi @abdhaleegit ,
As @PavamanSubramaniyam mentioned, Dev suggested to use the hardcoded mask value for sriov interfaces . For direct attach interfaces, it can be calculated as done in the other part of the code. 
Thank you.

io/pci/EEH.py
```diff
@@ -143,6 +147,9 @@ def test_eeh_basic_pe(self):
         enter_loop = True
         num_of_miss = 0
         num_of_hit = 0
+        if self.sriov == "yes":
+            self.cancel("EEH basic error injection on PE"
+                        "is not supported for SRIOV VF adapter ports")
```
Naresh-ibm, line 152: @PavamanSubramaniyam this contradict for what we are doing here, we are trying to inject the error on sriov VF and your cancelling on yes value.
I am confuse, can you please elaborate and re-check this condition, if something unknowingly get added

PavamanSubramaniyam: The test_eeh_basic_pe() function is handling EEH error injection for dedicated adapter ports and not for SRIOV VF ports. 
So, if the self.sriov is set to yes, then this function should not be executed since this function is only for dedicated adapter ports.

The test_eeh_sriov() function is handling EEH error injection for SRIOV VF ports.

abdhaleegit (approved):
LGTM

PR #3252, opened 2026-08-05
cpu/ebizzy.py fix: update ebizzy download link to use HTTPS

SamirMulani:
Fixed the wget issue where we are using the link to download the ebizzy workload, which failed to download the file.
Ex: http://sourceforge.net/projects/ebizzy/files/ebizzy/0.3/ebizzy-0.3.tar.gz

Now I have added the link below:
Ex: https://sourceforge.net/projects/ebizzy/files/ebizzy/0.3/ebizzy-0.3.tar.gz


Naresh-ibm (approved):
LGTM

